PR #6, opened 2019-09-25
feat(ZFSPV): adding encryption in ZFSVolume CR

pawanpraka1:
Also adding support to inherit the properties from ZPOOL
which are not listed in the storage class, ZFS driver will
not pass default values while creating the volume. Those
properties will be inherited from the ZPOOL.

we can use the encryption option in storage class 
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: openebs-zfspv
allowVolumeExpansion: true
parameters:
  blocksize: "4k"
  compression: "on"
  dedup: "on"
  thinprovision: "yes"
  encryption: "on"
  keyformat: "raw"
  keylocation: "file:///home/keys/key"
  poolname: "zfspv-pool"
provisioner: openebs.io/zfs
```

Just a note, the key file should be mounted inside the node-agent container so that we can use that file while provisioning the volume. keyformat can be raw, hex or passphrase.

### DIFF
--- a/pkg/builder/build.go
+++ b/pkg/builder/build.go
@@ -106,6 +106,18 @@ func (b *Builder) WithEncryption(encr string) *Builder {
 	return b
 }
 
+// WithKeyLocation sets the encryption key location on ZFSVolume
+func (b *Builder) WithKeyLocation(kl string) *Builder {
+	b.volume.Object.Spec.KeyLocation = kl
+	return b
+}
+
+// WithKeyFormat sets the encryption key format on ZFSVolume
+func (b *Builder) WithKeyFormat(kf string) *Builder {
+	b.volume.Object.Spec.KeyFormat = kf
+	return b
+}
+
 // WithCompression sets compression of ZFSVolume
 func (b *Builder) WithCompression(compression string) *Builder {
 	b.volume.Object.Spec.Compression = compression
--- a/pkg/builder/build.go
+++ b/pkg/builder/build.go
@@ -54,7 +54,7 @@ func BuildFrom(volume *apis.ZFSVolume) *Builder {
 	}
 }
 
-// WithNamespace sets the namespace of csi volume
+// WithNamespace sets the namespace of  ZFSVolume
 func (b *Builder) WithNamespace(namespace string) *Builder {
 	if namespace == "" {
 		b.errs = append(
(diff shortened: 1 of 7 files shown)


pkg/builder/build.go
```diff
@@ -100,42 +100,44 @@ func (b *Builder) WithCapacity(capacity string) *Builder {
...
+	b.volume.Object.Spec.Encryption = encr
+	return b
+}
 
-	comp := "off"
-	if compression == "on" {
-		comp = "on"
-	}
-	b.volume.Object.Spec.Compression = comp
+// WithKeyLocation sets the encryption key location on ZFSVolume
```
kmova: Is the location/path on the node - absolute or relative path?

pawanpraka1: absolute path. also if someone wants to use keylocation, he has to mount that path to the same path in node-agent.

kmova: It will be good to provide an example usage for this one in the PR description or the commit message.

pawanpraka1: done.

ptthon:
> Also adding support to inherit the properties from ZPOOL which are not listed in the storage class, ZFS driver will not pass default values while creating the volume. Those properties will be inherited from the ZPOOL.
> 
> we can use the encryption option in storage class
> 
> ```
> apiVersion: storage.k8s.io/v1
> kind: StorageClass
> metadata:
>   name: openebs-zfspv
> allowVolumeExpansion: true
> parameters:
>   blocksize: "4k"
>   compression: "on"
>   dedup: "on"
>   thinprovision: "yes"
>   encryption: "on"
>   keyformat: "raw"
>   keylocation: "file:///home/keys/key"
>   poolname: "zfspv-pool"
> provisioner: openebs.io/zfs
> ```
> 
> Just a note, the key file should be mounted inside the node-agent container so that we can use that file while provisioning the volume. keyformat can be raw, hex or passphrase.
> 
> Signed-off-by: Pawan [pawan@mayadata.io](mailto:pawan@mayadata.io)
I have mounted the key file inside the node-agent containered ,but create volume failt

zfs.csi.openebs.io_openebs-zfs-controller-0_5028c459-a44a-4063-9e5b-60efc06d2281  failed to provision volume with StorageClass "openebs-zfspv-enc": rpc error: code = Internal desc = not able to provision the volume, nodes [master3], err : zfs: volume creation failed

ptthon:
I have the secret key file on all nodes of /home/keys , but create volume failt

the err is:
zfs.csi.openebs.io_openebs-zfs-controller-0_5028c459-a44a-4063-9e5b-60efc06d2281  failed to provision volume with StorageClass "openebs-zfspv-enc": rpc error: code = Internal desc = not able to provision the volume, nodes [master3], err : zfs: volume creation failed

pawanpraka1:
@ptthon please raise an issue for this. Also please upload the node agent logs.

PR #7, opened 2019-10-15
adding topology support for zfspv

pawanpraka1:
We can provide the topology in storage class where the zfs pools are setup
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: openebs-zfspv
allowVolumeExpansion: true
parameters:
  blocksize: "4k"
  compression: "on"
  dedup: "on"
  thinprovision: "yes"
  poolname: "zfspv-pool"
provisioner: zfs-localpv
volumeBindingMode: WaitForFirstConsumer
allowedTopologies:
- matchLabelExpressions:
  - key: kubernetes.io/hostname
    values:
      - gke-zfspv-pawan-default-pool-c8929518-cgd4
      - gke-zfspv-pawan-default-pool-c8929518-dxzc
```

The scheduler will take care of scheduling the application on the appropriate node.

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -85,7 +85,14 @@ func (c *ZVController) syncZV(zv *apis.ZFSVolume) error {
 			zvol.RemoveZvolFinalizer(zv)
 		}
 	} else {
-		err = zvol.SetZvolProp(zv)
+		if zv.Finalizers != nil {
+			err = zvol.SetZvolProp(zv)
+		} else {
+			err = zvol.CreateZvol(zv)
+			if err == nil {
+				err = zvol.UpdateZvolInfo(zv)
+			}
+		}
 	}
 	return err
 }
@@ -101,11 +108,8 @@ func (c *ZVController) addZV(obj interface{}) {
 	if zvol.NodeID != zv.Spec.OwnerNodeID {
 		return
 	}
-	// TODO(pawan) scheduler will schedule the volume
-	// on a node and populate the OwnerNodeID accordingly.
-	// We need to create the zfs volume in that case.
 	logrus.Infof("Got add event for ZV %s/%s", zv.Spec.PoolName, zv.Name)
-	//c.enqueueZV(zv)
+	c.enqueueZV(zv)
 }
 
 // updateZV is the update event handler for CstorVolumeClaim
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -85,6 +85,9 @@ func (c *ZVController) syncZV(zv *apis.ZFSVolume) error {
 			zvol.RemoveZvolFinalizer(zv)
 		}
 	} else {
+		// if finalizer is not set then it means we are creating
+		// the volume. And if it is set then volume has already been
+		// created and this event is for property change only.
 		if zv.Finalizers != nil {
 			err = zvol.SetZvolProp(zv)
 		} else {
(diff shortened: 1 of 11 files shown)


deploy/sample/fio.yaml
```diff
@@ -12,8 +12,14 @@ parameters:
   #keyformat: "raw"
   #keylocation: "file:///home/pawan/key"
   poolname: "zfspv-pool"
-provisioner: openebs.io/zfs
+provisioner: zfs-localpv
```
kmova: Was there a change in the provisioner name recently? or was this never working?

cmd/controller/controller.go
```diff
@@ -85,7 +85,14 @@ func (c *ZVController) syncZV(zv *apis.ZFSVolume) error {
 			zvol.RemoveZvolFinalizer(zv)
 		}
 	} else {
-		err = zvol.SetZvolProp(zv)
+		if zv.Finalizers != nil {
```
kmova: Add a comment here please.

pkg/zfs/volume.go
```diff
@@ -31,9 +31,10 @@ const (
 	OpenEBSNamespaceKey string = "OPENEBS_NAMESPACE"
 	// ZFSFinalizer for the ZfsVolume CR
 	ZFSFinalizer string = "zfs.openebs.io/finalizer"
-	// ZFSNodeKey will be used to insert Label
-	// in ZfsVolume CR
+	// ZFSNodeKey will be used to insert Label in ZfsVolume CR
 	ZFSNodeKey string = "kubernetes.io/nodename"
+	// ZFSTopologyKey is supported topology key for the zfs driver
+	ZFSTopologyKey string = "kubernetes.io/hostname"
```
kmova: Can the topology be specified using node selectors ?

deploy/sample/fio.yaml
pawanpraka1: in latest provisioner --provisioner flag is depricated. We have to use the driver name instead of that. So it was working before, but after changeing the image, it stopped working.

deploy/zfs-operator.yaml
```diff
@@ -343,7 +343,7 @@ spec:
       hostNetwork: true
       containers:
         - name: csi-node-driver-registrar
-          image: quay.io/k8scsi/csi-node-driver-registrar:v1.0.1
+          image: quay.io/k8scsi/csi-node-driver-registrar:v1.2.0
```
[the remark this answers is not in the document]
pawanpraka1: yeah ,they can.

pkg/zfs/volume.go
pawanpraka1: yeah, you can specify both. But make sure both the setting are compliance with each other, otherwise pod will not be scheduled.

cmd/controller/controller.go
pawanpraka1: done.

cmd/controller/controller.go
```diff
@@ -85,7 +85,17 @@ func (c *ZVController) syncZV(zv *apis.ZFSVolume) error {
 			zvol.RemoveZvolFinalizer(zv)
 		}
 	} else {
-		err = zvol.SetZvolProp(zv)
+		// if finalizer is not set then it means we are creating
```
kmova: Will there be a case where finalizer is removed as part of deletion. should also check for deletion flag to ignore the event?

pawanpraka1: we are already doing that in the "if" check. this is else of that.

PR #8, opened 2019-11-04
feat(ZFSPV): scheduler for ZFSPV

pawanpraka1:
The scheduler will go through all the nodes as per
topology information and it will pick the node which has less
volume provisioned in the given pool.

lets say there are 2 nodes node1 and node2 with below pool configuration :-
```
node1
|
|-----> pool1
|         |
|         |------> pvc1
|         |------> pvc2
|-----> pool2
          |------> pvc3

node2
|
|-----> pool1
|         |
|         |------> pvc4
|-----> pool2
          |------> pvc5
          |------> pvc6
```
So if application is using pool1 as shown in the below storage class, then ZFS driver will schedule it on node2 as it has one volume as compared to node1 which has 2 volumes in pool1.
```yaml
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: openebs-zfspv
parameters:
  blocksize: "4k"
  compression: "on"
  dedup: "on"
  thinprovision: "yes"
  scheduler: "VolumeWeighted"
  poolname: "pool1"
```

So if application is using pool2 as shown in the below storage class, then ZFS driver will schedule it on node1 as it has one volume only as compared node2 which has 2 volumes in pool2.
```yaml
kind: StorageClass
apiVersion: storage.k8s.io/v1
metadata:
  name: openebs-zfspv
parameters:
  blocksize: "4k"
  compression: "on"
  dedup: "on"
  thinprovision: "yes"
  scheduler: "VolumeWeighted"
  poolname: "pool2"
```
In case of same number of volumes on all the nodes for the given pool, it can pick any node and schedule the PV on that.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -51,7 +51,7 @@ func main() {
 	)
 
 	cmd.PersistentFlags().StringVar(
-		&config.DriverName, "name", "zfs-localpv", "Name of this driver",
+		&config.DriverName, "name", "zfs.csi.openebs.io", "Name of this driver",
 	)
 
 	cmd.PersistentFlags().StringVar(
--- a/deploy/sample/fio.yaml
+++ b/deploy/sample/fio.yaml
@@ -13,7 +13,6 @@ parameters:
   #keylocation: "file:///home/pawan/key"
   poolname: "zfspv-pool"
 provisioner: zfs-localpv
-volumeBindingMode: WaitForFirstConsumer
 allowedTopologies:
 - matchLabelExpressions:
   - key: kubernetes.io/hostname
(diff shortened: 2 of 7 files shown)


cmd/main.go
```diff
@@ -51,7 +51,7 @@ func main() {
 	)
 
 	cmd.PersistentFlags().StringVar(
-		&config.DriverName, "name", "zfs-localpv", "Name of this driver",
+		&config.DriverName, "name", "zfs.csi.openebs.io", "Name of this driver",
```
kmova: can it be `zfs-localpv.csi.openebs.io` ?

PR #15, opened 2019-11-13
feat(ZFSPV): adding support for applications to create "zfs" flesystem

pawanpraka1:
Application can now create a storageclass to create zfs filesystem

apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: openebs-zfspv
allowVolumeExpansion: true
parameters:
  recordsize: "4k"
  fstype: "zfs"
  poolname: "zfspv-pool"

ZFSPV was supporting ext2/3/4 and xfs filesystem only which
adds one extra filesystem layer on top of ZFS filesystem. So now
we can directly write to the ZFS filesystem and get the optimal performance
by directly creating ZFS filesystem for storage.

Renamed the blocksize to volblocksize, which will be used for creating the zvols and if it is provided for dataset, it will be ignored and recordsize will be used for creating the zfs datasets, if it is provided for zvol, it will be ignored

Also added support to change the recordsize for zfs datasets. 

### DIFF
--- a/deploy/sample/percona.yaml
+++ b/deploy/sample/percona.yaml
@@ -4,7 +4,7 @@ metadata:
   name: openebs-zfspv
 allowVolumeExpansion: true
 parameters:
-  recordsize: "4k"
+  volblocksize: "4k"
   compression: "on"
   dedup: "on"
   thinprovision: "yes"
--- a/deploy/sample/percona.yaml
+++ b/deploy/sample/percona.yaml
@@ -4,7 +4,7 @@ metadata:
   name: openebs-zfspv
 allowVolumeExpansion: true
 parameters:
-  blocksize: "4k"
+  recordsize: "4k"
   compression: "on"
   dedup: "on"
   thinprovision: "yes"
--- a/deploy/sample/zfspvcr.yaml
+++ b/deploy/sample/zfspvcr.yaml
@@ -4,7 +4,6 @@ metadata:
   name: pvc-37b07ad6-db68-11e9-bbb6-000c296e38d9
   namespace: openebs
 spec:
-  blocksize: 4k
   capacity: "4294967296"
   compression: "off"
   dedup: "off"
@@ -13,5 +12,6 @@ spec:
   keylocation: ""
   ownerNodeID: zfspv-node1
   poolName: zfspv-pool
+  recordsize: 8k
   thinProvision: "off"
-
+  volumeType: DATASET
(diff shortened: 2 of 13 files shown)


deploy/sample/zfspvcr.yaml
```diff
@@ -13,5 +12,6 @@ spec:
   keylocation: ""
   ownerNodeID: zfspv-node1
   poolName: zfspv-pool
+  recordsize: 8k
   thinProvision: "off"
-
+  volumeType: DATASET
```
kmova: can we use the terms used by the ZFS itself.. is it zvol and zfs?

deploy/sample/percona.yaml
```diff
@@ -4,7 +4,7 @@ metadata:
   name: openebs-zfspv
 allowVolumeExpansion: true
 parameters:
-  blocksize: "4k"
+  recordsize: "4k"
```
kmova: should we support both depending on zfs or zvol? to match the zfs naming convention?

deploy/sample/zfspvcr.yaml
pawanpraka1: zfs dataset and zfs volume(zvol) are the terms used by zfs.

https://www.freebsd.org/doc/handbook/zfs-term.html
http://fibrevillage.com/storage/168-zfs-pool-zfs-datasets-and-zfs-volumes

deploy/sample/percona.yaml
pawanpraka1: I am missing C union here :). I feel recordsize for block devices will mean volblock size only. Just keeping both the field for both of them does not look good.

deploy/sample/zfspvcr.yaml
kmova: Also can this be auto determined from within the code based on the fs type. Also to keep it consistent with SC - how about adding fstype to the ZFSVolume CR instead of volumeType.

deploy/sample/percona.yaml
kmova: it might get confusing to have recordsize for zvol.  recordsize and volblocksize are better. Each should have a default value and be optional.

pawanpraka1: hmmm, so while doing zfs property change, user will see both volblocksize and recordsize.

deploy/sample/zfspvcr.yaml
pawanpraka1: the idea here is to store all the info related to ZFSVolume. By looking at resource we can get all the info like what is size, what are all properties that are provided while creating it, whether it is a dataset or zvol.

deploy/sample/percona.yaml
pawanpraka1: added volblocksize and recordsize both. They will make sese for zvol and dataset respectively.

PR #36, opened 2020-01-16
chore(metrics): adding list of zfs metrics exposed by prometheus

pawanpraka1:
adding list of zfs metrics exposed by prometheus in the prometheus setup doc.

### DIFF
--- a/docs/prometheus-monitoring.md
+++ b/docs/prometheus-monitoring.md
@@ -1,6 +1,10 @@
+### Prerequisite
+
+We need to have k8s version 1.15+ to access the CSI volume metrics.
+
 ### Setup helm
 
-This step uses helm the kubernetes package manager. If you not setup the helm then do the below the configuration, otherwise move to next step.
+This step uses helm the kubernetes package manager. If you have not setup the helm then do the below configuration, otherwise move to the next step.
 
 ```
 $ helm version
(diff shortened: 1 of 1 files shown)


docs/prometheus-monitoring.md
```diff
@@ -394,3 +398,207 @@ gke-zfspv-pawan-default-pool-3e407350-xvzp   Ready    <none>   103m   v1.15.4-gk
 ```
 
 In this case we can access the alert manager via url http://34.94.3.140:30093/
+
+### ZFS metrics exposed by Prometheus
```
kmova: Are there any addon exporters required or this comes from node exporter? Does it automatically detect zpool on node? What if two pools are present?

pawanpraka1: no, we don;t need to deploy any addon. This comes from the node exporter only. Yes it automatically detects the zpool on the node. Sample o/p of node_zfs_zpool_reads metrics if there are two pools(pool1 and zfspv-pool) in the system :- 

node_zfs_zpool_reads{endpoint="metrics",instance="10.168.0.50:9100",job="node-exporter",namespace="default",pod="prometheus-operator-prometheus-node-exporter-qmvnm",service="prometheus-operator-prometheus-node-exporter",zpool="pool1"} | 11
-- | --
node_zfs_zpool_reads{endpoint="metrics",instance="10.168.0.50:9100",job="node-exporter",namespace="default",pod="prometheus-operator-prometheus-node-exporter-qmvnm",service="prometheus-operator-prometheus-node-exporter",zpool="zfspv-pool"}

PR #49, opened 2020-02-26
feat(analytics): adding google analytics for ZFSPV

pawanpraka1:
Whenever a volume is provisioned and de-provisioned we will send a google event with following details : 
1.  pvName (will shown as app title in google analytics)
2.  size of the volume
3.  event type : volume-provision, volume-deprovision
4. storage type zfs-localpv
5. replicacount as 1
6. ClientId as default namespace uuid

Apart from this, we send the event once in 24 hr, which will have some info like number of nodes, node type, kubernetes version etc.

This metric is cotrolled by OPENEBS_IO_ENABLE_ANALYTICS env. We can set it to false if we don't want to send the metrics.


deploy/zfs-operator.yaml
```diff
@@ -512,6 +512,9 @@ rules:
   - apiGroups: [""]
     resources: ["secrets"]
     verbs: ["get", "list"]
+  - apiGroups: [""]
+    resources: ["namespaces"]
```
kmova: should access to node be provided?

kmova:
Hi @pawanpraka1 : Can you update the PR description with the following details: 

- Where the events are inserted
- Event type and parameters passed with each event.

deploy/zfs-operator.yaml
pawanpraka1: that is already there as a part of openebs-zfs-attacher-role.

pkg/usage/const.go
```diff
@@ -0,0 +1,49 @@
...
+
+	// Install event is sent on pod starts
+	InstallEvent string = "install"
+	// Ping event is sent periodically
+	Ping string = "ping"
+	// VolumeProvision event is sent when a volume is created
+	VolumeProvision string = "volume-provision"
+	//VolumeDeprovision event is sent when a volume is deleted
+	VolumeDeprovision string = "volume-deprovision"
+	AppName           string = "OpenEBS"
```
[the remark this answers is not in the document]
pawanpraka1: The thought is Appname will be OpenEBS which is used to set appID, in OpenEBS we are using "zfs-localpv" storage engine which we sending as sgtype, so here AppId is OpenEBS and Appname is "zfs-localpv".

pkg/usage/ping.go
```diff
@@ -0,0 +1,63 @@
...
+	u := New()
+	duration := getPingPeriod()
+	ticker := time.NewTicker(duration)
+	for _ = range ticker.C {
+		u.Build().
+			InstallBuilder(true).
+			SetCategory(Ping).
+			Send()
+	}
+}
```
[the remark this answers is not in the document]
pawanpraka1: this can work seperately without openebs instaled, so added this ping.

deploy/zfs-operator.yaml
```diff
@@ -647,6 +650,10 @@ spec:
               value: unix:///var/lib/csi/sockets/pluginproxy/csi.sock
             - name: OPENEBS_NAMESPACE
               value: openebs
+            - name: OPENEBS_IO_INSTALLER_TYPE
```
vharsh: Pending addition to helm charts cc @ChandanSagar

pkg/driver/controller.go
```diff
@@ -58,6 +59,19 @@ var SupportedVolumeCapabilityAccessModes = []*csi.VolumeCapability_AccessMode{
 	},
 }
 
+// sendEventOrIgnore sends anonymous local-pv provision/delete events
+func sendEventOrIgnore(pvName, capacity, stgType, method string) {
+	if zfs.GoogleAnalyticsEnabled == "true" {
+		analytics.New().Build().ApplicationBuilder().
```
vharsh: DoubleCheck: Confirm if AppName, AppID, AppVersion, etc are being set, etc.

pkg/usage/const.go
```diff
@@ -0,0 +1,49 @@
...
+const (
+	// GAclientID is the unique code of OpenEBS project in Google Analytics
+	GAclientID string = "UA-127388617-1"
+
+	// supported events categories
+
+	// Install event is sent on pod starts
+	InstallEvent string = "install"
+	// Ping event is sent periodically
+	Ping string = "ping"
```
vharsh: As suggested by @kmova 
```suggestion
	Ping string = "zfs-ping"
```

pkg/usage/versionset.go
```diff
@@ -0,0 +1,111 @@
...
+	v.k8sArch = k8s.Platform
+	v.k8sVersion = k8s.GitVersion
+	env.Set(clusterArch, v.k8sArch)
+	env.Set(clusterVersion, v.k8sVersion)
+	v.nodeType, err = k8sapi.GetOSAndKernelVersion()
+	env.Set(nodeType, v.nodeType)
+	if err != nil {
+		return err
+	}
+	v.openebsVersion = openebsversion.GetVersionDetails()
```
vharsh: (optional) `zfsLocalpvVersion`

Aside, if the app-versions are made to match the OpenEBS releases, we are good, if they are not we might want to add a prefix such as `zfs_lpv`, so that the version `zfs_lpv-0.9.1` and OpenEBS `0.9.1` aren't mixed up together esp if the AppNames are kept same as `OpenEBS`.

cc @kmova opinions?

pkg/usage/usage.go
```diff
@@ -0,0 +1,257 @@
...
+}
+
+// Wrapper for setting replica count for volume events
+// NOTE: This doesn't get the replica count in a volume de-provision event.
+// TODO: Pick the current value of replica-count from the CAS-engine
+func (u *Usage) SetReplicaCount(count, method string) *Usage {
+	if method == VolumeProvision && count == "" {
+		// Case: When volume-provision the replica count isn't specified
+		// it is set to three by default by the m-apiserver
+		u.SetAction(DefaultReplicaCount)
```
vharsh: Is the default replication factor in zfs also set to `3`, if nothing is specified in the SC/something else in the CSI?

pkg/usage/const.go
```diff
@@ -0,0 +1,49 @@
...
+	RunningStatus      string = "running"
+	EventLabelNode     string = "nodes"
+	EventLabelCapacity string = "capacity"
+
+	// Event action
+	Replica             string = "replica:"
+	DefaultReplicaCount string = "replica:3"
+
+	// Event application name constant for volume event
+	DefaultCASType string = "jiva"
```
vharsh:
```suggestion
	DefaultCASType string = "zfs-localpv"
```
Also I guess there's no concept of CASType in zfslocalpv and there's only one, so this can be named appropriately or removed as this isn't used from the place where `sendEventOrIgnore` is being called.

pkg/usage/const.go
```diff
@@ -0,0 +1,49 @@
...
+	AppName           string = "OpenEBS"
+
+	// Event labels
+	RunningStatus      string = "running"
+	EventLabelNode     string = "nodes"
+	EventLabelCapacity string = "capacity"
+
+	// Event action
+	Replica             string = "replica:"
+	DefaultReplicaCount string = "replica:3"
```
vharsh: Please modify the default replication count value

pkg/usage/const.go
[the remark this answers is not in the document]
vharsh: cc @kmova PTAL again from the perspective of knowing the versions of OpenEBS & ZFS-localPV being used in the world, having a same AppName might leave us guessing the app-versions if the AppName is same as what's being sent out via m-apiserver.

pkg/usage/usage.go
pawanpraka1: usign  SetReplicaCount(analytics.LocalPVReplicaCount, method). which has 1 replica count. If that is not passed then Dafault will be taken.

pawanpraka1: will change DefaultReplicaCount to 1

pkg/usage/versionset.go
pawanpraka1: added "zfs-" as prefix to the version tag.

PR #51, opened 2020-03-02
feat(resize): adding Online volume expansion support for ZFSPV

pawanpraka1:
We can resize the volume by updating the PVC yaml to
the desired size and apply it. The ZFS Driver will take care
of updating the quota in case of dataset. If we are using a
Zvol and have mounted it as ext4 or xfs filesystem, the driver will take
care of expanding the volume via reize2fs/xfs_growfs binaries.

For resize, storageclass that provisions the pvc must support resize. We should have allowVolumeExpansion as true in storageclass
    
    ```yaml
    apiVersion: storage.k8s.io/v1
    kind: StorageClass
    metadata:
      name: openebs-zfspv
    allowVolumeExpansion: true
    parameters:
      poolname: "zfspv-pool"
    
    ```

### DIFF
--- a/pkg/zfs/resize.go
+++ b/pkg/zfs/resize.go
@@ -1,0 +1,83 @@
+/*
+Copyright 2020 The Kubernetes Authors.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


pkg/zfs/resize.go
```diff
@@ -0,0 +1,82 @@
...
+	cmd := exec.Command("xfs_growfs", path)
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		logrus.Errorf("zfspv: resize failed error: %s", string(out))
+		return err
+	}
+	return nil
+}
+
+// HandleVolResize resizes the filesystem
```
kmova: how does resize get handled for ZFS datasets?

pawanpraka1: for zfs dataset, we just have to set the quota, nothing else needs to be done.

Issue #55, opened 2020-03-10
quota is not set for the cloned volume in case of zfs file system


w3aman:
**what happened:**
Deploy application with pvc of 2Gi. After exec into the application if we do `df -h` we see the correct claimsize on mount point.
```root@percona-58579467dc-rmw5n:/# df -h
Filesystem                                          Size  Used Avail Use% Mounted on
overlay                                              98G  9.3G   84G  10% /
tmpfs                                                64M     0   64M   0% /dev
tmpfs                                               3.9G     0  3.9G   0% /sys/fs/cgroup
/dev/sda1                                            98G  9.3G   84G  10% /etc/hosts
shm                                                  64M     0   64M   0% /dev/shm
test-pool/pvc-6c79f1f9-dc91-408e-b0e1-aa9120dae6b3  2.0G  215M  1.8G  11% /var/lib/mysql
tmpfs                                               3.9G   12K  3.9G   1% /run/secrets/kubernetes.io/serviceaccount
tmpfs                                               3.9G     0  3.9G   0% /proc/acpi
tmpfs                                               3.9G     0  3.9G   0% /proc/scsi
tmpfs                                               3.9G     0  3.9G   0% /sys/firmware
```
But when tried to clone the volume and used it, exec into application show `df -h` as some different size which is not the clone volume size. In my case it is showing the zpool capacity.

```
root@percona1-7c776f8558-6cskw:/# df -h
Filesystem                                          Size  Used Avail Use% Mounted on
overlay                                              98G  9.3G   84G  10% /
tmpfs                                                64M     0   64M   0% /dev
tmpfs                                               3.9G     0  3.9G   0% /sys/fs/cgroup
/dev/sda1                                            98G  9.3G   84G  10% /etc/hosts
shm                                                  64M     0   64M   0% /dev/shm
test-pool/pvc-fe8c7583-9896-4ae8-9e59-0393f0858279  9.7G  215M  9.4G   3% /var/lib/mysql
tmpfs                                               3.9G   12K  3.9G   1% /run/secrets/kubernetes.io/serviceaccount
tmpfs                                               3.9G     0  3.9G   0% /proc/acpi
tmpfs                                               3.9G     0  3.9G   0% /proc/scsi
tmpfs                                               3.9G     0  3.9G   0% /sys/firmware
```
**Note that:**
This above scenario is with only zfs file system. Didn't see any size mismatch with ext4/xfs file system

w3aman:
[...]

PR #57, opened 2020-03-10
fix(clone): setting properties on the clone volume

pawanpraka1:
fixes https://github.com/openebs/zfs-localpv/issues/55.

When we clone the zvol, it inherits volsize and block size from the snapshot from which this clone is created. This is not true for datasets, we have to set all the properties while creating the clone. Also, properties like compression, dedup will not be set on the cloned volume for both zvol and dataset. Here I am setting all the prperties for zvol and dataset while creating the clone volume.

### DIFF
--- a/pkg/zfs/zfs_util.go
+++ b/pkg/zfs/zfs_util.go
@@ -127,9 +127,41 @@ func buildCloneCreateArgs(vol *apis.ZFSVolume) []string {
 	ZFSVolArg = append(ZFSVolArg, ZFSCloneArg)
 
 	if vol.Spec.VolumeType == VOLTYPE_DATASET {
+		if len(vol.Spec.Capacity) != 0 {
+			quotaProperty := "quota=" + vol.Spec.Capacity
+			ZFSVolArg = append(ZFSVolArg, "-o", quotaProperty)
+		}
+		if len(vol.Spec.RecordSize) != 0 {
+			recordsizeProperty := "recordsize=" + vol.Spec.RecordSize
+			ZFSVolArg = append(ZFSVolArg, "-o", recordsizeProperty)
+		}
+		if vol.Spec.ThinProvision == "no" {
+			reservationProperty := "reservation=" + vol.Spec.Capacity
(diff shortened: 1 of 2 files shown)


pkg/zfs/zfs_util.go
```diff
@@ -127,9 +127,41 @@ func buildCloneCreateArgs(vol *apis.ZFSVolume) []string {
...
+		if len(vol.Spec.Capacity) != 0 {
+			quotaProperty := "quota=" + vol.Spec.Capacity
+			ZFSVolArg = append(ZFSVolArg, "-o", quotaProperty)
+		}
+		if len(vol.Spec.RecordSize) != 0 {
+			recordsizeProperty := "recordsize=" + vol.Spec.RecordSize
+			ZFSVolArg = append(ZFSVolArg, "-o", recordsizeProperty)
+		}
+		if vol.Spec.ThinProvision == "no" {
+			reservationProperty := "reservation=" + vol.Spec.Capacity
```
kmova: Is this the right spec value?

pawanpraka1: yes it is.

 reservation=size|none
       The minimum amount of space guaranteed to a dataset and its descendants.  When the amount of space used is below this value, the dataset is treated as if it
       were taking up the amount of space specified by its reservation.  Reservations are accounted for in the parent datasets' space used, and count against the
       parent datasets' quotas and reservations.

       This property can also be referred to by its shortened column name, reserv.

kmova: cool. trying to understand this a little more. 

Say: 

PV1 = 5G ( quota and reservation will be set to 5G)
PV1Clone1 = 5G (quota and reservation will be set to 5G) 
PV1Clone2 = 5G (quota and reservation will be set to 5G) 

Now, PV1Clone1 and PVClone2 will be considered as children of PV1. Will the PV1 run out of space if PV1Clone1 and PV1Clone2 start consuming more space?

pawanpraka1: clone is not a child of original volume. It will have it's separate quota and reservation.

Issue #56, opened 2020-03-10
Volume is not able to mount/attach in case of "xfs" file system due to duplicate UUID


w3aman:
**what happened:**
i provisioned volume with fstype as "xfs" in storageclass. Application was succesfully deployed. When i tried to clone the volume, pvc gets bound but application stuck in container creating state.
Giving error as below:
```Events:
  Type     Reason                  Age                   From                     Message
  Warning  FailedScheduling        <unknown>             default-scheduler        Failed to bind volumes: pv "pvc-aeadf930-2f7f-4e66-94d3-bd4366e0550b" node affinity doesn't match node "e2e1-node3": No matching NodeSelectorTerms
  Normal   Scheduled               <unknown>             default-scheduler        Successfully assigned aman/percona1-7c776f8558-l2vxk to e2e1-node1
  Warning  FailedScheduling        <unknown>             default-scheduler        AssumePod failed: pod 300233c3-1de6-42a4-87d2-82bba6c284a3 is in the cache, so can't be assumed
  Normal   SuccessfulAttachVolume  5m17s                 attachdetach-controller  AttachVolume.Attach succeeded for volume "pvc-aeadf930-2f7f-4e66-94d3-bd4366e0550b"
  Warning  FailedMount             3m14s                 kubelet, e2e1-node1      Unable to attach or mount volumes: unmounted volumes=[data-vol], unattached volumes=[data-vol default-token-tjqfq]: timed out waiting for the condition
  Warning  FailedMount             62s (x10 over 5m13s)  kubelet, e2e1-node1      MountVolume.SetUp failed for volume "pvc-aeadf930-2f7f-4e66-94d3-bd4366e0550b" : kubernetes.io/csi: mounter.SetupAt failed: rpc error: code = Internal desc = rpc error: code = Internal desc = not able to format and mount the zvol
  Warning  FailedMount             58s                   kubelet, e2e1-node1      Unable to attach or mount volumes: unmounted volumes=[data-vol], unattached volumes=[default-token-tjqfq data-vol]: timed out waiting for the condition
```
Storage class yaml is used as below:
```apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: zfs-sc
allowVolumeExpansion: true
parameters:
  volblocksize: "4k"
  compression: "off"
  dedup: "off"
  fstype: "xfs"
  poolname: "test-pool"
provisioner: zfs.csi.openebs.io
volumeBindingMode: WaitForFirstConsumer
[...]

PR #64, opened 2020-03-20
fix(xfs): clearing the xfs log before generating UUID

pawanpraka1:
xfs_admin command to generate the new UUID for the cloned
volume fails without returning error if there is log available
in the filesystem :

ERROR: The filesystem has valuable metadata changes in a log which needs to
be replayed.  Mount the filesystem to replay the log, and unmount it before
re-running xfs_admin.  If you are unable to mount the filesystem, then use
the xfs_repair -L option to destroy the log and attempt a repair.
Note that destroying the log may cause corruption -- please attempt a mount
of the filesystem before doing this.

No UUID will be generated in this case and application can not mount the volume.

Here mounting the filesystem to the temp location with "nouuid" mount option first
so that it can replay the logs first and system is in clean state and then unmount it
and after that generating the UUID with the xfs_admin command.

### DIFF
--- a/pkg/zfs/zfs_util.go
+++ b/pkg/zfs/zfs_util.go
@@ -364,12 +364,20 @@ func CreateClone(vol *apis.ZFSVolume) error {
 	if vol.Spec.FsType == "xfs" {
 		// for mounting the cloned volume for xfs, a new UUID has to be generated
 		device := ZFS_DEVPATH + volume
-		cmd := exec.Command("xfs_admin", "-U", "generate", device)
+		cmd := exec.Command("xfs_repair", "-L", device)
 		out, err := cmd.CombinedOutput()
+		if err != nil {
+			logrus.Errorf("zfspv: Clone XFS uuid xfs_repair failed error: %s", string(out))
+			return err
+		}
+
+		cmd = exec.Command("xfs_admin", "-U", "generate", device)
+		out, err = cmd.CombinedOutput()
 		if err != nil {
 			logrus.Errorf("zfspv: Clone XFS uuid generate failed error: %s", string(out))
 			return err
 		}
+		logrus.Infof("generated UUID for the cloned xfs volume %s \n %v", volume, string(out))
 	}
 	return nil
 }
--- a/pkg/zfs/zfs_util.go
+++ b/pkg/zfs/zfs_util.go
@@ -362,31 +362,7 @@ func CreateClone(vol *apis.ZFSVolume) error {
 	}
 
 	if vol.Spec.FsType == "xfs" {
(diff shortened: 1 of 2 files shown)


pkg/zfs/zfs_util.go
```diff
@@ -364,12 +364,20 @@ func CreateClone(vol *apis.ZFSVolume) error {
 	if vol.Spec.FsType == "xfs" {
 		// for mounting the cloned volume for xfs, a new UUID has to be generated
 		device := ZFS_DEVPATH + volume
-		cmd := exec.Command("xfs_admin", "-U", "generate", device)
+		cmd := exec.Command("xfs_repair", "-L", device)
```
kmova: Q_: while taking snapshot on a xfs mounted volume, is there something that can be done to flush the logs?

pawanpraka1: there are no options available as such, There is xfs_freeze option there but that is also not helping. So followed the solution which there are following for LVM snapshot :-
https://learn.redhat.com/t5/Platform-Linux/How-to-Mounting-an-XFS-LVM-snapshot/td-p/287

pawanpraka1: there is another solution which I have suggested on the bug(https://github.com/openebs/zfs-localpv/issues/56) was use below storageclass for xfs cloned volumes :-
```
kind: StorageClass
metadata:
  name: openebs-zfspv1
allowVolumeExpansion: true
parameters:
  fstype: "xfs"
  poolname: "zfspv-pool"
volumeBindingMode: WaitForFirstConsumer
mountOptions:
  - nouuid
```
As with nouuid it will not check for the UUID to mount the fs.
```
       nouuid Don't check for double mounted file systems using the file
              system uuid.  This is useful to mount LVM snapshot volumes,
              and often used in combination with "norecovery" for mounting
              read-only snapshots.
```

kmova: makes sense. one additional comment. Can we also include a comment in the code here as to why we are doing this?

pawanpraka1: After reading so much about this, I am really not sure what is the right fix for it. The only thing which worried me with the current fix is this line "Note that destroying the log may cause corruption -- please attempt a mount of the filesystem before doing this." Now I am thinking, should we ask user to use nouuid mount option for this in the storageclass for the cloned volumes, as with that it will replay the log.
```
mountOptions:
  - nouuid
```

kmova: thanks for working with the XFS community and coming with this solution of:
- temporary mount with no-uuid, 
- after the logs are replayed, generate new uuid and mount it.

PR #66, opened 2020-03-20
feat(validation): adding validation for ZFSPV CR parameters

pawanpraka1:
Validating few parameters for the ZFSVolume custom resource

- compression can be "on", "off", "lzjb", "gzip", "gzip-[1-9]", "zle" and "lz4"
- encryption can be "on", "off", "aes-128-ccm", "aes-192-ccm", "aes-256-ccm", "aes-128-gcm", "aes-192-gcm", and "aes-256-gcm"
- dedup can be "on" and "off"
- poolname can be string
- ownernodeid can be present
- thinprovision can be "yes" and "no"
- volumetype can be "DATASET" and "ZVOL"

Also added required fields needed to create ZFSVolume CR
- ownerNodeID
- poolname
- volumeType
- capacity

For the below wrong yaml
```yaml
apiVersion: openebs.io/v1alpha1
kind: ZFSVolume
metadata:
  name: pvc-1667faf6-d79b-4a5d-9811-2c4886127f67
  namespace: openebs
spec:
  capacity: "8589934592"
  compression: "yes"
  ownerNodeID: "pawan-2"
  poolName: "zfspv-pool"
  volumeType: "ZVOL"
```
Applying it will throw this error

$ kubectl apply -f zfspvcr.yaml 
The ZFSVolume "pvc-test" is invalid: spec.compression: Invalid value: "": spec.compression in body should match '^(on|off|lzjb|gzip|gzip-[1-9]|zle|lz4)$'


deploy/zfs-operator.yaml
```diff
@@ -22,6 +22,40 @@ spec:
     shortNames:
     - zfsvol
     - zv
+  validation:
+    openAPIV3Schema:
+      properties:
+        spec:
+          properties:
+            compression:
+              pattern: "^(on|off|lzjb|gzip|gzip-[1-9]|zle|lz4)$"
```
akhilerm: @pawanpraka1 is there a default value to these fields?

pawanpraka1: No @akhilerm, the way zfs works is if nothing is given while provisioning the volume it inherits the properties from pool. So here we don't need to have any default values.

kmova:
@pawanpraka1 -- any validations required for the ZFS Snapshot CRD?

pawanpraka1:
> @pawanpraka1 -- any validations required for the ZFS Snapshot CRD?

no @kmova, user will modify ZFSVolume CR only for the property change.

kmova:
Makes sense. on second thought, does adding validations for non-editable fields help in making sure the code also doesn't misbehave, can work as an added validation as the contributions to the project increase?

pawanpraka1:
Sure, I will the validation for snapshot also.

buildscripts/generate-manifests.sh
```diff
@@ -0,0 +1,80 @@
...
+
+echo '
+
+##############################################
+###########                       ############
+###########   ZFSVolume CRD       ############
+###########                       ############
+##############################################
+
+# ZFSVolume CRD is autogenerated via `make manifests` command.
```
kmova, line 45: Probably a note saying "DO NOT MODIFY THIS FILE" can be the first line before this description on how this file can be updated.

kmova: same comment applies to all auto generated yamls.

deploy/yamls/zfssnapshot-crd.yaml
```diff
@@ -104,10 +115,13 @@ spec:
                 volume should be cloned
               type: string
             thinProvision:
-              description: Thinprovision specifies if we should thin provisioned the
-                volume or not
+              description: ThinProvision describes whether we have reserved the space
```
kmova, line 118: what is the default value, if not specified by user? 

"yes" => volume can be provisioned even if the ZPOOL doesn't have the required capacity. 
"no"  => volume can be provisioned only if the ZPOOL has enough capacity and capacity required by volume can be reserved.

kmova: Btw, could not comment on the unchanged lines above. Please check the description for the `properties` field.  It says volumeinfo. Also as these are snapshots, it is probably worth mentioning how values are populated.

also it will be good to indicate if a given attribute is editable or only set during the creation time.

kmova: n_: compression description has a grammar error.

upgrade/crd.yaml
```diff
@@ -1,68 +1,362 @@
...
+          type: string
+        kind:
+          description: 'Kind is a string value representing the REST resource this
+            object represents. Servers may infer this from the endpoint the client
+            submits requests to. Cannot be updated. In CamelCase. More info: https://git.k8s.io/community/contributors/devel/api-conventions.md#types-kinds'
+          type: string
+        metadata:
+          type: object
+        spec:
+          description: VolumeInfo contains the volume related info for all types of
```
kmova, line 83: The description for VolumeInfo could be changed to something like: spec.Properties defines ZFS volume parameters for all modes in which ZFS volumes can be created like - ZFS volume with filesystem, ZFS Volume exposed as zfs or ZFS volume exposed as raw block device. Some of the parameters can be only set during creation time (as specified in the details of the parameter), and a few are editable. In case of Cloned volumes, the parameters are assigned the same values as the source volume.

upgrade/crd.yaml
```diff
@@ -1,68 +1,362 @@
...
+            volumes in ZFSVolume
+          properties:
+            capacity:
+              description: Capacity of the volume
+              minLength: 1
+              type: string
+            compression:
+              description: Controls the compression algorithm used for this dataset.
+                Compression specifies if the it should enabled on the zvol. Setting
```
kmova, line 92: n_ : `if the it`

upgrade/crd.yaml
```diff
@@ -1,68 +1,362 @@
...
+                and implemented appropriate recovery practices, such as regular backups.
+                As an alternative to deduplication consider using compression=lz4,
+                as a less resource-intensive alternative. should be enabled on the
+                zvol
+              enum:
+              - "on"
+              - "off"
+              type: string
+            encryption:
+              description: Enabling the encryption feature allows for the creation
```
kmova, line 117: for each attribute it will be good to specify what is the default value in the description.

upgrade/crd.yaml
```diff
@@ -1,68 +1,362 @@
...
+            fsType:
+              description: FsType specifies filesystem type for the zfs volume/dataset
+              type: string
+            keyformat:
+              description: KeyFormat specifies format of the encryption key
+              type: string
+            keylocation:
+              description: KeyLocation is the location of key for the encryption
+              type: string
+            ownerNodeID:
```
kmova, line 135: What is this field?

upgrade/crd.yaml
```diff
@@ -1,68 +1,362 @@
...
+                metadata related to the pool structure, including dataset and snapshot
+                names, dataset hierarchy, properties, file size, file holes, and deduplication
+                tables (though the deduplicated data itself is encrypted).
+              pattern: ^(on|off|aes-128-[c,g]cm|aes-192-[c,g]cm|aes-256-[c,g]cm)$
+              type: string
+            fsType:
+              description: FsType specifies filesystem type for the zfs volume/dataset
+              type: string
+            keyformat:
+              description: KeyFormat specifies format of the encryption key
```
kmova, line 130: what are some of the supported keyformats?

upgrade/crd.yaml
```diff
@@ -1,68 +1,362 @@
...
+                of encrypted filesystems and volumes. ZFS will encrypt file and zvol
+                data, file attributes, ACLs, permission bits, directory listings,
+                FUID mappings, and userused / groupused data. ZFS will not encrypt
+                metadata related to the pool structure, including dataset and snapshot
+                names, dataset hierarchy, properties, file size, file holes, and deduplication
+                tables (though the deduplicated data itself is encrypted).
+              pattern: ^(on|off|aes-128-[c,g]cm|aes-192-[c,g]cm|aes-256-[c,g]cm)$
+              type: string
+            fsType:
+              description: FsType specifies filesystem type for the zfs volume/dataset
```
kmova, line 127: any default value, probably can get into details of what happens if filesystem is specified as zfs vs xfs/ext4.

upgrade/crd.yaml
pawanpraka1: added description.

buildscripts/generate-manifests.sh
pawanpraka1: this line will go the zfs-operator file also. Might confuse the user who is usng it. Feeling better to not put that line?

deploy/yamls/zfssnapshot-crd.yaml
```diff
@@ -35,20 +46,26 @@ spec:
...
-                should be used. The current default compression algorithm is either
-                lzjb or, if the lz4_compress feature is enabled, lz4. Changing this
-                property affects only newly-written data.
+                Compression specifies if it should be enabled on the zvol/dataset.
+                Setting compression to "on" indicates that the current default compression
+                algorithm should be used. The current default compression algorithm
+                is either lzjb or, if the lz4_compress feature is enabled, lz4. Changing
+                this property affects only newly-written data. By default compression
+                is set to "off". This attribute is editable, we can update this attribute
+                to the desired compression property later on once volume has provisioned.
```
kmova, line 68: n_: The sentences should be framed using a third-person tone. Avoid using any kind of pronouns like he/she/we/it. 

This description could be changed to the following:
```
Compression specifies the block-level compression algorithm to be applied to the ZFS Volume. 
The value "on" indicates ZFS to use the default compression algorithm. The default compression
 algorithm used by ZFS will be either lzjb or, if the lz4_compress feature is enabled, lz4.
 
Compression property can be edited after the volume has been created.  The change will only
be applied to the newly-written data. For instance, if the Volume was created with "off" and 
the next day the compression was modified to "on", the data written prior to setting "on" will
not be compressed. 

Default Value: Off. 
```

PR #70, opened 2020-03-23
feat(crd): refactoring CRD

pawanpraka1:
Changed the group name from openebs.io to zfs.openebs.io.

Now ZFS Volume CR will look like this : 
```
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: zfszvolumes.zfs.openebs.io
spec:
  group: zfs.openebs.io
  version: v1alpha1
  scope: Namespaced
  names:
    plural: zfsvolumes
    singular: zfsvolume
    kind:ZFSVolume
    shortNames:
    - zfsvol
    - zv
```

Snapshot CR will look like this :
```
apiVersion: apiextensions.k8s.io/v1beta1
kind: CustomResourceDefinition
metadata:
  name: zfssnapshots.zfs.openebs.io
spec:
  group: zfs.openebs.io
  version: v1alpha1
  scope: Namespaced
  names:
    plural: fssnapshots
    singular: zfssnapshot
    kind: ZFSSnapshot
    shortNames:
    - zfssnapshot
    - zfssnap

```


kmova:
@pawanpraka1 - Can you confirm that the changes are mainly the following:
=> group changes from `openebs.io` to `zfs.openebs.io`
=> Removed the prefix of ZFS from the CRD names

There is no change to the short names, from earlier versions to this version. 

While I see the reasoning behind cutting the prefix (which seems redundant), I am wondering if we should continue to have them for clarity.  When using kubectl, only full name, and the short names matter. 

This PR takes care of short-names being unique. I think we should maintain the same for full name.  Can we revert the changes done to CRD names and only keep the group change.

pawanpraka1:
Yes @kmova, this PR does two things :-
=> group changes from openebs.io to zfs.openebs.io
=> Removed the prefix of ZFS from the CRD names

> I think we should maintain the same for full name.

@kmova, the full name format will change anyway as we are changing the group name since  full name format depends on group name
fullname == <plural-name><group-name>.

So we are changing CRD from zfsvolumes.openebs.io to volumes.zfs.openebs.io.

sonasingh46:
@pawanpraka1  According to the PR message : 
```
=> Removed the prefix of ZFS from the CRD names
```
I could not find this change or I could be interpreting it wrong. Can you please help here?

PR #76, opened 2020-04-06
feat(zfspv): handling unmounted volume .

pawanpraka1:
There can be cases where openebs namespace has been accidentally deleted. If that happens, the volume CRs deletion will be triggered. Volume CR deletion will trigger dataset deletion process. Prior to the actual deletion of the data, the driver will attempt to do the following:
- Unmount the dataset(not the case with zvol as it will be unmounted via NodeUnPublish) Unmounting the dataset will Set the mount="no"
- Delete the zvol or dataset

But since the volume is actively consumed by a pod, the destroy will fail, as the volume is busy. The zvols continue to operate. However with datasets, it is umounted resulting in the setting of mount="no", but then delete will fail.

Now, there are two actions that the user can take:
(a) Continue to clean-up, so that setup can be re-created
(b) Reinstall openebs and try to create volumes and point to the underlying datasets. 

In case (a), the user will have to go delete the application pods and then start deleting the PVCs. However, the PV deletion will fail, because the steps to clean up do not find that volume is mounted and will be aborted. 

Let us assume case (b), and prior to actually reinstalling, as the volumes are still intact, applications are expected to continue to access the data. 

However, if a node restart occurs, due to the mount being set to no, the pods will not be able to access the volume. The data stored by the pod will not persist as it is not backed by persistence storage.

 To recover from the partial clean up steps. the following needs to be done on each of the nodes:
- `zfs get mounted` : check if there is any unmounted dataset with this option as "no". 
- For all the datasets that showed mounted as no, do the following:
 `zfs mount <dataset name>`
  The above commands will result in mounting the dataset. 

Here in this PR :
- automating the manual steps performed above to check that volume can be mounted, even if a manual operation left it in a non-mountable state. 
- helping with the case (a), by into NodeUnPublish operation - that continues to destroy in whatever state the volume may be in. 
- add helpful debug messages. 

### DIFF
--- a/pkg/driver/agent.go
+++ b/pkg/driver/agent.go
@@ -133,10 +133,8 @@ func (ns *node) NodeUnpublishVolume(
 ) (*csi.NodeUnpublishVolumeResponse, error) {
 
 	var (
-		err           error
-		vol           *apis.ZFSVolume
-		devpath       string
-		currentMounts []string
+		err error
+		vol *apis.ZFSVolume
 	)
 
 	if err = ns.validateNodeUnpublishReq(req); err != nil {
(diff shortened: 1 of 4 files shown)


pkg/driver/agent.go
```diff
@@ -147,37 +145,17 @@ func (ns *node) NodeUnpublishVolume(
...
-		goto NodeUnpublishResponse
-	}
-
-	currentMounts, err = zfs.GetMounts(devpath)
-	if err != nil {
-		return nil, err
-	} else if len(currentMounts) == 0 {
-		return nil, status.Error(codes.Internal, "umount request for not mounted volume")
-	} else if len(currentMounts) == 1 {
-		if currentMounts[0] != targetPath {
```
kmova, line 163: I think this condition check went away with the refactored code. Is this condition not required anymore?

pawanpraka1: We are already checking in UmountVolume that if volume is mounted at the targetpath then only unmount it, otherwise we should always return 0 OK.

Issue #79, opened 2020-04-11
Pod creation problems due to slow mounting/attaching


spencergilbert:
We have 28 total nodes (3 masters, 25 worker)
Each node has 48 cores, 252GB ram

~228 zfs volumes (305 including old path volumes)
~212 volumeattachments

```I0409 14:22:42.084575 1 reconciler.go:288] attacherDetacher.AttachVolume started for volume "pvc-f029b12f-92d6-44fa-aa28-33a64f420b34" (UniqueName: "kubernetes.io/csi/zfs.csi.openebs.io^pvc-f029b12f-92d6-44fa-aa28-33a64f420b34") from node "halfling-mexico"
    I0409 17:26:53.814862 1 operation_generator.go:359] AttachVolume.Attach succeeded for volume "pvc-f029b12f-92d6-44fa-aa28-33a64f420b34" (UniqueName: "kubernetes.io/csi/zfs.csi.openebs.io^pvc-f029b12f-92d6-44fa-aa28-33a64f420b34") from node "halfling-mexico"```

Volumes provisioned by the zfs provider all seem to take >1hr to complete the mounting of a volume. We haven't seen any throttling logs form the controller manager.

Happy to provide any additional information.

**Comment by pawanpraka1** [comment]
Analysis so far => The problem seems to be related to k8s, ZFS-LocalPV seems to be fine. K8s services are taking too long to create the volumeattachment object, which will attach the volume to the node. the pod will be stuck in ContainerCreating state, until volumeattachment object has been created. Once it is created the pod will come into running state.

**Comment by spencergilbert** [comment]
following this https://github.com/kubernetes/kubernetes/issues/84169#issuecomment-545692734 setting attach-detach-reconcile-sync-period to 300s improves the situation. Perhaps LIST_VOLUMES_PUBLISHED_NODES should be implemented?

**Comment by pawanpraka1** [comment]
```
I don't think LIST_VOLUMES_PUBLISHED_NODES helps reduce API load. One possible scaling problem that was pointed out earlier is that we do a Get() on every VolumeAttachment every minute. We should change that to use an informer instead. This should greatly reduce API load.
```
Discussed here :- https://github.com/kubernetes/kubernetes/issues/84169

**Comment by w3aman** [comment]
I reproduced this issue with 200s of volumes. i saw that when this no increases Volume attachment is getting very slow. I've some stats regarding this timings.
```
first 50 pods   ----> 43 seconds
next 50   ---------> 1 min 11 sec
[...]

PR #85, opened 2020-04-17
feat(attach): avoid creation of volumeattachment object

pawanpraka1:
fixes :- https://github.com/openebs/zfs-localpv/issues/79

k8s is very slow in attaching the volumes when dealing with the 
large number of volume attachment object.

(k8s issue https://github.com/kubernetes/kubernetes/issues/84169)

The volumeattachment is not required for ZFSPV, so avoid creation
of attachment object, also removed the csi-attacher container as
this is also not needed as it acts on volumeattachment object.

k8s is very slow in attaching the volumes when dealing with the 
large number of volume attachment object : 

k8s issue https://github.com/kubernetes/kubernetes/issues/84169).

Volumeattachment is a CR created just to tell the watcher of it
which is csi-attacher, that it has to call the Controller Publish/Unpublish grpc.
Which does all the tasks to attach the volumes to a node for example call to the 
DigitalOcean Block Storage API service to attach a created volume to a specified node.
Since for ZFSPV, volume is already present locally, nothing needs to done in Controller
Publish/Unpublish, so it is good to remove them.

so avoiding creation of attachment object in this change, also removed the csi-attacher
container as this is also not needed as it acts on volumeattachment object.

Removed csi-cluster-driver-registrar container also as it is deprecated and not needed anymore.

We are using csidriver beta CRDs so minimum k8s version required is 1.14+.

### DIFF
--- a/deploy/yamls/zfs-driver.yaml
+++ b/deploy/yamls/zfs-driver.yaml
@@ -1,6 +1,19 @@
 
 ---
 
+# Create the CSI Driver object
+apiVersion: storage.k8s.io/v1beta1
+kind: CSIDriver
+metadata:
+  name: zfs.csi.openebs.io
+spec:
+  # do not require volumeattachment
+  attachRequired: false
+  podInfoOnMount: false
+  volumeLifecycleModes:
+  - Persistent
+---
+
 apiVersion: apiextensions.k8s.io/v1beta1
 kind: CustomResourceDefinition
 metadata:
@@ -463,6 +476,9 @@ rules:
   - apiGroups: ["coordination.k8s.io"]
     resources: ["leases"]
     verbs: ["get", "watch", "list", "delete", "update", "create"]
+  - apiGroups: [""]
+    resources: ["nodes"]
+    verbs: ["get", "list", "watch"]
   - apiGroups: ["*"]
     resources: ["zfsvolumes", "zfssnapshots"]
     verbs: ["*"]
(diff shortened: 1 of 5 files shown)


upgrade/cleanup.sh
```diff
@@ -2,27 +2,53 @@
...
-# remove the finalizer from the old CR
-sed -i "/zfs.openebs.io\/finalizer/d" volumes.yaml
-kubectl apply -f volumes.yaml
+echo "Fetching ZFS Volumes"
+numVol=`kubectl get zfsvolumes.openebs.io --no-headers -n $ZFSPV_NAMESPACE | wc -l`
 
-# delete the old CR
-kubectl delete -f volumes.yaml
+if [ $numVol -gt 0 ]; then
+	echo "Cleaning the ZFS Volumes(`expr $numVol - 1`)"
```
w3aman, line 16: why $numVol - 1  ??

pawanpraka1: earier I was not using --no-headers, which will give one extra count, so added -1 here. Now added  --no-headers but forgot to fix this message.  Thanks @w3aman .fixed it

deploy/yamls/zfs-driver.yaml
```diff
@@ -543,7 +559,7 @@ spec:
           image: quay.io/k8scsi/snapshot-controller:v2.0.1
           args:
             - "--v=5"
-            - "--leader-election=false"
+            - "--leader-election=true"
```
kmova, line 562: why is this required?

pawanpraka1: needed if HA customer is using HA.

Issue #84, opened 2020-04-16
add support to have custom topology key for ZFSPV


pawanpraka1:

pawanpraka1:
fixed via https://github.com/openebs/zfs-localpv/pull/94.

PR #94, opened 2020-04-27
feat(topology): adding support for custom topology keys

pawanpraka1:
fixes :- https://github.com/openebs/zfs-localpv/issues/84

Now user can label the nodes with the required topology, the ZFSPV
driver will support all the node labels as topology keys.

We should label the nodes first and then deploy the driver to make it aware of
all the labels that node has. If we want to add labels after ZFS-LocalPV driver
has been deployed, a restart all the node agents are required so that the driver
can pick the labels and add them as supported topology keys.

Note that if storageclass is using Immediate binding mode and topology key is not mentioned then all the nodes should be labeled using same key, that means, same key should be present on all nodes, nodes can have different values for those keys. If nodes are labeled with different keys i.e.
some nodes are having different keys, then ZFSPV's default scheduler can not effictively
do the volume count based scheduling. In this case the CSI provisioner will pick keys from
any random node and then prepare the preferred topology list using the nodes which has those
keys defined. And ZFSPV scheduler will schedule the PV among those nodes only.

Changed the csi-provisioner image to 1.6.0 as this has the upstream fix https://github.com/kubernetes-csi/external-provisioner/pull/421.

### DIFF
--- a/docs/faq.md
+++ b/docs/faq.md
@@ -67,7 +67,7 @@ The above storage class tells that ZFS pool "zfspv-pool" is available on nodes z
 Please note that the provisioner name for ZFS driver is "zfs.csi.openebs.io", we have to use this while creating the storage class so that the volume provisioning/deprovisioning request can come to ZFS driver.
 
 
-### 3. How to install the provisioner in HA
+### 5. How to install the provisioner in HA
 
 To have HA for the provisioner(controller), we can update the replica count to 2(or more as per need) and deploy the yaml. Once yaml is deployed, you can see 2(or more) controller pod running. At a time only one will be active and once it is down, the other will take over. They will use lease mechanism to decide who is active/master. Please note that it has anti affinity rules, so on one node only one pod will be running, that means, if you are using 2 replicas on a single node cluster, the other pod will be in pending state because of the anti-affinity rule. So, before changing the replica count, please make sure you have sufficient nodes.
 
@@ -89,7 +89,7 @@ spec:
 ---
 ```
 
-### 4. How to add custom topology key
+### 6. How to add custom topology key
 
 To add custom topology key, we can label all the nodes with the required key and value :-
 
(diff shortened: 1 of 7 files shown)


pkg/driver/agent.go
```diff
@@ -171,7 +172,16 @@ func (ns *node) NodeGetInfo(
...
-	topology := map[string]string{zfs.ZFSTopologyKey: ns.driver.config.NodeID}
+	node, err := k8sapi.GetNode(ns.driver.config.NodeID)
+	if err != nil {
+		logrus.Errorf("failed to get the node %s", ns.driver.config.NodeID)
+		return nil, err
+	}
+
+	// support all the keys that node has
+	topology := node.Labels
+	topology[zfs.ZFSTopologyKey] = ns.driver.config.NodeID
```
kmova, line 183: An comment here with example labels will be good.

pawanpraka1: added comment.

kmova:
Changes look good. 

Couple of doc items. Can the release notes be added as per the new guidelines and also, can the readme/examples updated on how to use this feature.

docs/faq.md
```diff
@@ -88,3 +88,31 @@ spec:
...
+$ kubectl get pods -n kube-system -l role=openebs-zfs
+
+NAME                       READY   STATUS    RESTARTS   AGE
+openebs-zfs-controller-0   4/4     Running   0          5h28m
+openebs-zfs-node-4d94n     2/2     Running   0          5h28m
+openebs-zfs-node-gssh8     2/2     Running   0          5h28m
+openebs-zfs-node-twmx8     2/2     Running   0          5h28m
+```
+
+Note that if storageclass is using Immediate binding mode then all the nodes should be labeled using same key, that means, same key should be present on all nodes, nodes can have different values for those keys. If nodes are labeled with different keys i.e. some nodes are having different keys, then ZFSPV's default scheduler can not effictively do the volume count based scheduling. Here, in this case the CSI provisioner will pick keys from any random node and then prepare the preferred topology list using the nodes which has those keys defined and ZFSPV scheduler will schedule the PV among those nodes only.
```
kmova, line 118: Before this node, can you add an example StorageClass YAML that uses the above custom topology key. 

And describe how driver uses this custom key to schedule the PV.

docs/faq.md
```diff
@@ -88,3 +88,31 @@ spec:
...
+$ kubectl label node pawan-node-1 openebs.io/rack=rack1
+node/pawan-node-1 labeled
+
+$ kubectl get nodes pawan-node-1 --show-labels
+NAME           STATUS   ROLES    AGE   VERSION   LABELS
+pawan-node-1   Ready    worker   16d   v1.17.4   beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,kubernetes.io/arch=amd64,kubernetes.io/hostname=pawan-node-1,kubernetes.io/os=linux,node-role.kubernetes.io/worker=true,openebs.io/rack=rack1
+
+```
+
+Once we have labeled the node, we can install the zfs driver. The driver will pick the node labels and add that as the supported topology key. If the driver is already installed and you want to add a new topology information, you can label the node with the topology information and then restart of the nodes daemonset are required so that the driver can pick the labels and add them as supported topology keys. We should restart the pod in kube-system namespace with the name as openebs-zfs-node-[xxxxx] which is the node agent pod for the ZFS-LocalPV Driver.
```
kmova, line 106: n_: nodes daemonset -> zfs pv csi driver daemon set ( `openebs-zfs-node`)

docs/faq.md
pawanpraka1: Thanks @kmova, added sample storageclass.

PR #95, opened 2020-04-27
feat(yaml): update the v0.4.x yaml to the latest 0.4.1 tag

pawanpraka1:


kmova:
Probably not a must, but can we change the other image reference in the operator also to 0.4.1 
https://github.com/openebs/zfs-localpv/blob/7641fcfa38c8143696abccdf6d0a949d30f43e99/deploy/zfs-operator.yaml#L878

pawanpraka1:
> Probably not a must, but can we change the other image reference in the operator also to 0.4.1
it is must. I forgot to update it for node agent. Thanks. Fixed it.

PR #102, opened 2020-05-04
feat(block): adding block volume support for ZFSPV

pawanpraka1:
Fixes: https://github.com/openebs/openebs/issues/2971

We can create a Raw Block Volume request using volumemode as block in PVC :-

```
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: block-claim
spec:
  volumeMode: Block
  storageClassName: zfspv-block
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 5Gi
```
we can use the below storageclass for RAW block volume:
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: zfspv-block
allowVolumeExpansion: true
parameters:
  poolname: "zfspv-pool"
```
The driver will create a zvol for this volume and bind mount the block  device at the given path.

### DIFF
--- a/pkg/driver/agent.go
+++ b/pkg/driver/agent.go
@@ -110,14 +110,18 @@ func (ns *node) NodePublishVolume(
 
 	vol, mountInfo, err := GetVolAndMountInfo(req)
 	if err != nil {
-		goto PublishVolumeResponse
+		return nil, err
 	}
-	// attempt mount operation on the requested path
-	if err = zfs.MountVolume(vol, mountInfo); err != nil {
-		goto PublishVolumeResponse
+	// If the access type is block, do nothing for stage
+	switch req.GetVolumeCapability().GetAccessType().(type) {
+	case *csi.VolumeCapability_Block:
+		// attempt block mount operation on the requested path
+		err = zfs.MountBlock(vol, mountInfo)
+	case *csi.VolumeCapability_Mount:
+		// attempt filesystem mount operation on the requested path
+		err = zfs.MountFilesystem(vol, mountInfo)
 	}
 
-PublishVolumeResponse:
 	if err != nil {
 		return nil, status.Error(codes.Internal, err.Error())
 	}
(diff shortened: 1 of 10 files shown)


pkg/driver/agent.go
```diff
@@ -110,14 +110,18 @@ func (ns *node) NodePublishVolume(
 
 	vol, mountInfo, err := GetVolAndMountInfo(req)
 	if err != nil {
-		goto PublishVolumeResponse
+		return nil, err
```
kmova, line 113: should the error be wrapped like it is being done below.

pkg/zfs/mount.go
```diff
@@ -208,11 +208,37 @@ func MountDataset(vol *apis.ZFSVolume, mount *apis.MountInfo) error {
...
 		return MountDataset(vol, mount)
 	default:
 		return MountZvol(vol, mount)
 	}
 }
+
+func MountBlock(vol *apis.ZFSVolume, mountinfo *apis.MountInfo) error {
+	target := mountinfo.MountPath
+	devicePath := ZFS_DEVPATH + vol.Spec.PoolName + "/" + vol.Name
+	mountopt := []string{"bind"}
```
kmova, line 223: Just noticed this. Not for this PR, but can you check what is the process of capturing the mountOptions specified via StorageClass and using them. Does MountInfo come with those parameters?

pkg/driver/agent.go
pawanpraka1: yes, we should, added.

pkg/zfs/mount.go
pawanpraka1: mount option has no meaning for block volume. Even if storage class has it, csi provisioner will pass it to the zfs plugin.

kmova: I saw this. https://github.com/openebs/zfs-localpv/blob/d501cc8a996b3db00e5935979ea44ee5f206add8/pkg/zfs/mount.go#L18 

Thanks!

Issue #111, opened 2020-05-12
pvc is getting bound even if the zpool name is given wrong


w3aman:
**What steps did you take and what happened:**
[A clear and concise description of what the bug is, and what commands you ran.]

A storage class was created with wrong zpool name. Now if we try to apply the pvc yaml with this storage class it gets bound status.
in `kubectl get zv -n openebs` command zpool name is also there which is the wrong one. I don't have any zpool with that name.

**What did you expect to happen:**
controller should verifiy if volume is really created or not. if not then pvc should not bound as we dont have that pool in which zfs dataset will be created.

**Anything else you would like to add:**
[Miscellaneous information that will assist in solving the issue.]

PR #121, opened 2020-05-19
feat(zfspv): pvc should be bound only if volume has been created.

pawanpraka1:
fixes: https://github.com/openebs/zfs-localpv/issues/111

The controller does not check whether the volume has been created or not
and return successful. Which in turn binds the pvc to the pv.

**What this PR does?**:

The PVC should not bound until corresponding zfs volume has been created.
Now controller will check the ZFSVolume CR state to be "Ready" before returning
successful. The CSI will retry the CreateVolume request when it will get
a error reply and when the ZFS node agent creates the ZFS volume and sets the
ZFSVolume CR state to be "Ready", the controller will return success for the
CreateVolume Request and then PVC will be bound.

**Checklist:**
- [x] Fixes #<issue number>
- [ ] Commit has unit tests
- [ ] (Optional) If documentation changes are required, which issue on


deploy/yamls/zfsvolume-crd.yaml
```diff
@@ -32,6 +32,10 @@ spec:
     description: Size of the volume
     name: Size
     type: string
+  - JSONPath: .status.state
+    description: Status of the volume
```
kmova, line 36: should we include the list of values that this can have in the description?

kmova: I see it below.. ignore.

deploy/yamls/zfsvolume-crd.yaml
```diff
@@ -216,6 +220,18 @@ spec:
           - poolName
           - volumeType
           type: object
+        status:
+          properties:
+            state:
+              description: State specifies the current state of the volume provisioning
```
kmova, line 226: :+1:

changelogs/unreleased/121-pawanpraka1
```diff
@@ -0,0 +1 @@
+enhancement to bound the pvc only if volume has been created
```
kmova, line 1: n_ : bound -> bind

kmova: Fixes an issue where ZFS PV with an invalid pool was bound to PVC.

pkg/driver/controller.go
```diff
@@ -205,6 +218,18 @@ func (cs *controller) CreateVolume(
...
 	}
 
+	_, state, err = zfs.GetZFSVolumeStatus(req.Name)
+
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "createvolume: failed to fetch the volume %v", err.Error())
+	}
+
+	if state == zfs.ZFSStatusPending {
+		return nil, status.Errorf(codes.Internal, "volume %s is being created", volName)
```
kmova, line 228: Can this be warning.

pkg/zfs/volume.go
```diff
@@ -157,6 +157,20 @@ func GetZFSVolume(volumeID string) (*apis.ZFSVolume, error) {
 	return vol, err
 }
 
+// GetZFSVolumeStatus returns ZFSVolume status
+func GetZFSVolumeStatus(volID string) (string, string, error) {
```
kmova, line 161: Need to add more comments/context for this function. 
what does the OwnerID contain? Probably should rename as it is returning OwnerID or may be add another function to get the OwnerID.

pkg/driver/controller.go
pawanpraka1: this is not the log. The is the error returned to the CSI provisioner so that it can retry the volume provisioning.

pkg/zfs/volume.go
pawanpraka1: updated the comment.

changelogs/unreleased/121-pawanpraka1
pawanpraka1: updated the description.

changelogs/unreleased/121-pawanpraka1
```diff
@@ -0,0 +1 @@
+Fixes an issue where PVC was bound to PV because of invalid/wrong parameters in PVC/Storageclass.
```
kmova, line 1: n_ : Fixes an issue where PVC was bound to invalid PV created using incorrect values provided in PVC/Storageclass. 

Need to find a better word for "invalid" to qualify this PV.

pkg/zfs/volume.go
kmova: Let us rename this to GetZFSVolumeState. Status usually means online/offline. Of course, the same argument can be made for State as well. See if there is a better name, if not resolve this comment.

changelogs/unreleased/121-pawanpraka1
pawanpraka1: made it unusable.
Fixes an issue where PVC was bound to unusable PV created using incorrect values provided in PVC/Storageclass.

PR #124, opened 2020-05-20
fix(zfspv): changing image pull policy to IfNotPresent

pawanpraka1:

### DIFF
--- a/deploy/yamls/zfs-driver.yaml
+++ b/deploy/yamls/zfs-driver.yaml
@@ -551,7 +551,6 @@ spec:
           env:
             - name: ADDRESS
               value: /var/lib/csi/sockets/pluginproxy/csi.sock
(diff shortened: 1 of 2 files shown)


deploy/yamls/zfs-driver.yaml
```diff
@@ -724,6 +723,7 @@ spec:
       containers:
         - name: csi-node-driver-registrar
           image: quay.io/k8scsi/csi-node-driver-registrar:v1.2.0
+          imagePullPolicy: IfNotPresent
```
kmova, line 726: We should probably remove this line.. as default will be `IfNotPresent` ?

pawanpraka1: that is correct. other places specifically mention "IfNotPresent" for imagePullPolicy. So made it in sync with other places.

Issue #119, opened 2020-03-02
Support for deploying to CentOS


zimmertr:
## Description

The `zfs-Operator` manifest is incompatible with ZFS. After deployment some hostPath references to libraries fail.

```
Warning  FailedMount  20m (x5 over 20m)    kubelet, localhost.localdomain  MountVolume.SetUp failed for volume "libzpool" : hostPath type check failed: /lib/libzpool.so.2.0.0 is not a file
  Warning  FailedMount  20m (x5 over 20m)    kubelet, localhost.localdomain  MountVolume.SetUp failed for volume "libnvpair" : hostPath type check failed: /lib/libnvpair.so.1.0.1 is not a file
```

## Context
I use CentOS/RHEL in production not Ubuntu.

## Possible Solution
This [manifest](https://gist.github.com/zimmertr/e177f4db6cea1b1960af32db210b384b) with modified libraries was sent to me via Slack and is deploying successfully.

jlcox1970:
I am also facing issue on CentOS 8 nodes with this driver.
I see the following int the openebs-zfs-plugin container logs

time="2020-05-21T23:39:27Z" level=error msg="zfs: could not set mountpoint on dataset main/pvc-b27284ae-3033-49e3-9c18-89f0c4895fcb cmd [set mountpoint=/var/lib/kubelet/pods/e1d092c2-7120-4dd8-a0de-4b523f19b972/volumes/kubernetes.io~csi/pvc-b27284ae-3033-49e3-9c18-89f0c4895fcb/mount main/pvc-b27284ae-3033-49e3-9c18-89f0c4895fcb] error: zfs: error while loading shared libraries: libssl.so.1.1: cannot open shared object file: No such file or directory\n"
22/05/2020 09:39:27

zimmertr:
Please use the manifests and image described in t[his comment](https://github.com/openebs/zfs-localpv/issues/119#issuecomment-598032945) by @pawanpraka1 until this has been resolved upstream.

jlcox1970:
I have used that image and the also the SC yaml that is also mentioned.
This got me to the point of no other errors in the logs apart from the above.

jlcox1970:
finally worked this out, the image is CentOS7 based...
it has libc-2.17 and all the libs that go with that
however, I am on CentOS8 and that a libc-2.28

I can update the docker image but I need to know how to run the make file to get the zfs-driver

alternatively can I just copy that out of the existing container and into the the new one ?

[...]

Issue #72, opened 2020-03-26
Auto provisioning with k3s


jemershaw:
I am trying to use k3s to deploy the zfs-localpv but it seems like everytime I try to deploy https://github.com/openebs/zfs-localpv/blob/master/deploy/sample/fio.yaml I have to manually create the dataset on the zpool. I'm not sure if this is due to k3s removing most of the extra bits k8s have in it or not.

I also tried changing the provisioner to zfs-localpv vs zfs.csi.openebs.io and that didn't change anything. If this should work then I can try to debug further with my configuration.

jemershaw:
I believe this is a zfs root problem because it can't find /sbin/zfs even though the md5sum is correct it just can't execute the binary.

jemershaw:
It isn't a zfs root problem -- This is a glibc difference -- the container is using libc-2.27 but my host is using 2.28. Only solution around this for me would be to build the images using a different container.

pawanpraka1:
@jemershaw , I guess you are facing this issue : https://github.com/openebs/zfs-localpv/issues/67. There is already PR for this https://github.com/openebs/zfs-localpv/pull/68. Just to confirm can you paste o/p of command "ldd `which zfs`". Also could you let us know what is your use case? How do you want to use zfs-locapv, what the setup and other details in this bug?

jemershaw:
@pawanpraka1 Thanks -- That partially would fix my problem. I'm running void linux and it seems like the correct volume mounts would have to be like below.

```
[...]

PR #134, opened 2020-05-26
feat(centos): adding support for deploying ZFS-LocalPV on CentOS machine

pawanpraka1:
fixes: https://github.com/openebs/zfs-localpv/issues/119, https://github.com/openebs/zfs-localpv/issues/72

Testing:
Verified: Ubuntu 18.04, 16.04, centsos7, centos8,
TODO:  k3os and any other

Since we mount the zfs host binary into the ZFS-LocalPV container, it needs to have all the compatible libraries available inside the container. With every relaese of Ubuntu, we have to change the docker base (ubuntu version) so that all the required library versions(for example libc, libssl and others which are not mounted from the host) are available. As the newer version of library will be backward compatible, so this will work for old ubuntu version also. But there is a pain of changing the docker base with every release of ubuntu version.

Approaching this issue in a different way and mounting the libraries (/lib and /lib64) into the conatiner and setting the LD_LIBRARY_PATH so that we search in that directory for the required libraries. So, with every ubuntu release, the new libraries will be mounted and we don't need to change the ubuntu base version with every release of it.

Also, this way, we can add support for centos also, as we are mounting /lib and /lib64, the centos libraries will also be mounted. Here, as the zfs binrary path is different(/usr/sbin/zfs) in centos, we are mounting it with FileOrCreate, so that we can use the same yaml for the OS where
the path is /sbin/zfs.

### DIFF
--- a/deploy/yamls/zfs-driver.yaml
+++ b/deploy/yamls/zfs-driver.yaml
@@ -783,10 +783,10 @@ spec:
               mountPath: /bin/zfs
               readOnly: true
             - name: hostlib
-              mountPath: /lib/host
+              mountPath: /lib/tls
               readOnly: true
             - name: hostlib64
-              mountPath: /lib64/host
+              mountPath: /lib/x86_64
               readOnly: true
             - name: pods-mount-dir
               mountPath: /var/lib/kubelet/
--- a/deploy/yamls/zfs-driver.yaml
+++ b/deploy/yamls/zfs-driver.yaml
@@ -776,20 +776,18 @@ spec:
               mountPath: /plugin
             - name: device-dir
               mountPath: /dev
-            - name: encr-keys
-              mountPath: /home/keys
             - name: zfs-bin
               mountPath: /sbin/zfs
-            - name: libzpool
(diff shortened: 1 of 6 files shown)


kmova:
This approach looks promising. Can we list the different platforms where this approach has been verified?

pawanpraka1:
@kmova, still testing needs to done (need help, asked few users to help me verifying this). I have verified on ubuntu only. Will come back to this PR once I am done with the api migration changes.

deploy/yamls/zfs-driver.yaml
```diff
@@ -776,20 +776,18 @@ spec:
               mountPath: /plugin
             - name: device-dir
               mountPath: /dev
-            - name: encr-keys
```
kmova, line 779: q_: what is the reason for removing this? is this not being used?

pawanpraka1: No, this is not used.

pawanpraka1:
closing this PR as mounting host's lib is little risky.

Issue #69, opened 2020-03-21
Invest on ZFS + NFS combination


stevefan1999-personal:
Ditto as title, we could use ZFS and NFS with great experience.
Conceptual code:
```bash
# # Creates a ZFS dataset that automatically shares a NFS mount on /rpool/openebs/zfs-root
# zfs create -o sharenfs="@10.42.0.0/16,no_root_squash" rpool/openebs/zfs-root
# # Upsert any CIDRs required on the sharenfs options.
```

orville-wright:
After multiple years of analyzing the NFS market and use-case within K8s, we feel that it's very unclear that NFS is a major use-case and significant consumer of K8s storage services for large enterprise storage applications in K8s. It seems that NFS users are the least understood of all storage users and we don't know what use-cases are driving NFS working k8s.

sinhaashish:
As @orville-wright  tagged it as wont fix, i am closing this issue. Please feel free to reopen.

PR #139, opened 2020-05-30
[WIP] make zfs provisioner set sharenfs property

cruwe:

With issue #69 , sharing via NFS has been requested. Setting the sharenfs property is easy. However, to actually use the property on a system with the nfs-kernel-server, the dataset needs to be mounted and probably shared calling `zfs share -a`.

When invoked locally on a shell, it works. From the provisioner, it doesn't.

**What this PR does?**:

See above

I have not tested intensively, I believe it doesn"t.

- create new dataset with sharenfs set
- change dataset's sharenfs properties
- destroy dataset

The PR is not yet ready to be merged. I kindly request help on how to properly mount the dataset on the providing machine, so that it may be exported via NFS.

In the second step, I would probably need assistance on how mount the dataset via nfs without fighting the affinity scheduler.

In both cases, I think pointing out the interface would be sufficient :-)

In any case, thank you very much for developing the zfs local provisioner,

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

**PLEASE REMOVE BELOW INFORMATION BEFORE SUBMITTING** (which I would love to do when it is ready, which it is not - by far)

The PR title message must follow convention:
   `<type>(<scope>): <subject>`.

Where: <br />
- `type` is defining if release will be triggering after merging submitted changes, details in [CONTRIBUTING.md](../CONTRIBUTING.md).
    Most common types are:
    - `feat`      - for new features, not a new feature for build script
    - `fix`       - for bug fixes or improvements, not a fix for build script
    - `chore`     - changes not related to production code
    - `docs`      - changes related to documentation
    - `style`     - formatting, missing semi colons, linting fix etc; no significant production code changes
    - `test`      - adding missing tests, refactoring tests; no production code change
    - `refactor`  - refactoring production code, eg. renaming a variable or function name, there should not be any significant production code changes

- `scope` is a single word that best describes where the changes fit.
  Most common scopes are like:
    - data engine (`localpv`, `jiva`, `cstor`)
    - feature (`provisioning`, `backup`, `restore`, `exporter`)
    - code component (`api`, `webhook`, `cast`, `upgrade`)
    - test (`tests`, `bdd`)
    - chores (`version`, `build`, `log`, `travis`)

- `subject` is a single line brief description of the changes made in the pull request.

### DIFF
--- a/deploy/yamls/ubuntu/zfs-driver.yaml
+++ b/deploy/yamls/ubuntu/zfs-driver.yaml
@@ -526,6 +526,14 @@ spec:
                 values:
                 - openebs-zfs-controller
             topologyKey: "kubernetes.io/hostname"
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: hosttype
+                    operator: In
+                    values:
+                      - zfsprovider
       priorityClassName: system-cluster-critical
       serviceAccount: openebs-zfs-controller-sa
       containers:
@@ -702,7 +710,6 @@ roleRef:
   apiGroup: rbac.authorization.k8s.io
 
 ---
-
 kind: DaemonSet
 apiVersion: apps/v1
 metadata:
@@ -712,12 +719,22 @@ spec:
   selector:
     matchLabels:
       app: openebs-zfs-node
+      role: openebs-zfs
   template:
     metadata:
       labels:
         app: openebs-zfs-node
         role: openebs-zfs
     spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: hosttype
+                    operator: In
+                    values:
+                      - zfsprovider
       priorityClassName: system-node-critical
       serviceAccount: openebs-zfs-node-sa
       hostNetwork: true
@@ -791,11 +808,42 @@ spec:
               mountPath: /lib/libuutil.so.1
             - name: libnvpair
               mountPath: /lib/libnvpair.so.1
+            - name: exportfs
+              mountPath: /usr/sbin/exportfs
+            - name: mount-nfs
+              mountPath: /usr/sbin/mount.nfs
+            - name: rpc-nfsd
+              mountPath: /usr/sbin/rpc.nfsd
+            - name: rpc-mountd
+              mountPath: /usr/sbin/rpc.mountd
+            - name: libtirp
+              mountPath: /usr/lib/x86_64-linux-gnu/libtirpc.so.3
             - name: pods-mount-dir
               mountPath: /var/lib/kubelet/
+              mountPropagation: "Bidirectional"
               # needed so that any mounts setup inside this container are
               # propagated back to the host machine.
+            - name: libgssapi
+              mountPath: /usr/lib/x86_64-linux-gnu/libgssapi_krb5.so.2
+            - name: libkrb
+              mountPath: /usr/lib/x86_64-linux-gnu/libkrb5.so.3
+            - name: libkrbcrypto
+              mountPath: /usr/lib/x86_64-linux-gnu/libk5crypto.so.3
+            - name: libkrbsupport
+              mountPath: /usr/lib/x86_64-linux-gnu/libkrb5support.so.0
+            - name: libkeyutils
+              mountPath: /usr/lib/x86_64-linux-gnu/libkeyutils.so.1
+            - name: libwrap
+              mountPath: /usr/lib/x86_64-linux-gnu/libwrap.so.0
+            - name: exports
+              mountPath: /etc/exports
+            - name: nfs-state
+              mountPath: /var/lib/nfs
+            - name: tank
+              mountPath: /tank
               mountPropagation: "Bidirectional"
+              # needed so that any mounts setup inside this container are
+              # propagated back to the host machine.
       volumes:
         - name: device-dir
           hostPath:
@@ -829,6 +877,50 @@ spec:
           hostPath:
             path: /lib/libnvpair.so.1.0.1
             type: File
+        - name: exportfs
+          hostPath:
+            path: /usr/sbin/exportfs
+            type: File
+        - name: mount-nfs
+          hostPath:
+            path: /usr/sbin/mount.nfs
+            type: File
+        - name: rpc-nfsd
+          hostPath:
+            path: /usr/sbin/rpc.nfsd
+            type: File
+        - name: rpc-mountd
+          hostPath:
+            path: /usr/sbin/rpc.mountd
+            type: File
+        - name: libtirp
+          hostPath:
+            path: /usr/lib/x86_64-linux-gnu/libtirpc.so.3
+            type: File
+        - name: libgssapi
+          hostPath:
+            path: /usr/lib/x86_64-linux-gnu/libgssapi_krb5.so.2
+            type: File
+        - name: libkrb
+          hostPath:
+            path: /usr/lib/x86_64-linux-gnu/libkrb5.so.3
+            type: File
+        - name: libkrbcrypto
+          hostPath:
+            path: /usr/lib/x86_64-linux-gnu/libk5crypto.so.3
+            type: File
+        - name: libkrbsupport
+          hostPath:
+            path: /usr/lib/x86_64-linux-gnu/libkrb5support.so.0
+            type: File
+        - name: libkeyutils
+          hostPath:
+            path: /usr/lib/x86_64-linux-gnu/libkeyutils.so.1
+            type: File
+        - name: libwrap
+          hostPath:
+            path: /usr/lib/x86_64-linux-gnu/libwrap.so.0
+            type: File
         - name: registration-dir
           hostPath:
             path: /var/lib/kubelet/plugins_registry/
@@ -841,4 +933,225 @@ spec:
           hostPath:
             path: /var/lib/kubelet/
             type: Directory
+        - name: exports
+          hostPath:
+            path: /etc/exports
+            type: File
+        - name: nfs-state
+          hostPath:
+            path: /var/lib/nfs
+            type: Directory
+        - name: tank
+          hostPath:
+            path: /tank
+            type: Directory
+
 ---
+kind: DaemonSet
+apiVersion: apps/v1
+metadata:
+  name: openebs-zfs-nfs
+  namespace: kube-system
+spec:
+  selector:
+    matchLabels:
+      app: openebs-zfs-node
+      role: openebs-nfs
+  template:
+    metadata:
+      labels:
+        app: openebs-zfs-node
+        role: openebs-nfs
+    spec:
+      affinity:
+        nodeAffinity:
+          requiredDuringSchedulingIgnoredDuringExecution:
+            nodeSelectorTerms:
+              - matchExpressions:
+                  - key: hosttype
+                    operator: In
+                    values:
+                      - nfsconsumer
+      priorityClassName: system-node-critical
+      serviceAccount: openebs-zfs-node-sa
+      hostNetwork: true
+      containers:
+        - name: csi-node-driver-registrar
+          image: quay.io/k8scsi/csi-node-driver-registrar:v1.2.0
+          imagePullPolicy: IfNotPresent
+          args:
+            - "--v=5"
+            - "--csi-address=$(ADDRESS)"
+            - "--kubelet-registration-path=$(DRIVER_REG_SOCK_PATH)"
+          lifecycle:
+            preStop:
+              exec:
+                command: ["/bin/sh", "-c", "rm -rf /registration/zfs-localpv /registration/zfs-localpv-reg.sock"]
+          env:
+            - name: ADDRESS
+              value: /plugin/csi.sock
+            - name: DRIVER_REG_SOCK_PATH
+              value: /var/lib/kubelet/plugins/zfs-localpv/csi.sock
+            - name: KUBE_NODE_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+            - name: NODE_DRIVER
+              value: openebs-zfs
+          volumeMounts:
+            - name: plugin-dir
+              mountPath: /plugin
+            - name: registration-dir
+              mountPath: /registration
+        - name: openebs-zfs-plugin
+          securityContext:
+            privileged: true
+            capabilities:
+              add: ["CAP_MKNOD", "CAP_SYS_ADMIN", "SYS_ADMIN"]
+            allowPrivilegeEscalation: true
+          image: quay.io/openebs/zfs-driver:ci
+          imagePullPolicy: IfNotPresent
+          args:
+            - "--nodeid=$(OPENEBS_NODE_ID)"
+            - "--endpoint=$(OPENEBS_CSI_ENDPOINT)"
+            - "--plugin=$(OPENEBS_NODE_DRIVER)"
+          env:
+            - name: OPENEBS_NODE_ID
+              valueFrom:
+                fieldRef:
+                  fieldPath: spec.nodeName
+            - name: OPENEBS_CSI_ENDPOINT
+              value: unix:///plugin/csi.sock
+            - name: OPENEBS_NODE_DRIVER
+              value: agent
+            - name: OPENEBS_NAMESPACE
+              value: openebs
+          volumeMounts:
+            - name: plugin-dir
+              mountPath: /plugin
+            - name: device-dir
+              mountPath: /dev
+            - name: encr-keys
+              mountPath: /home/keys
+            - name: zfs-bin
+              mountPath: /sbin/zfs
+            - name: libzpool
+              mountPath: /lib/libzpool.so.2
+            - name: libzfscore
+              mountPath: /lib/libzfs_core.so.1
+            - name: libzfs
+              mountPath: /lib/libzfs.so.2
+            - name: libuutil
+              mountPath: /lib/libuutil.so.1
+            - name: libnvpair
+              mountPath: /lib/libnvpair.so.1
+            - name: exportfs
+              mountPath: /usr/sbin/exportfs
+            - name: mount-nfs
+              mountPath: /usr/sbin/mount.nfs
+            - name: libtirp
+              mountPath: /usr/lib/x86_64-linux-gnu/libtirpc.so.3
+            - name: pods-mount-dir
+              mountPath: /var/lib/kubelet/
+              mountPropagation: "Bidirectional"
+              # needed so that any mounts setup inside this container are
+              # propagated back to the host machine.
+            - name: libgssapi
+              mountPath: /usr/lib/x86_64-linux-gnu/libgssapi_krb5.so.2
+            - name: libkrb
+              mountPath: /usr/lib/x86_64-linux-gnu/libkrb5.so.3
+            - name: libkrbcrypto
+              mountPath: /usr/lib/x86_64-linux-gnu/libk5crypto.so.3
+            - name: libkrbsupport
+              mountPath: /usr/lib/x86_64-linux-gnu/libkrb5support.so.0
+            - name: libkeyutils
+              mountPath: /usr/lib/x86_64-linux-gnu/libkeyutils.so.1
+            - name: exports
+              mountPath: /etc/exports
+            - name: nfs-state
+              mountPath: /var/lib/nfs
+      volumes:
+        - name: device-dir
+          hostPath:
+            path: /dev
+            type: Directory
+        - name: encr-keys
+          hostPath:
+            path: /home/keys
+            type: DirectoryOrCreate
+        - name: zfs-bin
+          hostPath:
+            path: /sbin/zfs
+            type: File
+        - name: libzpool
+          hostPath:
+            path: /lib/libzpool.so.2.0.0
+            type: File
+        - name: libzfscore
+          hostPath:
+            path: /lib/libzfs_core.so.1.0.0
+            type: File
+        - name: libzfs
+          hostPath:
+            path: /lib/libzfs.so.2.0.0
+            type: File
+        - name: libuutil
+          hostPath:
+            path: /lib/libuutil.so.1.0.1
+            type: File
+        - name: libnvpair
+          hostPath:
+            path: /lib/libnvpair.so.1.0.1
+            type: File
+        - name: exportfs
+          hostPath:
+            path: /usr/sbin/exportfs
+            type: File
+        - name: mount-nfs
+          hostPath:
+            path: /usr/sbin/mount.nfs
+            type: File
+        - name: libtirp
+          hostPath:
+            path: /usr/lib/x86_64-linux-gnu/libtirpc.so.3
+            type: File
+        - name: libgssapi
+          hostPath:
+            path: /usr/lib/x86_64-linux-gnu/libgssapi_krb5.so.2
+            type: File
+        - name: libkrb
+          hostPath:
+            path: /usr/lib/x86_64-linux-gnu/libkrb5.so.3
+            type: File
+        - name: libkrbcrypto
+          hostPath:
+            path: /usr/lib/x86_64-linux-gnu/libk5crypto.so.3
+            type: File
+        - name: libkrbsupport
+          hostPath:
+            path: /usr/lib/x86_64-linux-gnu/libkrb5support.so.0
+            type: File
+        - name: libkeyutils
+          hostPath:
+            path: /usr/lib/x86_64-linux-gnu/libkeyutils.so.1
+            type: File
+        - name: registration-dir
+          hostPath:
+            path: /var/lib/kubelet/plugins_registry/
+            type: DirectoryOrCreate
+        - name: plugin-dir
+          hostPath:
+            path: /var/lib/kubelet/plugins/zfs-localpv/
+            type: DirectoryOrCreate
+        - name: pods-mount-dir
+          hostPath:
+            path: /var/lib/kubelet/
+            type: Directory
+        - name: exports
+          hostPath:
+            path: /etc/exports
+            type: File
+        - name: nfs-state
+          hostPath:
+            path: /var/lib/nfs
+            type: Directory
--- a/pkg/zfs/mount.go
+++ b/pkg/zfs/mount.go
@@ -2,6 +2,8 @@ package zfs
 
 import (
 	"fmt"
+	"github.com/openebs/zfs-localpv/pkg/client/k8s/v1alpha1"
+	"net"
 	"os"
 	"os/exec"
 
@@ -69,13 +71,16 @@ func UmountVolume(vol *apis.ZFSVolume, targetPath string,
 		return err
 	}
 
-	if err = SetDatasetLegacyMount(vol); err != nil {
-		// ignoring the failure as the volume has already
-		// been umounted, now the new pod can mount it
-		klog.Warningf(
-			"zfs: failed to set legacy mountpoint: %s err: %v",
-			vol.Name, err,
-		)
+	// legacy mount does not work with the "organic" ZFS zfs->nfs logic
+	if vol.Spec.ShareNfs == "" || vol.Spec.ShareNfs == "off" {
+		if err = SetDatasetLegacyMount(vol); err != nil {
+			// ignoring the failure as the volume has already
+			// been umounted, now the new pod can mount it
+			klog.Warningf(
+				"zfs: failed to set legacy mountpoint: %s err: %v",
+				vol.Name, err,
+			)
+		}
 	}
 
 	if err := os.Remove(targetPath); err != nil {
@@ -136,7 +141,8 @@ func verifyMountRequest(vol *apis.ZFSVolume, mountpath string) error {
 	}
 
 	if len(vol.Spec.OwnerNodeID) > 0 &&
-		vol.Spec.OwnerNodeID != NodeID {
+		vol.Spec.OwnerNodeID != NodeID &&
+		vol.Spec.FsType != "nfs" {
 		return status.Error(codes.Internal, "verifyMount: volume is owned by different node")
 	}
 	if vol.Finalizers == nil {
@@ -150,8 +156,9 @@ func verifyMountRequest(vol *apis.ZFSVolume, mountpath string) error {
 		return status.Errorf(codes.Internal, "verifyMount: GetVolumePath failed %s", err.Error())
 	}
 
-	// if it is not a shared volume, then make sure it is not mounted to more than one path
-	if vol.Spec.Shared != "yes" {
+	klog.Warningf("vol.Spec is %v", vol.Spec)
+
+	if vol.Spec.FsType != FSTYPE_NFS && vol.Spec.Shared != "yes" {
 		/*
 		 * This check is the famous *Wall Of North*
 		 * It will not let the volume to be mounted
@@ -166,7 +173,7 @@ func verifyMountRequest(vol *apis.ZFSVolume, mountpath string) error {
 			return status.Errorf(codes.Internal, "verifyMount: Getmounts failed %s", err.Error())
 		} else if len(currentMounts) >= 1 {
 			klog.Errorf(
-				"can not mount, volume:%s already mounted dev %s mounts: %v",
+				"cannot mount, volume:%s already mounted dev %s mounts: %v",
 				vol.Name, devicePath, currentMounts,
 			)
 			return status.Errorf(codes.Internal, "verifyMount: device already mounted at %s", currentMounts)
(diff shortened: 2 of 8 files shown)


pawanpraka1:
@cruwe This PR is great!! Thanks for working on it. You might want to look at the creation of ZFSVolume CRDs part here :- https://github.com/openebs/zfs-localpv/blob/42ed7d85ee183249c5e6d2fedc482c988651f75d/pkg/driver/controller.go#L75

This will pick the paramaters from the storageclass and add them to the ZFSVolume CR.

Regarding the mounting, the Driver mounts the dataset in /var/lib/kubelet path which is the path where kubelet asks the driver to mount the volumes so that it is visible to the application 
https://github.com/openebs/zfs-localpv/blob/42ed7d85ee183249c5e6d2fedc482c988651f75d/deploy/yamls/zfs-driver.yaml#L841

pkg/apis/openebs.io/zfs/v1alpha1/zfsvolume.go
```diff
@@ -192,6 +192,11 @@ type VolumeInfo struct {
 	// FsType can not be modified once volume has been provisioned.
 	// Default Value: ext4.
 	FsType string `json:"fsType,omitempty"`
+
+	// ShareNfs specifies if, and when, how, to share the dataset via NFS.
+	// It depends on the Linux NFS Kernel Server to be running on the machine
+	// providing the volume and the volume to be mounted.
+	ShareNfs string `json:"sharenfs,omitempty"`
```
pawanpraka1, line 199: we have shifted to v1 apis, can remove this from here and make the changes in pkg/apis/openebs.io/zfs/v1 directory where v1 apis are there?

cruwe: Sure ... I will touch it again later this week. Currently I am reaching out to openzfs, because I do not understand how the `zfs share` code is supposed to interact with the `nfs-kernel-server`. 

Calling `execsnoop`, I see `zfs share` forking off to `exportfs`, but even with `hostIPC` and `hostPID`, I do not see the share being exported. Successful NFS export relies on calling `zfs share` from the host, which clearly is undesirable in our case.

You wouldn't happen to have an idea what might be happing there?

pawanpraka1: let me check that.

pawanpraka1:
@cruwe you need to do two things here 

1. mount exportfs binary from the host inside the ZFS-Driver, for reference you can see how we are mounting zfs binary from the host https://github.com/openebs/zfs-localpv/blob/f5ae3ff476e8823c3c35a19221ef3f8bd4bb5016/deploy/zfs-operator.yaml#L1276
2. Second thing is you also have to moount /var/lib/nfs directory inside the ZFS-Driver node daemonset. It is needed to do nfs stuff.

Once above two are done, you should not face any issue.

pawanpraka1:
@cruwe one more thing, the dataset mountpoint should also be mounted from the host. By default all the pod related mounts are happening in /var/lib/kubelet directory which driver mounts. So if you are mounting the dataset to the different path, then you have to mount it inside the daemonset. see here :  https://github.com/openebs/zfs-localpv/blob/f5ae3ff476e8823c3c35a19221ef3f8bd4bb5016/deploy/zfs-operator.yaml#L1263

cruwe:
@pawanpraka1: thanks for the hint with the `/var/lib/nfs`. Tried writing into `/etc/exportfs` directly first and got that as an error-msg. Tried it with the node-ds until that worked and then did it with `zfs share`, which suceeded.

I implemented a dirty hack to temporarily disable your legacy-mountpoint - zfs dataset, sharenfs and legacy do not play together. That we need to reconcile:

Assuming a sharenfs dataset is only exported and not used by any local pods, it is possible to legacy mount with sharenfs off or unset. If it is desired to mount locally _and_ sharenfs, we need to work with `/etc/exports` directly. I'd rather avoid that it that option is still open, because I ran into lots of issues when trying to tear that down before deletions.

In any case, thanks!

cruwe:
The last commit lead to a working deployment, mounted and exported datasets, which may be mounted and written to. It's still dirty (PoC-style), needs lots of love clearing up and validating, and could use two or three lines of documentation. Let's talk about it later this week. Looking forward to hearing (reading) from you, cheers!

pkg/zfs/mount.go
```diff
@@ -69,13 +76,16 @@ func UmountVolume(vol *apis.ZFSVolume, targetPath string,
...
 	}
 
-	if err = SetDatasetLegacyMount(vol); err != nil {
-		// ignoring the failure as the volume has already
-		// been umounted, now the new pod can mount it
-		logrus.Warningf(
-			"zfs: failed to set legacy mountpoint: %s err: %v",
-			vol.Name, err,
-		)
+	// TODO: ignore the legacy mount feature for a moment until correctly sharing
```
cruwe, line 79: This is the part I temporarily disabled. Sorry about that, just wanted the sharing to work. Let's talk about options later on.

pkg/zfs/zfs_util.go
```diff
@@ -141,7 +155,12 @@ func buildCloneCreateArgs(vol *apis.ZFSVolume) []string {
 			reservationProperty := "reservation=" + vol.Spec.Capacity
 			ZFSVolArg = append(ZFSVolArg, "-o", reservationProperty)
 		}
-		ZFSVolArg = append(ZFSVolArg, "-o", "mountpoint=legacy")
+		if vol.Spec.ShareNfs == "" || vol.Spec.ShareNfs == "off" {
+
```
cruwe, line 159: same here

pkg/zfs/zfs_util.go
```diff
@@ -234,10 +253,18 @@ func buildDatasetCreateArgs(vol *apis.ZFSVolume) []string {
...
-	ZFSVolArg = append(ZFSVolArg, "-o", "mountpoint=legacy", volume)
-
+	if len(vol.Spec.ShareNfs) != 0 {
+		// TODO consider validation of at least important opts of exports(5)
+		shareNfsProperty := "sharenfs=" + vol.Spec.ShareNfs
+		ZFSVolArg = append(ZFSVolArg, "-o", shareNfsProperty)
+		// TODO: vol needs to be mounted to be shared via NFS
+	}
+	if vol.Spec.ShareNfs == "" || vol.Spec.ShareNfs == "off" {
+		// set the mount path to none, by default zfs mounts it to the default dataset path
```
cruwe, line 263: and here

pkg/zfs/zfs_util.go
```diff
@@ -390,8 +424,15 @@ func SetDatasetMountProp(volume string, mountpath string) error {
 func MountZFSDataset(vol *apis.ZFSVolume, mountpath string) error {
 	volume := vol.Spec.PoolName + "/" + vol.Name
 
-	// set the mountpoint to the path where this volume should be mounted
-	err := SetDatasetMountProp(volume, mountpath)
+	var err error
+	// WARNING: this is a dirty hack just for PoC
+	if vol.Spec.ShareNfs == "" || vol.Spec.ShareNfs == "off" {
```
cruwe, line 429: and here ... sorry

cruwe:
Sorry for my radio-silence.

@pawanpraka1:

I'd suggest treating the concept of mountung a pvc as a regular zfs dataset for some pod _and_ exporting it at the same time as a special case, which may be implemented later, but is not necessary for completion of the task at this point.

Then, an NFS-export may be declared by either setting the `sharenfs` property or declaring a special `fsType: nfs`. (I'd favour the latter, because the former makes an implementation of the case 'mounted at pod _and_ exported' harder.)

I'd then proceed introducing a disambiguation as in 'when nfs share, _do_ mount with legacy != none irrespective of pod consumption, when ! nfs, proceed as normally'.

Having that, the pvc may be consumed by as an nfs pvc (less elegant, but quick).

Eventually, the mount-logic should be adapted replacing the `mount -t zfs` part with `-t nfs` when `fsType: nfs` or equivalent. My understanding of the mount logic is still sparse, but I believe then, a pod with corresponding pvc should be created on the same node which provides the volume.

As a last step, this host affinity (which I do not know much about) needs to be resolved conditionally, _i.e._, affinity on when zfs, off when nfs. I have no idea how to do that, but I figure I'd find out soon enough - might require some help, though.

How does this sound to you?

pawanpraka1:
@cruwe Sorry for the delay, Was busy with the release activity and some bugs in k8s.

We need to do few things here 

1. Don't set the affinity on the PV. We should avoid this step while creating the volume. I want nfs volume can be mounted by the pod running on different, so for nfs volumes, we should not set the affinity on the PV. Please note that we still have to run scheduler to find out where I have to create the volume but we don't need to set the affinity for the PV.

https://github.com/openebs/zfs-localpv/blob/master/pkg/driver/controller.go#L237

2. In a lot of situations, you would not be happy to give every host on your network access to your NFS. You probably only want to give one specific host, or a specific group of hosts, access. So we still need to do something like sharenfs="rw=@192.168.11.0/24".

3. We need to dedicate one directory /var/tmp/nfs on the host machine, where we can setup the volume mounts. the volumes will be mounted at /var/tmp/nfs/<poolname>/<volume> path. And the pods using this volume will use this path while mounting it locally. this will be part of volume creation. While creating the volume, if sharenfs is set, we will switch the mountpoint to /var/tmp/nfs/<poolname>/<volume> from legacy 

https://github.com/openebs/zfs-localpv/blob/ab32a5a4267b137896a247bf648f0f77fc3e7c62/pkg/zfs/zfs_util.go#L239

4. For nfs volumes, we have to by-pass the "Wall of north"  which does not allow the mounting of the  device to more than one path. It is there to prevent the volumes getting mounted by more than one pod.

https://github.com/openebs/zfs-localpv/blob/ab32a5a4267b137896a247bf648f0f77fc3e7c62/pkg/zfs/mount.go#L151

cruwe:
Hi ... don't apologize for any delays. As you might notice, I am sluggish atm myself ...

Regarding (1), I think I successfully unset the node topology, although my code is dirty at best. I'll look into that to clean that up.

The sharenfs restriction (2) to a network is already implemented, although optional from the beginning. Would you make that mandatory?

Regarding (3), I do not know if I read you correctly. `mountpoint=legacy` will not work with the zfsutils sharenfs magic. Then, it would be necessary to manage (and validate) the exportfs logic oneself, which is a daunting task.

Wall of North (4)? Did I miss something important in popular culture?  In any case, the branch I just pushed contains logic to provide a ZFS share via NFS and allow a pod on a different node to mount that. I am still investigation an issue when ZFS datasets are not unshared and not destroyed. This issue prevents using further NFS mounts until cleaning up manually.

Furthermore, what I did to speed up the testing was to introduce a zfs-driver on nodes without zpools (dividing btw worker and storage nodes), which kind of works, but feels very ugly to me. I don't have a better idea, though.

In any case, thanks for helping me along. Looking forward to your comments, cheers!

pawanpraka1:
> Regarding (1), I think I successfully unset the node topology, although my code is dirty at best. I'll look into that to clean that up.

Great!!

> The sharenfs restriction (2) to a network is already implemented, although optional from the beginning. Would you make that mandatory?

It is not mandatory as sharenfs=on is still a valid configuration and if someone wants to put restriction on it, he can use sharenfs="rw=@10.43.11.0/24".

> Regarding (3), I do not know if I read you correctly. mountpoint=legacy will not work with the zfsutils sharenfs magic. Then, it would be necessary to manage (and validate) the exportfs logic oneself, which is a daunting task.

So in (3), the real question is how do the applications mount the nfs volumes? we need to do "mount -t nfs zfs.host.com:/pool-name/dataset-name /path/to/local/mount", where /pool-name/dataset-name is the default mount point. For this to work, the driver has to mount the /pool-name directory, which is not static as pool-name can change. So have one directory /var/tmp/nfs where we will mount the datasets. Each volume will have mountpoint set to  /var/tmp/nfs/pool-name/volname (since we can have more than one pools on a node). Let me know if you need more clarification on it.


> Wall of North (4)? Did I miss something important in popular culture?
Game of Thrones :)

> In any case, the branch I just pushed contains logic to provide a ZFS share via NFS and allow a pod on a different node to mount that. I am still investigation an issue when ZFS datasets are not unshared and not destroyed. This issue prevents using further NFS mounts until cleaning up manually.

Let me know if you need help debugging this.

> Furthermore, what I did to speed up the testing was to introduce a zfs-driver on nodes without zpools (dividing btw worker and storage nodes), which kind of works, but feels very ugly to me. I don't have a better idea, though.

Another way to test is have two or more application pods, which should be using the same pvc.
 Make sure the pods are running on different nodes. And write the data from one pod and verify it from the other pod.

cruwe:
Hi,

I got all functional aspects woring. Datasets get created and nfs-mounted. The data is shared btw multiple pods on multiple hosts and the storage host itself and datasets get "unshared" correctly before `zfs destroy` without manual intervention. 

Regarding (3), I do not understand why a possible change of the pool's name would affect the mountpoint. As far as I understand, it is set on the host and the storageclass must (musn't it?) have the `poolname` propery set correspondingly? Then, wouldn't the poolname be known before volume provisioning and may be read from `vol.Spec.Poolname`? This is what I did with the `mount -t nfs` logic and it seemed to work.

In any case, I'm equally fine with setting it manually, I respectfully ask, however, the mount-root not to be `/var/tmp`, which I think should be reserved for temporary files. Could I somehow convince you to make it sth like `/exports/openebs/<poolname>/<volname>` per default and perhaps configurable from the parameters? That would allow those with dedicated storage only partially dedicated to k8s to differentiate and those with multiple zpools to separate concerns.

Regarding the testing, I tested with two pods on two nodes, one of them the storage provider and one of them an nfs-only consumer. I set `watch find` on the paths and `touch`ed files manually on both pods. Sharing worked fine between both nodes and the host.

I believe I saw a situation where file-propagation seemed to be stuck for some seconds. This is, I believe, impossible to rule out completely, especially when testing with two VirtualBoxen and flannel.

I am trying to debug a curios situation, where the lock-manager is non-functional yielding
````
time="2020-06-23T19:50:50Z" level=error msg="mount: could not mount the dataset 192.168.100.235:/tank/pvc-2bb3a01a-366e-4c4a-b766-4b635459dc15 cmd [-t nfs 192.168.100.235:/tank/pvc-2bb3a01a-366e-4c4a-b766-4b635459dc15 /var/lib/kubelet/pods/82664564-5fc8-473a-95e0-c2647c86fb79/volumes/kubernetes.io~csi/pvc-2bb3a01a-366e-4c4a-b766-4b635459dc15/mount] error: mount.nfs: rpc.statd is not running but is required for remote locking.\nmount.nfs: Either use '-o nolock' to keep locks local, or start statd.\n"
time="2020-06-23T19:50:50Z" level=error msg="GRPC error: rpc error: code = Internal desc = rpc error: code = Internal desc = dataset: mount failed err : mount.nfs: rpc.statd is not running but is required for remote locking.\nmount.nfs: Either use '-o nolock' to keep locks local, or start statd.\n
```
I do not know if that is caused by my flaky testing setup or if that is a "real" issue.

Lastly, the destroy volume logic leaves directories formerly used as mountpoints lying around. Formerly, this was a non-problem because the were set under the kubelet mount. Now, they either are at the zpool root-dataset's root or under whatever is chosen to be the export-root. What's your take on this? I'd suggest looking at the volumes reclaim-policy (btw, where do I get that if it is unset at the parameters hash?) and then doing a conditional `rmdir`.

Anyhow, I am reasonably happy and if you do not completely disagree, I'd start commenting, documenting and in general straitening and prettying that thing up.

Thank's for your help and your patience, cheers!

pawanpraka1:
> I got all functional aspects woring. Datasets get created and nfs-mounted. The data is shared btw multiple pods on multiple hosts and the storage host itself and datasets get "unshared" correctly before zfs destroy without manual intervention

question: Are you unsharing it explicitely before destroy? Isn't zfs destroy unshares (and umount also) before deleting the datatset?

> Regarding (3), I do not understand why a possible change of the pool's name would affect the mountpoint. As far as I understand, it is set on the host and the storageclass must (musn't it?) have the poolname propery set correspondingly? Then, wouldn't the poolname be known before volume provisioning and may be read from vol.Spec.Poolname? This is what I did with the mount -t nfs logic and it seemed to work.

hmmm, what if I have two storageclasses, one for pool1 and another for pool2. how do we handle that, sorry for going back and forth on this.

> I am trying to debug a curios situation, where the lock-manager is non-functional yielding

check if nfs utils are installed whre pods are running. Will check the code.

> Lastly, the destroy volume logic leaves directories formerly used as mountpoints lying around. Formerly, this was a non-problem because the were set under the kubelet mount. Now, they either are at the zpool root-dataset's root or under whatever is chosen to be the export-root. What's your take on this? I'd suggest looking at the volumes reclaim-policy (btw, where do I get that if it is unset at the parameters hash?) and then doing a conditional rmdir.

we can cleanup that at the destryo time : https://github.com/openebs/zfs-localpv/blob/2b13a04db427c752648486858ca09daba4b79271/pkg/zfs/zfs_util.go#L505

cruwe:
> question: Are you unsharing it explicitely before destroy? Doesn't zfs destroy unshare (and umount also) before deleting the datatset?

Yes and yes. I unshare explicitly which usually would not be necessary, because `zfs destroy` is supposed to and usually does unshare.

When called from inside the container. it doesn't, though, or perhaps it tries and doesn't succeed. What happens is that the dataset is allegedly busy, so neither unmounting nor destroying is possible.

I tried numerous attempts, found out that restarting the `nfs-kernel-server` on the host will free the dataset, did an `execsnoop` from the `bpfcc-tools` suite to find out which processes are forked when restarting the `nfs-kernel-server` and boiled that set of five or six down to the minimal number of exec-calls from the inside which free the dataset so that it may be destroyed. That is what I then replicated in the `UnshareZFSDataset` func.

I still could not drill down to the real cause and all my acquaintances knowledgeable about ZFS are Solaris guys and a mail to `zfsdiscuss` did not bear any fruit so far.

> hmmm, what if I have two storageclasses, one for pool1 and another for pool2. how do we handle that, sorry for going back and forth on this.

Without `legacy` mount, the mount-points then would be just the standard ZFS `/pool1/pcv-...` and `/pool2/pvc-...` which is `<vol.Spec.Poolname>/<vol.Name>`.  With `legacy`, one could do `/exports/openebs/<vol.Spec.Poolname>/<vol.Name>` for better structure on the host or, as I believe the pvc-names to be unique, scrap the `<vol.Spec.Poolname>`. Unless I am missing something. Do I?

> check if nfs utils are installed whre pods are running. Will check the code.

I believe they are, but I will double-check and perhaps trace when and preferably why the lock-manager stops working.

> we can cleanup that at the destroy time : 

I'll implement that, but possibly not before Friday.

Cheers!

cruwe:
@pawanpraka1: I very much fear I got something completely wrong here:

>  Don't set the affinity on the PV. We should avoid this step while creating the volume. I want nfs volume to be mounted by the pod running on different nodes, so for nfs volumes, we should not set the affinity on the PV. Please note that we still have to run scheduler to find out where I have to create the volume but we don't need to set the affinity for the PV.
> https://github.com/openebs/zfs-localpv/blob/master/pkg/driver/controller.go#L237

Initially, I thought topology to govern what constraints a node must satisfy to mount a volume, wich is sort of correct, I beleive. What I did not understand is, that it also governs what constraints a node must satisfy _to create_ the zvol. I started buld-testing today and then, I learned that the system tried to create the zvols on nodes without any zpools - I divide btw storage- and worker-nodes. Of course, this fails.

I monkey-patched the scheduler to stop scheduling and accept the decision taken when ownerNodeId is set at the storageclass (labels probably would have been better, but it is meant as a stop-gap measure). Then, of course, only pods on that storage nodes may be scheduled. which thwarts the idea.

How would you solve the problem that a volume may be _created_ on specific nodes only, but _mounted_ from everywhere / inside certain network segments / etc?

Thanks and cheers!

pawanpraka1:
> How would you solve the problem that a volume may be created on specific nodes only, but mounted from everywhere / inside certain network segments / etc?

@cruwe You can label the storage nodes something like : openebs.io/node=storage. And then use this label in zfs-operator.yaml to deploy the daemonset. This way driver will run on storage nodes only and will create the volumes there only. Now you don't need to put any topology in the storageclass as kubernetes scheduler will schedulre the nodes there only, so avoid doing this. Let me know if it make sense.

cruwe:
@pawanpraka1 : I feel we might be talking about different aims here.

My idea was to have nodes in a cluster which provide persistent storage and different nodes, which don't, but only consume that storage. What I am trying to achieve is to have the scheduler put the zfs dataset one one node (labelled as you suggest, for instance)  and have a pod of `openebs-zfs-node` mount the volume via nfs on a different node, which does'nt have zpools.

When the `openebs-zfs-node` may be scheduled only on nodes with zpools present, it wouldn't be possible to use the driver to schedule mounts for nfs consumers on zpool-less nodes. This is why I looked for ways for the scheduler to respect the affinity constraint for the creation of a ZFS dataset and to disregard it when it is of `fstype=nfs` and mounting it.

I hope what I try to achive does not run completely against your ideas of the architecture of your software.

pawanpraka1:
I looked at the scheduler code (https://github.com/openebs/zfs-localpv/pull/139/files#diff-61566437d9b17d18936960aa0bce1ad4R245) and it seems like we are setting the topology in the createvolume response. That will make all the application pod to be scheduled on the same node only which is what we don't want. We should avoid setting the topology for the volume. See this comment :-

> Don't set the affinity on the PV. We should avoid this step while creating the volume. I want nfs volume to be mounted by the pod running on different nodes, so for nfs volumes, we should not set the affinity on the PV. Please note that we still have to run scheduler to find out where I have to create the volume but we don't need to set the affinity for the PV.
https://github.com/openebs/zfs-localpv/blob/master/pkg/driver/controller.go#L237


Coming to the testing part, we can let the scheduler to schedule the volume on any node(but don't set the affinity), then we can use the pod affinity in application deployment to deploy the application to node where volume is not present. here, note that the application pod can be scheduled to the node where node daemonset are running as we need it to mount the volume.

cruwe:
Hi,

thanks again for your help. I am sorry if I am "difficult", I do not mean to and I fear I have been a bit dense understanding your intention regarding topology and placement.

Do I read you that the `NewCreateVolumeResponseBuilder()` should drop the topology-constraint when the volume is of nfs-type? When it is of `type=zfs`, I believe it cannot, because mounting then requires same-host semantics. 

To have ZFS datasets provisioned only where `zpool`s exist, you suggest to use `allowedTopologies` and labels, correct? That part took me very long to understand, sorry to have demanded so much of your patience here.

In any case, if I hope to have that correct now so that I can start debugging the `rpc-statd` I observed in earnest. 

Thanks and cheers!

pawanpraka1:
@cruwe the issue with setting topology(with will set nodeSelectorTerms in PV) in volume is that all the pods using that volume will be scheduled to the same node. You can check `kubectl get pv -oyaml` it will look something like this :-

```yaml
apiVersion: v1
kind: PersistentVolume
metadata:
  name: fio-vol-pv # some unique name
spec:
  accessModes:
  - ReadWriteOnce
  capacity:
    storage: 4Gi
  claimRef:
    apiVersion: v1
    kind: PersistentVolumeClaim
    name: fio-vol-pvc
    namespace: default # namespace for the pvc 
  csi:
    driver: zfs.csi.openebs.io
    fsType: zfs 
    volumeAttributes:
      openebs.io/poolname: zfspv-pool # change the pool name accordingly
    volumeHandle: fio-vol # This should be same as the zfs volume name
  nodeAffinity:
    required:
      nodeSelectorTerms:
      - matchExpressions:
        - key: openebs.io/nodename
          operator: In
          values:
          - pawan-3
  persistentVolumeReclaimPolicy: Delete
  storageClassName: openebs-zfspv
  volumeMode: Filesystem
```

Now when POD uses this volume, the kubernetes scheduler will check if there is any nodeSelectorTerms in the PV, and try to schedule the pod matching with that term.

The above is good for Local PV solution but with NFS as it is over the Network, it is not need to the pod to be scheduled on the same node. The volume can be created on the one node and the pod can use this volume from different node via mount -t nfs volumenode:/pool-name/dataset-name /path/to/local/mount.

For example lets say we have two nodes ; node1 and node2. i have created a storageclass and pvc like below 
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: openebs-zfspv
allowVolumeExpansion: true
parameters:
  poolname: "zfspv-pool"
  fstype: "zfs"
```

```yaml
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: csi-zfspv
spec:
  storageClassName: openebs-zfspv
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 4Gi 
```

Now after applying the aboe yaml, the scheduler decided to put the volume on node1. The volume will be created on the pool zfspv-pool that is there on node1. Now if I am deploying the application using above pvc, and kubernets scheduler decides to put the pod on node2. The daemonset agent running on node2 just have to check if it is a nfs volume, then mount the volume using
`mount -t nfs volumenode:/pool-name/dataset-name /target/path`. here note the since the volume is not present on node2, we can not fire any zfs set/get command which we are doing in case local volume, where pod will always come to the same node. In NFS since it is not necessary that pod will come to the same node where volume is present and pod running anywhere can moun the volume. 

I am so sorry for not describing it in the details. Let me know if it makes sense to you. Also, if you want, you can join openebs-dev and openebs channel to discuss anything
1. https://kubernetes.slack.com/messages/openebs/
2. https://kubernetes.slack.com/messages/openebs-dev/

cruwe:
Hi,

thank you for your detailed explanation. Please do not be sorry, I believe I learned more this way.

I believe I understood what you are telling me and implemented that _not setting_ the `WithTopology(topology)` when creating zfs volumes _and_ configuring the storageclass to only allow topologies with matching label. When I deploy, it now works as I expect. Perhaps you could have a brief look and I start diving into writing tests next week.

I just joined the slack channels and will commit a (possible dynamic) slice of an eye, Thanks very much for your kind help and your patience. Cheers!

pawanpraka1:
@cruwe I will take a look at the PR. Can you resolve the conflicts and update the it?

cruwe:
Re-based on master and squashed to improve legibility. Thank your for your patience!

pkg/zfs/zfs_util.go
```diff
@@ -446,6 +485,37 @@ func SetDatasetLegacyMount(vol *apis.ZFSVolume) error {
...
 
+func UnshareZFSDataset(vol *apis.ZFSVolume) error {
+	// ZFS datasets cannot be destroyed without being unshared first, as without
+	// the rpcbind will still hold handles on the ds, resulting in a is busy err
+	// on destroy, unmount etc
+	volume := vol.Spec.PoolName + "/" + vol.Name
+	mntPnt := "/" + volume
+
+	var UnshareVolArg []string
+	UnshareVolArg = append(UnshareVolArg, "-u", "192.168.0.0/16:"+mntPnt)
```
pawanpraka1, line 496: harcoded ip?

pkg/zfs/zfs_util.go
```diff
@@ -236,13 +249,25 @@ func buildDatasetCreateArgs(vol *apis.ZFSVolume) []string {
...
 	}
 
-	// set the mount path to none, by default zfs mounts it to the default dataset path
-	ZFSVolArg = append(ZFSVolArg, "-o", "mountpoint=legacy", volume)
-
+	// NOTICE: with mountpoint legacy, the zfs logic which automagically exports
+	// NOTICE: on sharenfs will not work and exporting is left to the operator
+	if vol.Spec.FsType == FSTYPE_NFS {
+		//mountPoint := NFS_ROOT + "/" + vol.Name
+		mountPoint := "/" + vol.Spec.PoolName + "/" + vol.Name
```
pawanpraka1, line 256: this mountpoint will be set inside the container. This mountpoint will not be propageted to the host. see the 3rd point here : https://github.com/openebs/zfs-localpv/pull/139#issuecomment-645250165. We need to dedicate one directory from the host and mount that in biderectional mode so that mount within the container can propagate to the host. The mount path can be anything "/var/lib/zfspv/" or anything.

deploy/yamls/ubuntu/zfs-driver.yaml
```diff
@@ -841,4 +933,225 @@ spec:
...
+          hostPath:
+            path: /usr/sbin/mount.nfs
+            type: File
+        - name: libtirp
+          hostPath:
+            path: /usr/lib/x86_64-linux-gnu/libtirpc.so.3
+            type: File
+        - name: libgssapi
+          hostPath:
+            path: /usr/lib/x86_64-linux-gnu/libgssapi_krb5.so.2
```
pawanpraka1, line 1120: question: which os you are using and which zfs version? I have seen this library is needed for centos

deploy/yamls/ubuntu/zfs-driver.yaml
```diff
@@ -791,11 +808,42 @@ spec:
...
+              mountPath: /usr/lib/x86_64-linux-gnu/libkrb5support.so.0
+            - name: libkeyutils
+              mountPath: /usr/lib/x86_64-linux-gnu/libkeyutils.so.1
+            - name: libwrap
+              mountPath: /usr/lib/x86_64-linux-gnu/libwrap.so.0
+            - name: exports
+              mountPath: /etc/exports
+            - name: nfs-state
+              mountPath: /var/lib/nfs
+            - name: tank
```
pawanpraka1, line 842: what is this directory?

deploy/yamls/ubuntu/zfs-driver.yaml
```diff
@@ -829,6 +877,50 @@ spec:
...
+          hostPath:
+            path: /usr/sbin/exportfs
+            type: File
+        - name: mount-nfs
+          hostPath:
+            path: /usr/sbin/mount.nfs
+            type: File
+        - name: rpc-nfsd
+          hostPath:
+            path: /usr/sbin/rpc.nfsd
```
pawanpraka1, line 890: why we need this?

pkg/zfs/zfs_util.go
cruwe: good that you spotted that ... "quick fix" and then forgotten. I'll fix that

pkg/zfs/zfs_util.go
cruwe: Not sure if I read you right: I used the more or less "canonical" ZFS dataset root so that I do not need to care about machines with multiple zpools. Constructing the bidirectional mount correctly then needs to be done when deploying the driver.

You would like a "fixed" path better so that there is no disambiguation regardless whether one or more zpools are present on the machine? Like `"/var/lib/zfspv+  "/" + vol.Spec.PoolName + "/" + vol.Name`?

deploy/yamls/ubuntu/zfs-driver.yaml
cruwe: I am on Ubuntu Focal using zfs 0.8.3-1ubuntu12.1.

deploy/yamls/ubuntu/zfs-driver.yaml
cruwe: That is the "canonical" dataset root for any datasets under the zpool named tank

deploy/yamls/ubuntu/zfs-driver.yaml
cruwe: I do not find out the "exact" path how sharing and unsharing datasets needs to be done to free the kernel locks after a dataset has been shared.

I did the whole bunch for loss of better knowledge of zfs share and nfs-kernel interaction using trial and error.

I left that alone when I got the unsharing and destroying to work. I am unsure if it is really needed. Considering that I still observe that it sometimes takes up to five minutes to get the share mounted on my VirtualBoxen, I'd like to leave the task "derive the minimal set of necessary binaries"  as one of the last steps.  Do you know more about the zfs and in kernel NFS interaction?

pkg/zfs/zfs_util.go
pawanpraka1: The thing here is the operator will be deployed once and after that user is free to setup the ZFS POOL he wants. So he can setup the  multiple ZPOOL. Like on a node he might have added two kinds of disk - ssd nad nvme. And he might have created two ZFS Pools - zpool-nvme, zpool-ssd. He he will have two storageclasses :-
```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: openebs-zfspv
allowVolumeExpansion: true
parameters:
  poolname: "zpool-ssd"
``` 

```
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: openebs-zfspv
allowVolumeExpansion: true
parameters:
  poolname: "zpool-nvme"
``` 

he can use the storageclasses for his applications. For example the application which need higher IOPS, he will use nvme storageclass and for others applications he cal use ssd storageclass.

How this use case will work here?

pawanpraka1:
@cruwe let me know if you are blocked on this? I see the PR is still in draft stat.

cruwe:
Hi Pawan ... I am so sorry, I am loaded in work at the moment and will be very sluggish to reply for at least the next two weeks. I'll try to respond properly later the week, but I cannot promise anything without blatantly lying. Please excuse. Cheers!

sinhaashish:
closing this PR as it has been stale for long . @cruwe  please raise another PR for this change

PR #144, opened 2020-06-01
fix(provisioning): make volumes with fsType=zfs datasets of type filesystem

### LINKED ISSUE #143: CreateZFSVolume does not respect fsType=zfs -> dataset
With storageclass
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
  name: zfs-dataset
allowVolumeExpansion: true
parameters:
  fsType: "zfs"
```
the created datasets end up with type volume.

With fsType set, I'd like to have a dataset.

The parameter `req.GetParameters()["fstype"]` in `func CreateZFSVolume(req *csi.CreateVolumeRequest) (string, error)` is inconsistent with the CRD. Either it is fstype in the CRD or it is `req.GetParameters()["fsType"]`.

cruwe:

fixes #143 -> inconsistent naming of fsType with fallback to ext4 and all datasets created as volumes

**What this PR does?**:
Reformulating the getter allows volumes to be created with zfs-type, which will result in a "filesystem" dataset.

I do not think so.

Without this fix, a dataset as described in #143 is created as a volume. With the fix, it is created as a dataset of type filesystem. Conersly, with type xfs, it is still created as a dataset of type volume. I imagine my testing could have been more thorough, this is what I managed to do to with the time at hand.

**Checklist:**
- [x] Fixes #143 
- [ ] Commit has unit tests

### DIFF
--- a/changelogs/unreleased/144-cruwe
+++ b/changelogs/unreleased/144-cruwe
@@ -1,0 +1,1 @@
+Fixes an issue where volumes meant to be filesystem datasets got created as zvols and generally makes storageclass parameter spelling insensitive to case
--- a/pkg/common/helpers/helpers.go
+++ b/pkg/common/helpers/helpers.go
@@ -1,0 +1,43 @@
+/*
+Copyright 2020 The OpenEBS Authors
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package helpers
+
+import (
+	"strings"
+)
+
+// Coerce the map's keys to lower case, which only works when unicode char is in
+// ASCII subset. May overwrite key-value pairs on different permutations of key
+// case as in Key and key. DON'T force values to the lower case unconditionally,
+// because values for keys such as mountpoint or keylocation are case-sensitive.
+// Note that although keys such as 'comPREssion' are accepted and processed,
+// even if they are technically invalid, updates to rectify such typing will be
+// prohibited as a forbidden update.
+func GetCaseInsensitiveMap(dict *map[string]string) map[string]string {
+	insensitiveDict := map[string]string{}
+
+	for k, v := range *dict {
+		insensitiveDict[strings.ToLower(k)] = v
+	}
+	return insensitiveDict
+}
+
+// special case of GetCaseInsensitiveMap looking up one key-value pair only
+func GetInsensitiveParameter(dict *map[string]string, key string) string {
+	insensitiveDict := GetCaseInsensitiveMap(dict)
+	return insensitiveDict[strings.ToLower(key)]
+}
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -27,6 +27,7 @@ import (
 	"github.com/openebs/zfs-localpv/pkg/builder/snapbuilder"
 	"github.com/openebs/zfs-localpv/pkg/builder/volbuilder"
 	errors "github.com/openebs/zfs-localpv/pkg/common/errors"
+	"github.com/openebs/zfs-localpv/pkg/common/helpers"
 	csipayload "github.com/openebs/zfs-localpv/pkg/response"
 	analytics "github.com/openebs/zfs-localpv/pkg/usage"
 	zfs "github.com/openebs/zfs-localpv/pkg/zfs"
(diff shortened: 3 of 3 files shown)


pawanpraka1:
@cruwe thanks for this PR, but this is a breaking change and if customers are already using fstype in the storageclass. After this change, that will stop working and it will create ZVOL instead of dataset.

The right way to handle this is make fstype it case insensitive. The way kubernetes does this.

cruwe:
Ok, I understand. So that I read correctly, you propose to make the controller accept both, `fstype` and `fsType`.?

The change will, I fear, still be breaking, because customers who thought they'd get a filesystem-type actually got a zvol with ext4, I believe. I'd imagine the breakage to be smaller, but it is a change of behaviour. How would you deal witht that situation?

In any case, thank you very much for your input, I'll mark that WIP for the time being.

pawanpraka1:
> Ok, I understand. So that I read correctly, you propose to make the controller accept both, fstype and fsType?

> The change will, I fear, still be breaking, because customers who thought they'd get a filesystem-type dataset actually got a zvol with ext4, I believe.

I am suggesting something like this : fetch the fstype from req.GetParameters()["fstype"] if not availabel then from req.GetParameters()["fsType"] or req.GetParameters()["FStype"] or req.GetParameters()["FSTYPE"]. So it will not break as old storageclass will have fstype. The intention is to make it case insensitive and user can use anything.

But I would not worry about making this in-sync with ZFSVolume CRD as it is there just for the display purpose. It serves no purpose other than that. 

I would want the controller to understand all possible fstype string. But this is fine since as of now it only supports fstype string.

pkg/driver/controller.go
```diff
@@ -27,6 +27,7 @@ import (
 	"github.com/openebs/zfs-localpv/pkg/builder/snapbuilder"
 	"github.com/openebs/zfs-localpv/pkg/builder/volbuilder"
 	errors "github.com/openebs/zfs-localpv/pkg/common/errors"
+	"github.com/openebs/zfs-localpv/pkg/common/helpers"
```
cruwe, line 30: I am really unsure where to place helper functions. What is your opinion and/or advise in such matters?

cruwe:
I implemented looking up what is meant to be the fstype parameter (semantically) based on your suggestion of precedence. 

I have an idea on how to solve `GetInsensitiveToCase` (should I rename it yo `GetInsensitivelyToCase`?), I do not know, however, how to deal with UTF-8 when the keys are not in the ASCII-equivalent subset. I'd rather leave that part out. How do you feel about that?

pkg/driver/controller.go
pawanpraka1: this place is fine.

pawanpraka1:
Nice!! I like the idea of GetInsensitiveToCase. Can we rename it to GetInsensitiveParameters. Let us use this to convert all the parameter into a map of lower case keys and then use lower case keys for the lookup?

cruwe:
Sure, changed the name. You're right, it increases readability. At worst, it might bring a smile to the inner child of some telenovela watcher, who might imagine a passionate argument btw someone and a parameter who is "just sooo insensitive". As long as we do not have a `func GetInsensitiveCharacter()`. As we see again, there are only two hard things in computer science: Naming and cache invalidation.

Ok, joking aside, when we want do be insensitive (no pun intended, seriously!) with all parameters, I'd suggest to create a map with case-insensitive keys and do lookups on this map to avoid getting & copying the original map the umpteenth time. I have not yet tried that out and I still need to look at the tests to see if is necessary to adapt some code there.

pawanpraka1:
@cruwe sounds good.

cruwe:
In this case, I will run a set of superficial tests this evening and upon successful completion, remove draft status and WIP.

cruwe:
Hi, I've tested with some common and some perhaps not so common spelling variations, for instance, 'comPREssion', 'Dedup', 'POOLNAME' or even 'fStYpE'. 

PVCs derived from this storageclass are created correctly. It is not possible to rectify this sort of creative spelling afterwards due to the storageclass parameters being immutable. That cannot be helped, I fear.

I added some additional comments to explain the reasoning to the casual reader and then squash-rebased to reduce history clutter.

Thank you very much your comments and generally, helping with developing the code. I'll now remove the WIP and draft status. Looking forward to hearing from you with the sharenfs part. Cheers!

pkg/common/helpers/helpers.go
```diff
@@ -0,0 +1,43 @@
+/*
+Copyright 2019 The OpenEBS Authors
```
kmova, line 2: n_: 2020 :-)

cruwe: oops ...

pawanpraka1:
@cruwe can you add a changelog also https://github.com/openebs/zfs-localpv/blob/master/CONTRIBUTING.md#adding-a-changelog.  See the example here :- https://github.com/openebs/zfs-localpv/blob/master/changelogs/unreleased/121-pawanpraka1

cruwe:
@kmova: Done. Thanks for spotting that.
@pawanpraka1: Done. Thanks for pointing out how to do it.

changelogs/unreleased/144-cruwe
```diff
@@ -0,0 +1 @@
+Fixes an issue where volumes meant to be filesystem datasets got created as zvols
```
pawanpraka1, line 1: n_: support for making the parameters case insensitive in the storageclass?

cruwe: What does "n_" mean? Besides, added the general case. Which is more important to you and should accordingly come first? Or would you like me to drop the issue altogether?

kmova: nit pick comment.

pawanpraka1: @cruwe issue is fine. No worries there. Thanks for working on it.

PR #146, opened 2020-06-04
chore(changelog): add missing changelog for v0.8 release

pawanpraka1:


changelogs/released/122-w3aman
```diff
@@ -0,0 +1 @@
+Updated the doc for custom-topology support
```
kmova, line 1: shouldn't these be under `changelogs/unreleased` ?
If already  `released`, they should be under `changelogs/released/<release tag>`.

pawanpraka1: my bad. Did see that I added them in released folder. They are all unreleased changes. Fixed it.

PR #148, opened 2020-06-05
feat(modules): migrate to go modules and bump go version 1.14.4

prateekpandey14:
Signed-off-by: prateekpandey14 <prateek.pandey@mayadata.io>

### Why is this PR required? What issue does it fix?:
This PR enables the zfs localpv to be built using Go Modules

### Changes Made include

- changes in the Makefile to add go mod tidy, verify & vendor commands and GO111MODULE=on
- changes in travis.yml & travis.sh to enable its building from gomodules 
- adds vendor folder
- removes Gopkg lock & toml
- bump travis go version to 1.14.4 for better go module support

### Steps Performed
- Go mod init
- Go mod tidy
- Added go mod tidy and god mod verify in the Makefile
- Removed gopkg.toml and gopkg.lock
- Other fixes mentioned above
- go mod vendor


Makefile
```diff
@@ -114,13 +114,31 @@ test: format
...
+
+.PHONY: deps
+deps:
+	@echo "--> Tidying up submodules"
+	@go mod tidy
+	@echo "--> Verifying submodules"
+	@go mod verify
+
+.PHONY: verify-deps
+verify-deps: deps
```
akhilerm, line 126: question: @prateekpandey14  why is this check required?

prateekpandey14: to validate, if we missed to commit a updated mod files

.travis.yml
```diff
@@ -12,7 +12,7 @@ services:
   - docker
 language: go
 go:
-  - 1.12.5
+  - 1.14.x
```
kmova, line 15: n_: can this be 1.14.1 or a specific version

buildscripts/travis-build.sh
```diff
@@ -20,8 +20,7 @@ function checkGitDiff() {
 	if [[ `git diff --shortstat | wc -l` != 0 ]]; then echo "Some files got changed after $1";printf "\n";git diff;printf "\n"; exit 1; fi
 }
 
-#make golint-travis
```
kmova, line 23: Why is golint disabled?

prateekpandey14: forget to revert the change earlier , reverted now

.travis.yml
prateekpandey14: fixed to 1.14.4 , similar to maya repo

buildscripts/travis-build.sh
prateekpandey14: We don't have Makefile target for lint currently , so let it as same as ealier ,
later we can add `golangci-lint` for better linting support.

PR #151, opened 2020-06-08
feat(mount): moving to legacy mount

pawanpraka1:

## Pull Request template

Please, go through these steps before you submit a PR.

We can not mount the datasets to more than one path via zfs mount command, shifting to the legacy way of handling ZFS volumes where we can mount/umount the datasets via legacy mount and umount commands.

This will also add a building block for SINGLE-NODE-MULTI-WRITER Capability.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests
- [ ] (Optional) If documentation changes are required, which issue on

### DIFF
--- a/pkg/zfs/mount.go
+++ b/pkg/zfs/mount.go
@@ -3,6 +3,7 @@ package zfs
 import (
 	"fmt"
 	"os"
+	"os/exec"
 
 	"github.com/Sirupsen/logrus"
 	apis "github.com/openebs/zfs-localpv/pkg/apis/openebs.io/zfs/v1"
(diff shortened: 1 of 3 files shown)


pkg/zfs/mount.go
```diff
@@ -60,29 +61,28 @@ func UmountVolume(vol *apis.ZFSVolume, targetPath string,
...
-		}
-	} else {
-		if err = mounter.Unmount(targetPath); err != nil {
-			logrus.Errorf(
-				"zfspv failed to unmount zvol: path %s Error: %v",
-				targetPath, err,
-			)
-			return err
-		}
+	if err = mounter.Unmount(targetPath); err != nil {
```
kmova, line 64: q_: Is the case of volume (with dataset), invoking unmount prior to upgrade handled by this case?

pawanpraka1: > invoking unmount prior to upgrade
did not get the question, Prior to upgrade, no need to worry, as old code will handle the umount request via zfs umount command.

Here since umount works on dataset as well as zvol. So having one code to handle both the cases.

PR #162, opened 2020-06-25
feat(zfspv): adding backup and restore support

pawanpraka1:

This is the PR to handle backup and restore request at ZFS-LocalPV side. There will be changes at the velero plugin side also and the work is in progress for that.

Here velero plugin will create the backup CR to create a backup and ZFS-LocalPV driver will watch for that CR and create the back and transfer the data to the mentioned localtion. In the similar way to restore the data from the remote location, velero plugin will create the restore CR with the remote location information where the data is and ZFS-LocalPV will restore from that location.

**How to use?**:

Steps to use velero plugin for ZFS-LocalPV are :

1. install velero
```
velero install --provider aws --bucket velero --secret-file /home/pawan/pawan/credentials-minio --plugins velero/velero-plugin-for-aws:v1.0.0-beta.1 --backup-location-config region=minio,s3ForcePathStyle="true",s3Url=http://minio.velero.svc:9000 --use-volume-snapshots=true --use-restic

```
2. add openebs plugin
```
velero plugin add openebs/velero-plugin:latest
```
3. Create the volumesnapshot location :

for backup :-

```yaml
apiVersion: velero.io/v1
kind: VolumeSnapshotLocation
metadata:
  name: default
  namespace: velero
spec:
  provider: openebs.io/zfspv-blockstore
  config:
    bucket: velero
    prefix: zfs
    namespace: openebs # this is namespace where ZFS-LocalPV creates all the CRs, passed as OPENEBS_NAMESPACE env in the ZFS-LocalPV deployment
    provider: aws
    region: minio
    s3ForcePathStyle: "true"
    s3Url: http://minio.velero.svc:9000
```
4. Create backup
```
velero backup create my-backup --snapshot-volumes --include-namespaces=velero-ns --volume-snapshot-locations=aws-cloud-default --storage-location=default
```
5. Create Schedule
```
velero create schedule newschedule  --schedule="*/1 * * * *" --snapshot-volumes --include-namespaces=velero-ns --volume-snapshot-locations=aws-local-default --storage-location=default
```
6. Restore from backup
```
velero restore create --from-backup my-backup --restore-volumes=true --namespace-mappings velero-ns:ns-2
```

The PR in the Velero repo is here : https://github.com/openebs/velero-plugin/pull/102

no

**If the changes in this PR are manually verified, list down the scenarios covered and commands you used for testing with logs:**

- velero full backup and restore
- velero schedule(full) backup and restore
 
```yaml
apiVersion: velero.io/v1
kind: VolumeSnapshotLocation
metadata:
  name: default
  namespace: velero
spec:
  provider: openebs.io/zfspv-blockstore
  config:
    bucket: velero
    prefix: zfs
    namespace: openebs
    provider: aws
    region: minio
    s3ForcePathStyle: "true"
    s3Url: http://minio.velero.svc:9000
```

tested by doing backing up the full namespace, deleted it, and restored it and verified the data. 

**Any additional information for your reviewer?**:

1. added support for full backup
2. added support for scheduled(full) backup

**Limitation**:

if restoring in a different cluster, it should have same nodes and same zpool present.

**To Do**: 

incremental backup is supported fully. incremental restore is supported partially(with few limitation). Need to add full support for scheduled(incremental) restore.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests


kmova:
@pawanpraka1 @mynktl - can this limitation be eliminated, using some similar solution as was taken for restic local volumes?

```
if restoring in a different cluster, it should have same nodes and same zpool present.
```

pawanpraka1:
@kmova I am already working on that part https://github.com/openebs/zfs-localpv/issues/200.

deploy/yamls/zfsbackup-crd.yaml
```diff
@@ -0,0 +1,108 @@
...
+            backupDest:
+              description: BackupDest is the remote address for backup transfer
+              minLength: 1
+              type: string
+            ownerNodeID:
+              description: OwnerNodeID is a name of the nodes where the source volume
+                is
+              minLength: 1
+              type: string
+            prevSnapName:
```
kmova, line 74: q_: Are there names in sync with the proposed refactoring changes planned in velero plugin?

pawanpraka1: >proposed refactoring changes planned in velero plugin
not aware of this? can you share the proposal link.

pkg/driver/agent.go
```diff
@@ -65,6 +67,22 @@ func NewNode(d *CSIDriver) csi.NodeServer {
 		}
 	}()
 
+	// start the backup controller
+	go func() {
+		err := backup.Start(&ControllerMutex, stopCh)
+		if err != nil {
+			klog.Fatalf("Failed to start ZFS volume snapshot management controller: %s", err.Error())
```
kmova, line 74: n_: log message can probably say `volume snapshot backup controller` and the below one as `volume snapshot restore controller`.

pkg/zfs/volume.go
```diff
@@ -63,7 +67,7 @@ var (
 func init() {
 
 	OpenEBSNamespace = os.Getenv(OpenEBSNamespaceKey)
-	if OpenEBSNamespace == "" {
+	if OpenEBSNamespace == "" && os.Getenv("OPENEBS_NODE_DRIVER") != "" {
```
kmova, line 70: Is namespace only required for node driver?

pkg/zfs/volume.go
```diff
@@ -256,3 +260,44 @@ func RemoveSnapFinalizer(snap *apis.ZFSSnapshot) error {
...
+// RemoveBkpFinalizer removes finalizer from ZFSBackup CR
+func RemoveBkpFinalizer(bkp *apis.ZFSBackup) error {
+	bkp.Finalizers = nil
+
+	_, err := bkpbuilder.NewKubeclient().WithNamespace(OpenEBSNamespace).Update(bkp)
+	return err
+}
+
+// UpdateBkpInfo updates the backup info with the status
+func UpdateBkpInfo(bkp *apis.ZFSBackup, status apis.ZFSBackupStatus) error {
```
kmova, line 273: q_: Why are the finalizers being added to backup and not restore resources?

pkg/zfs/zfs_util.go
```diff
@@ -292,13 +296,56 @@ func buildVolumeResizeArgs(vol *apis.ZFSVolume) []string {
 	return ZFSVolArg
 }
 
+// builldVolumeBackupArgs returns volume send command for sending the zfs volume
+func buildVolumeBackupArgs(bkp *apis.ZFSBackup, vol *apis.ZFSVolume) []string {
+	var ZFSVolArg []string
+	backupDest := bkp.Spec.BackupDest
+
+	bkpAddr := strings.Split(backupDest, ":")
```
kmova, line 304: q_: how does this work with destination configured with credentials?

pkg/mgmt/restore/restore.go
```diff
@@ -0,0 +1,246 @@
...
+	c.workqueue.Add(key)
+}
+
+// synRestore is the function which tries to converge to a desired state for the
+// ZFSRestore
+func (c *RstrController) syncRestore(rstr *apis.ZFSRestore) error {
+	var err error = nil
+	// ZFSRestore should be deleted. Check if deletion timestamp is set
+	if !c.isDeletionCandidate(rstr) {
+		// if finalizer is not set then it means we are creating
```
kmova, line 81: comments need an update?

pkg/mgmt/restore/restore.go
```diff
@@ -0,0 +1,246 @@
+/*
```
kmova, line 1: Can we add a doc.go to this folder that lists the various low-level cases that are supported and any limitations? 

Few things that are not clear by going through the code easily are:
- how the restore differentiate between zvol and zfs datasets?
- how does the file-system information get passed on to the restore process?

For example: 

The restore flow is as follows:
- plugin creates a restore PV (is this correct?)
- plugin create a ZFS restore CR to restore the data from remote backup into the restore PV?
- restore controller (on node) keeps a a watch for new CRs associated with the node id. 
- if status != init or marked for deletion, restore controller will execute the zfs recv | create command. 

Limitations:
- If the restore fails due to network issues? Will it be re-attempted?
- If the restore doesn't have the specified backup, will it marked as failed?
- Similar to the need for nodeID being same, the expectation is also that capacity should be available on the dest node/pool?
- What happens when the same volume is restored twice or if the dest pool already has a volume with the same name.

pkg/mgmt/backup/backup.go
```diff
@@ -0,0 +1,250 @@
...
+}
+
+// synBkp is the function which tries to converge to a desired state for the
+// ZFSBackup
+func (c *BkpController) syncBkp(bkp *apis.ZFSBackup) error {
+	var err error = nil
+	// ZFSBackup should be deleted. Check if deletion timestamp is set
+	if c.isDeletionCandidate(bkp) {
+		// reconcile for the Destroy error
+		err = zfs.DestoryBackup(bkp)
```
kmova, line 81: Can this happen? looks like the finalizer gets added both for success/failed states. In case the failed state was due to backup not getting created, then destroy backup can fail too causing the backup  resource to be not deleted?

pkg/mgmt/backup/backup.go
```diff
@@ -0,0 +1,250 @@
...
+		if err == nil {
+			err = zfs.RemoveBkpFinalizer(bkp)
+		}
+	} else {
+		// if status is init then it means we are creating the zfs backup.
+		if bkp.Status == apis.BKPZFSStatusInit {
```
mynktl, line 87: With this condition backup resource needs to be created with init status. Should we add check here to execute backup if status is empty also? Or we need to mark empty status as invalid.

pkg/mgmt/restore/restore.go
```diff
@@ -0,0 +1,246 @@
...
+}
+
+// synRestore is the function which tries to converge to a desired state for the
+// ZFSRestore
+func (c *RstrController) syncRestore(rstr *apis.ZFSRestore) error {
+	var err error = nil
+	// ZFSRestore should be deleted. Check if deletion timestamp is set
+	if !c.isDeletionCandidate(rstr) {
+		// if finalizer is not set then it means we are creating
+		// the zfs backup.
```
mynktl, line 82: comment needs to be changed.

pkg/mgmt/backup/backup.go
mynktl: nit: else / if can be merged here

mynktl (commented):
changes looks good.

pkg/zfs/volume.go
pawanpraka1: node-driver env can be "agent" or "controller". So it it needed if we are using ZFS-LocalPV driver.

pkg/zfs/volume.go
pawanpraka1: as part of backup request, we have a dependant object, which is the snapshot. So when you clean up that, snapshot should also gets cleaned up. There is no dependant object for restore.

pkg/zfs/zfs_util.go
pawanpraka1: credentials or managed by velero. Here it is plain tcp connect to te velero plugin. Then velero plugin reads this stream and uploca/download it to the cloud.

pkg/mgmt/backup/backup.go
pawanpraka1: no, at the destory time we do a check for the backup resource, so destory will not fail.

pkg/mgmt/backup/backup.go
pawanpraka1: will do the validation in the CR. working on adding validation schema for this.

pkg/mgmt/backup/backup.go
```diff
@@ -0,0 +1,250 @@
...
+	bkp, ok := obj.(*apis.ZFSBackup)
+	if !ok {
+		runtime.HandleError(fmt.Errorf("Couldn't get bkp object %#v", obj))
+		return
+	}
+
+	if zfs.NodeID != bkp.Spec.OwnerNodeID {
+		return
+	}
+	klog.Infof("Got add event for Bkp %s snap %s@%s", bkp.Name, bkp.Spec.VolumeName, bkp.Spec.SnapName)
```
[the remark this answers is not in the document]
pawanpraka1: klog I guess does not have debug. May be later we can add verbose to these logs.

pkg/mgmt/restore/restore.go
pawanpraka1: added.

pkg/mgmt/backup/doc.go
```diff
@@ -0,0 +1,47 @@
...
+
+- plugin takes the backup of ZFSVolume CR so that it can be restored.
+
+- It will save the namespace information where the pvc is created also while taking the backup. Plugin will use this info if restoring without a namespace mapping to find if volume has already been restored.
+
+- plugin then creates the ZFSBackup CR with the destination volume and remote location where the data needs to be send.
+
+- Backup controller (on node) keeps a watch for new CRs associated with the node id. This node ID will be same as the Node ID present in the ZFSVolume resource.
+
+- The Backup controller will take a snapshot and then send the data to the remote location.
```
kmova, line 28: This one probably needs little clarity. remote location here refers to velero controller that receives the data and then pushes to S3.

pawanpraka1: right, updated it.

pkg/mgmt/backup/doc.go
```diff
@@ -0,0 +1,47 @@
...
+*/
+
+/*
+The Backup flow is as follows:
+
+- plugin takes the backup of ZFSVolume CR so that it can be restored.
+
+- It will save the namespace information where the pvc is created also while taking the backup. Plugin will use this info if restoring without a namespace mapping to find if volume has already been restored.
+
+- plugin then creates the ZFSBackup CR with the destination volume and remote location where the data needs to be send.
```
mynktl, line 24: better to mention, status should be `init`

pkg/mgmt/backup/doc.go
mynktl: better to remove `send the data to remote location` since it is part of next step and may create confusion.

mynktl (approved):

kmova (approved):

Issue #182, opened 2020-07-15
clone creation is failing on CentOS when fstype is xfs


w3aman:
zfspv-clone getting failed on centos based cluster when fstype is used as `xfs`
getting such error:
```
time="2020-07-15T09:10:51Z" level=error msg="zfspv: failed to mount volume /dev/zvol/zfs-test-pool/pvc-05e22f2a-9901-40ba-817d-82a43a922713 [xfs] to /var/lib/kubelet/pods/341b2fca-8430-45d1-9009-375a81588210/volumes/kubernetes.io~csi/pvc-05e22f2a-9901-40ba-817d-82a43a922713/mount, error mount failed: exit status 32\nMounting command: mount\nMounting arguments: -t xfs -o defaults /dev/zvol/zfs-test-pool/pvc-05e22f2a-9901-40ba-817d-82a43a922713 /var/lib/kubelet/pods/341b2fca-8430-45d1-9009-375a81588210/volumes/kubernetes.io~csi/pvc-05e22f2a-9901-40ba-817d-82a43a922713/mount\nOutput: mount: /var/lib/kubelet/pods/341b2fca-8430-45d1-9009-375a81588210/volumes/kubernetes.io~csi/pvc-05e22f2a-9901-40ba-817d-82a43a922713/mount: wrong fs type, bad option, bad superblock on /dev/zd16, missing codepage or helper program, or other error.\n\n"
```

and from the dmesg log it says :
```
[ 6421.004650] XFS (zd16): Filesystem has duplicate UUID 3af6518b-8f14-4585-b513-949053e63a4c - can't mount
[ 6421.603229] XFS (zd16): Filesystem has duplicate UUID 3af6518b-8f14-4585-b513-949053e63a4c - can't mount
[ 6422.705286] XFS (zd16): Filesystem has duplicate UUID 3af6518b-8f14-4585-b513-949053e63a4c - can't mount
[ 6424.812780] XFS (zd16): Filesystem has duplicate UUID 3af6518b-8f14-4585-b513-949053e63a4c - can't mount
[ 6428.937560] XFS (zd16): Filesystem has duplicate UUID 3af6518b-8f14-4585-b513-949053e63a4c - can't mount
[ 6437.072385] XFS (zd16): Filesystem has duplicate UUID 3af6518b-8f14-4585-b513-949053e63a4c - can't mount
[...]

PR #183, opened 2020-07-15
fix(xfs, uuid): fixed uuid generation issue when mount fails

pawanpraka1:
fixes: https://github.com/openebs/zfs-localpv/issues/182

This issue is specific to xfs only, when we create a clone volume and system is taking time in creating the device.

When we create a clone volume from a xfs filesystem, ZFS-LocalPV will go ahead and generate a new UUID for the clone volumes as we need a new UUID to mount the new clone filesystem. To generate a new UUID for the clone volume, ZFS-LocalPV first replays the xfs log by mounting the device to a tmp localtion.

Here, what is happening is since device creation is slow, so we went ahead and created the tmp location to mount the clone volume but since device has not created yet, the mount failed. In the next try since the tmp location is present, it will keep failing there only at every reconciliation time.

**What this PR does?**:

xfs UUID generation was not written for recociliation purpose, handled it so that uid generation works at the reconciliation time.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/zfs/xfs_util.go
+++ b/pkg/zfs/xfs_util.go
@@ -19,6 +19,7 @@ package zfs
 import (
 	"os"
 	"os/exec"
+	"path/filepath"
 
 	"strings"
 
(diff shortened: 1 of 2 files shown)


pkg/zfs/xfs_util.go
```diff
@@ -29,35 +30,54 @@ func xfsTempMount(volume string) error {
...
-	out, err := cmd.CombinedOutput()
+	/*
+	 * Device might have already mounted at the tmp path but umount might have failed
+	 * in previous attempt. Checking here if device is not mounted then only attempt
+	 * to mount it, otherwise proceed with the umount.
+	 */
+	curMounts, err := GetMounts(dev)
 	if err != nil {
-		klog.Errorf("xfs: failed to mount volume %s=>%s error: %s", device, tmpdir, string(out))
+		klog.Errorf("xfs: get mounts failed dev: %s err: %v", device, err.Error())
```
kmova, line 54: is it possible to check if the mount has the nouuid option?

pawanpraka1: Why do we need that? Can you elaborate a little bit? The driver makes sure that the volume can not be mounted by the pod until this operation completes successfully so it is not possible that the volume is mounted by someone in the meanwhile.
Also we are using TMP path for mounting it, which will make sure that it is mounted by this guy only otherwise mount/umount will fail for TMP path.

kmova: First off, I missed that this is a temporary mount. Was thinking of a stale mount case from the previous try - that was probably mounted without adding nouuid option. 

Is this the flow:
- clone with pv-uuidx
- mount /tmp/pv-uuidx (with nouuid)
- unmount /tmp/pv-uuidx
- proceed with mounting to pod/pv-uuidx

pawanpraka1: Yes @kmova , that is correct. while creating clone only we generate the new UUID. To do that  mount it at the temp location to replay the log.

Issue #163, opened 2020-06-26
readme need to be updated


w3aman:
**What steps did you take and what happened:**

I tried to apply the zfs-operator file on k8s version 1.15. but yaml was not applying giving out the error.
```
namespace/openebs created
customresourcedefinition.apiextensions.k8s.io/zfsvolumes.zfs.openebs.io created
customresourcedefinition.apiextensions.k8s.io/zfssnapshots.zfs.openebs.io created
error: error validating "https://raw.githubusercontent.com/openebs/zfs-localpv/master/deploy/zfs-operator.yaml": error validating data: ValidationError(CSIDriver.spec): unknown field "volumeLifecycleModes" in io.k8s.api.storage.v1beta1.CSIDriverSpec; if you choose to ignore these errors, turn validation off with --validate=false
```

**What did you expect to happen:**

As readme says supported on k8s 1.14+ but looks like `volumeLIfecycleModes` field works only after k8s 1.16
And
OS support is also mentioned only ubuntu 18.04 but i could see CentOS 7/8 support is now available.

So either have to fix the above error or update the readme accordingly.
say, 
k8s :  1.16+ and OS: ubuntu 18.04, 16.04, centos 7/8
                                            OR
fix this issue and k8s: 1.14+ and OS: ubuntu 18.04, 16.04, centsos 7/8

pawanpraka1:
We also need to improve the setup part and add details about how can we setup the zpool. Also, if possible, we can add support to automatically setup the pool.

PR #186, opened 2020-07-21
fix(yaml): removing volumeLifecycleModes from the operator yaml

pawanpraka1:
fixes: https://github.com/openebs/zfs-localpv/issues/163

This field was added in Kubernetes 1.16 and it informs Kubernetes about
the volume modes that are supported by the driver. The default is
"Persistent" if it is not used. 

**What this PR does?**:

This operator yaml will not work on k8s 1.14 and 1.15, since the driver supports
those k8s version so no need to mention volumeLifecycleModes in the operator as
the default is "Persistent". We broke this in https://github.com/openebs/zfs-localpv/pull/85.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/changelogs/unreleased/186-pawanpraka1
+++ b/changelogs/unreleased/186-pawanpraka1
@@ -1,0 +1,1 @@
+removing volumeLifecycleModes from the operator yaml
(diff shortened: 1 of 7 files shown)


kmova:
@pawanpraka1 -- should this change be added to the openebs/charts YAML files as well?

pawanpraka1:
yes @kmova , I will raise the same PR in the charts repo.

PR #187, opened 2020-07-21
chore(doc): adding supported features doc for each k8s version

pawanpraka1:

updating readme with the supported features for each k8s version, also updated the setup doc with ZPOOL setup information.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/docs/features.md
+++ b/docs/features.md
@@ -1,0 +1,8 @@
+### Features
+
+| k8s | Features |
+| :--- | :--- |
+| 1.14 | Volume Provisioning, Volume Deprovisioning, ZFS Property Update, Raw Block Volume |
+| 1.15 | Volume Provisioning, Volume Deprovisioning, ZFS Property Update, Raw Block Volume, Resize |
+| 1.16 | Volume Provisioning, Volume Deprovisioning, ZFS Property Update, Raw Block Volume, Resize, Snapshot |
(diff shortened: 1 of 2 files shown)


docs/features.md
```diff
@@ -0,0 +1,8 @@
+### Features
+
+| k8s | Features |
```
kmova, line 3: Let us change this table as follows:

| Feature | Minimum K8s version |  Minimum ZFS Local PV CSI driver version |

pawanpraka1: should we list alpha release also in "Minimum ZFS Local PV CSI driver version"?

kmova: thats a good point. I think it doesn't make sense. but probably will help provide clarity in case someone wants to upgrade from their current version. 

Also, how do we compare to this list here: https://github.com/container-storage-interface/spec/blob/master/lib/go/csi/csi.pb.go#L242

pawanpraka1: updated the details with the ZFS driver version.

>Also, how do we compare to this list here: 

That is the list of capabilities. This may not transalate to features, for example PUBLISH_UNPUBLISH is a capability which driver can decide whether it needs that or not. It may not convert as feature for the user. But it may be needed to provide the features like provisioning/deprovisioning of volumes.

PR #188, opened 2020-07-30
chore(doc): adding volume capacity roundoff details in readme

pawanpraka1:

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,7 +133,7 @@ parameters:
 provisioner: zfs.csi.openebs.io
 ```
 
-The storage class contains the volume parameters like recordsize, compression, dedup and fstype. You can select what are all
+The storage class contains the volume parameters like recordsize(should be power of 2), compression, dedup and fstype. You can select what are all
 parameters you want. In case, zfs properties paramenters are not provided, the volume will inherit the properties from the ZFS Pool.
 Also currently supported fs types are ext2/3/4, xfs and zfs only. The *poolname* is the must argument.
 Also there must be a ZPOOL running on *all the nodes* with the name given in the storage class.
@@ -161,7 +161,7 @@ parameters:
 provisioner: zfs.csi.openebs.io
 ```
 
-Here please note that we are providing `volblocksize` instead of `recordsize` since we will create a ZVOL, for which we can choose the blocksize with which we want to create the block device.
+Here please note that we are providing `volblocksize` instead of `recordsize` since we will create a ZVOL, for which we can choose the blocksize with which we want to create the block device. Here, please note that for ZFS, volblocksize should be power of 2.
 
 ##### zfs as FsType
 
@@ -185,7 +185,7 @@ parameters:
 provisioner: zfs.csi.openebs.io
 ```
 
-Here please note that we are providing `recordsize` which will be used to create the ZFS datasets, which specifies the maximum block size for files in the zfs file system.
+Here please note that we are providing `recordsize` which will be used to create the ZFS datasets, which specifies the maximum block size for files in the zfs file system. The recordsize has to be power of 2 for ZFS datasets.
 
 ##### ZPOOL Availability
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -237,6 +237,55 @@ spec:
...
+spec:
+  storageClassName: openebs-zfspv
+  accessModes:
+    - ReadWriteOnce
+  resources:
+    requests:
+      storage: 4G
+```
+
+Here note that storage is 4G. The volume size requested in PVC 4G (4 * 1000 * 1000 * 1024) is not multiple of 4k (4 * 1024, for zfs k is 1024). So, volume creation will fail. We can fix it by changing the storage size to 4Gi which is 4 * 1024 * 1024 * 1024, which is multiple of 4k (4 * 1024).
```
w3aman, line 270: 4G (4 *1000*1000*1000) **

README.md
```diff
@@ -237,6 +237,55 @@ spec:
...
+metadata:
+  name: openebs-zfspv
+parameters:
+  poolname: "zfspv-pool"
+provisioner: zfs.csi.openebs.io
+```
+
+Here PVC size should be multiple of 8k (8 * 1024).
+
+2. If storageclass has fstype as "zfs" then we don't need to worry about the size. We can use any size we want as in this case a dataset will be created not zvol.
```
w3aman, line 287: for zfs we can add, that recordsize field should be in power of 2.

pawanpraka1: as this section for PVC, added that detail at the other place.

README.md
```diff
@@ -237,6 +237,55 @@ spec:
...
+metadata:
+  name: openebs-zfspv
+parameters:
+  volblocksize: "4k"
+  fstype: "ext4"
+  poolname: "zfspv-pool"
+provisioner: zfs.csi.openebs.io
+```
+
+Here, PVC should have the storage multiple of volblocksize which is 4k (4 * 1024). If is not the multiple of volblocksize then ZFS will not create the volume and pod will not able able to come into RUNNING state as mount will keep on failing. For example, if we have created the below PVC with the above storageclass
```
w3aman, line 254: small typo ** will not able able ==> will not be able

README.md
```diff
@@ -237,6 +237,55 @@ spec:
...
+metadata:
+  name: openebs-zfspv
+parameters:
+  volblocksize: "4k"
+  fstype: "ext4"
+  poolname: "zfspv-pool"
+provisioner: zfs.csi.openebs.io
+```
+
+Here, PVC should have the storage multiple of volblocksize which is 4k (4 * 1024). If is not the multiple of volblocksize then ZFS will not create the volume and pod will not be able to come into RUNNING state as mount will keep on failing. For example, if we have created the below PVC with the above storageclass
```
kmova, line 254: @pawanpraka1 - can we someone avoid this manual calculation. When setting the value on the zfs, can we not round it up to the next highest number which  is a multiple of the block/record size?

pawanpraka1: that will violate the concept of pvc claim. Provisioned size will be more than claimed size. Provisioned size should be equal to PVC size. Doing round up would not be the right approach IMHO.

kmova: Hmm.. so it is about usability vs bytes. another case where this is accepted in practice is when we format a filesystem of 100G, it doesn't always give us 100G usable capacity, so a few bytes is ok in that perspective. 

It is probably harder to do the math that the size should be strictly a multiple of 1024. It has to be rounded off to the closest multiple, I am assuming the diff will be a max of record size == like 4k to 128k bytes difference. 

It is very common for users to request for size as: 
- 5G
- 5Gi

If either of them fails, it violates the ease of use principle.

pawanpraka1: @kmova I somehow don't feel that is the right thing. are we doing the same thing in cstor? How are we handeling it there?

pawanpraka1: as per discussion rounding off the size seems to be good.

kmova, line 254: This one needs an update based on the implementation change. https://github.com/openebs/zfs-localpv/pull/191

pawanpraka1: updated @kmova

README.md
```diff
@@ -237,6 +237,56 @@ spec:
 
 Create a PVC using the storage class created for the ZFS driver.
 
+Here, we have to note that the size will be rounded off to the nearest Mi or Gi unit. Please note that M/G notation uses 1000 base and Mi/Gi notation uses 1024 base, so 1M will be 1000 * 1000 byte and 1Mi will be 1024 * 1024.
```
kmova, line 240: We can probably push the detailed explanation into FAQ and just have a couple of lines about this in the main README.

PR #191, opened 2020-08-06
fix(zfspv): rounding off the volume size to Gi and Mi

pawanpraka1:
ref: https://github.com/openebs/zfs-localpv/pull/188

ZFS does not create the zvol if volume size is not multiple of
the volblocksize. There are use cases where customer will create
a PVC with size as 5G, which will be 5 * 1000 * 1000 * 1000 bytes
and this is not the multiple of default volblocksize 8k.

**What this PR does?**:
In ZFS, volblocksize and recordsize must be power of 2 from 512B to 1M,
so keeping the size in the form of Gi or Mi should be
sufficient to make volsize multiple of volblocksize/recordsize.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests


pkg/driver/controller_test.go
```diff
@@ -0,0 +1,43 @@
...
+
+	tests := map[string]struct {
+		input    int64
+		expected int64
+	}{
+		"Minimum allocatable is 1Mi": {input: 1, expected: Mi},
+		"roundOff to same Mi size":   {input: Mi, expected: Mi},
+		"roundOff to nearest Mi":     {input: Mi + 1, expected: Mi * 2},
+		"roundOff to same Gi size":   {input: Gi, expected: Gi},
+		"roundOff to nearest Gi":     {input: Gi + 1, expected: Gi * 2},
```
kmova, line 35: Can we add tests for G and M  also? What is the expected result for (G) "1000 * 1000 * 1000"?

pawanpraka1: 1G will be 1000341504 (i.e. 954Mi).

Issue #202, opened 2020-09-03
Current quay.io/openebs/zfs-driver:ci image is not provisioning on CentOS 8


jlcox1970:
**What steps did you take and what happened:**
cleaned up all openebs config in cluster, remove all zfs based pvc and reinstalled  using the README

**What did you expect to happen:**
expected to be able to run 'zfs' inside openebs-zfs-plugin container as per previous install

**The output of the following commands will help us better understand what's going on**:

zfs: error while loading shared libraries: libtirpc.so.3: cannot open shared object file: No such file or directory

**Anything else you would like to add:**
I have tried with the yaml from "Support for deploying to CentOS #119" but this still failed due to changing libs in Centos8

**Environment:**
- ZFS-LocalPV version 
 v0.9.2
- Kubernetes version (use `kubectl version`):
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.3", GitCommit:"2e7996e3e2712684bc73f0dec0200d64eec7fe40", GitTreeState:"clean", BuildDate:"2020-05-20T12:43:34Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}

- Kubernetes installer & version:
as above
- Cloud provider or hardware configuration:

bear metal, 
NAME                     STATUS   ROLES    AGE    VERSION   INTERNAL-IP   EXTERNAL-IP   OS-IMAGE                KERNEL-VERSION                 CONTAINER-RUNTIME
k8s-node-01              Ready    <none>   104d   v1.18.3   10.1.1.221    <none>        CentOS Linux 8 (Core)   4.18.0-147.8.1.el8_1.x86_64    docker://19.3.12

vm-host.globelock.home   Ready    master   104d   v1.18.3   10.1.1.220    <none>        CentOS Linux 8 (Core)   4.18.0-193.14.2.el8_2.x86_64   docker://19.3.12

- OS (e.g. from `/etc/os-release`):
NAME="CentOS Linux"
VERSION="8 (Core)"
ID="centos"
ID_LIKE="rhel fedora"
VERSION_ID="8"
PLATFORM_ID="platform:el8"
PRETTY_NAME="CentOS Linux 8 (Core)"
ANSI_COLOR="0;31"
CPE_NAME="cpe:/o:centos:centos:8"
HOME_URL="https://www.centos.org/"
BUG_REPORT_URL="https://bugs.centos.org/"

[...]

PR #204, opened 2020-09-03
feat(zfspv): mounting the root filesystem to remove the dependency on the Operating system

pawanpraka1:
fixes : https://github.com/openebs/zfs-localpv/issues/202 and probably https://github.com/openebs/zfs-localpv/issues/72 also.

We are mounting the individual library to run the zfs
binary inside the ZFS-LocalPV daemonset. The problem with this
is each OS has different sets of libraries. We need to have different
Operator yamls for different OS versions.

**What this PR does?**:

Here we are mounting the root directory inside the ZFS-LocalPV daemonset Pod
which does chroot to this path and run the command. As all the libraries will
be available which are present on the host inside the Pod, so we don't need to mount each
library here and also it will work for all the Operating systems.

To be on the safe side, we are mounting the host's root directory
as Readonly filesystem. Will cleanup the centos yamls once we have migrated to this in e2e.

Thanks @jlcox1970 for helping me verify this change on CentOS 8.

**Checklist:**
- [x] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/deploy/yamls/ubuntu/zfs-driver.yaml
+++ b/deploy/yamls/ubuntu/zfs-driver.yaml
@@ -705,7 +705,7 @@ kind: ConfigMap
 apiVersion: v1
 metadata:
   name: openebs-zfspv-bin
-  namespace: kube-system
+  namespace: kube-system # should be the same namespace where it is getting mounted
 data:
   zfs: |
     #!/bin/sh
--- a/deploy/yamls/ubuntu/zfs-driver.yaml
+++ b/deploy/yamls/ubuntu/zfs-driver.yaml
@@ -701,6 +701,24 @@ roleRef:
 
 ---
 
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: openebs-zfspv-bin
+  namespace: kube-system
+data:
+  zfs: |
+    #!/bin/sh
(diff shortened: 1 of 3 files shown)


prateekpandey14 (approved):
looks good to me , do we want to archive the other distribution centric operators in upcoming PR ?

pawanpraka1:
Yes @prateekpandey14 , once we move e2e test cases to point to this yaml, I will cleanup other yamls.

mittachaitu (approved):
changes are good

kmova (approved):

deploy/yamls/ubuntu/zfs-driver.yaml
```diff
@@ -701,6 +701,24 @@ roleRef:
 
 ---
 
+kind: ConfigMap
+apiVersion: v1
+metadata:
+  name: openebs-zfspv-bin
+  namespace: kube-system
```
kmova, line 708: @pawanpraka1 can you add comments as to why this should be in kube-system namespace.

pawanpraka1: added @kmova

w3aman (approved):

PR #212, opened 2020-09-18
chore(doc): updating the doc with supported storageclass parameters

pawanpraka1:


docs/storageclasses.md
```diff
@@ -0,0 +1,219 @@
...
+The ZFS-LocalPV driver will create the Volume in the Pool “zfspv-pool” present on the node with fewer of volumes provisioned among the given node list. In the above StorageClass, if there provisioned volumes on node-1 are less, it will create the volume on node-1 only. Alternatively, we can use `volumeBindingMode: WaitForFirstConsumer` to let the k8s select the node where the volume should be provisioned.
+
+The problem with the above StorageClass is that it works fine if the number of nodes is less, but if the number of nodes is huge, it is cumbersome to list all the nodes like this. In that case, what we can do is, we can label all the similar nodes using the same key value and use that label to create the StorageClass.
+
+pawan@pawan-master:~/pawan$ kubectl label node pawan-node-2 openebs.io/zpool=nvme
+node/pawan-node-2 labeled
+pawan@pawan-master:~/pawan$ kubectl label node pawan-node-1 openebs.io/zpool=nvme
+node/pawan-node-1 labeled
+
+Now, restart the ZFS-LocalPV Driver (if already deployed otherwise, please ignore) so that it can pick the new node label as the supported topology.
```
w3aman, line 193: just thinking should we give the link pointing to faq where we describe this custom-topology and need of driver restart in deep.

(if already deployed otherwise, please ignore) ===> (if already deployed**,** otherwise please ignore)

docs/storageclasses.md
```diff
@@ -0,0 +1,219 @@
...
+
+$ kubectl delete po -n kube-system -l role=openebs-zfs
```
w3aman, line 195: put this in ```

docs/storageclasses.md
```diff
@@ -0,0 +1,219 @@
...
+     - node-2
+```
+
+Here we can have ZFS Pool of name “zfspv-pool” created on the nvme disks and want to use this high performing ZFS Pool for the applications that need higher IOPS. We can use the above SorageClass to create the PVC and deploy the application using that.
+
+The ZFS-LocalPV driver will create the Volume in the Pool “zfspv-pool” present on the node with fewer of volumes provisioned among the given node list. In the above StorageClass, if there provisioned volumes on node-1 are less, it will create the volume on node-1 only. Alternatively, we can use `volumeBindingMode: WaitForFirstConsumer` to let the k8s select the node where the volume should be provisioned.
+
+The problem with the above StorageClass is that it works fine if the number of nodes is less, but if the number of nodes is huge, it is cumbersome to list all the nodes like this. In that case, what we can do is, we can label all the similar nodes using the same key value and use that label to create the StorageClass.
+
+pawan@pawan-master:~/pawan$ kubectl label node pawan-node-2 openebs.io/zpool=nvme
```
w3aman, line 188: put this in ```

docs/storageclasses.md
```diff
@@ -0,0 +1,219 @@
...
+
+### recordsize (*optional* parameter)
+
+This parameter is application if fstype provided is "zfs" otherwise it will be ignored. It specifies a suggested block size for files in the file system.
+
+allowed values: Any power of 2 from 512 bytes to 128 Kbytes
+
+### volblocksize (*optional* parameter)
+
+This parameter is application if fstype is anyhting but "zfs" where we create a ZVOL a raw block device carved out of ZFS Pool. It specifies the block size to use for the zvol. The volume size can only be set to a multiple of volblocksize, and cannot be zero.
```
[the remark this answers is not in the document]
pawanpraka1: anything but "zfs" means that only (not).

docs/storageclasses.md
pawanpraka1: good point. added here.

w3aman (approved):

kmova (approved):

PR #230, opened 2020-10-20
feat(upgrade): support parallel/faster upgrades for node daemonset

pawanpraka1:

For ZFSPV, all the node daemonset pods can go into the terminating state at
the same time since it does not need any minimum availability of those pods.

**What this PR does?**:

Changing maxUnavailable to 100% so that K8s can upgrade all the daemonset
pods parallelly. Also added labels to all the pods.

### DIFF
--- a/changelogs/unreleased/230-pawanpraka1
+++ b/changelogs/unreleased/230-pawanpraka1
@@ -1,0 +1,1 @@
+support parallel/faster upgrades for node daemonset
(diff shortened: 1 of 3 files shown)


prateekpandey14 (approved):
lgtm

kmova:
@pawanpraka1 -- has upgrade from previously installed setup to this new operator.yaml with changes tested? In the past I had seen that when upgrade options are changed, we might have to delete some objects and then patch.

pawanpraka1:
@kmova, I have tested the upgrade on my Rancher setup, did not face any issue. Could you elaborate more about the issue, which object we need to delete and why?

kmova:
usually changing the default value of upgrade type, requires removing that json object from the deployment spec and recreating. Had faced this issue when changin from rollingUpgrade to recreate. In this case as it is not changing, it should be fine.

kmova (approved):

Issue #214, opened 2020-09-18
add the deployment via kustomize


pawanpraka1:
Add support to deploy ZFS-LocalPV via kustomize (https://github.com/kubernetes-sigs/kustomize).

codegagan:
Looking into this.

pawanpraka1:
@spencergilbert could you take a look at the PR(https://github.com/openebs/zfs-localpv/pull/231) and let us know if there is anything else needed.

PR #231, opened 2020-10-27
feat(kustomize): adding deployment via kustomize

pawanpraka1:
fixes : https://github.com/openebs/zfs-localpv/issues/214


codegagan:
Should not this be added in ci to test deployment via (-k) and then run all/some sanity tests ?

spencergilbert (approved):
Looks straight forward :+1:

pawanpraka1:
Thanks @spencergilbert .

>Should not this be added in ci to test deployment via (-k) and then run all/some sanity tests ?

@codegagan will you be able to work on that part?

deploy/yamls/kustomization.yaml
```diff
@@ -0,0 +1,10 @@
+images:
+  - name: quay.io/openebs/zfs-driver
+    newTag: ci
+resources:
+- namespace.yaml
+- zfsvolume-crd.yaml
+- zfssnapshot-crd.yaml
+- zfsbackup-crd.yaml
+- zfsrestore-crd.yaml
+- zfs-driver.yaml
```
prateekpandey14, line 10: Do we need to update README with the command to use the kustomize based install using `-k` with `kubectl apply` ?

prateekpandey14 (approved):
looks good, a small comment for adding install cmd

deploy/yamls/kustomization.yaml
pawanpraka1: added.

kmova (approved):

PR #232, opened 2020-11-02
feat(sanity): adding CSI Sanity test

pawanpraka1:

**What this PR does?**:

adding CSI sanity test for ZFS-LocalPV

**Changes?**:

1. convert volume name and snapshot name to lowercase as they are needed to create a CR and uppercase characters are not allowed
2.  returning proper error code in the code
3. added "wait" functionality to wait for volume/operation creation, needed since sanity does not have retry logic
4. add functionality to pass "node" where PV should be created, needed since sanity does not pass topology information
5. added few checks to return correct error codes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -235,6 +235,12 @@ zfs-driver-image: zfs-driver
 ci:
 	@echo "--> Running ci test";
 	$(PWD)/ci/ci-test.sh
+
+.PHONY: sanity
+sanity:
+	@echo "--> Running CSI Sanity test";
+	$(PWD)/ci/sanity.sh
+
 # Push images
 deploy-images:
 	@DIMAGE="${IMAGE_ORG}/zfs-driver" ./buildscripts/push
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -118,6 +118,21 @@ func waitForReadyVolume(volname string) error {
 	return nil
 }
 
+func waitForVolDestroy(volname string) error {
+	for true {
+		_, err := zfs.GetZFSVolume(volname)
+		if err != nil {
+			if k8serror.IsNotFound(err) {
+				return nil
+			}
+			return status.Errorf(codes.Internal,
+				"zfs: destroy wait failed, not able to get the volume %s %s", volname, err.Error())
+		}
+		time.Sleep(time.Second)
+	}
+	return nil
+}
+
 // CreateZFSVolume create new zfs volume from csi volume request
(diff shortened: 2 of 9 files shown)


Makefile
```diff
@@ -235,6 +235,12 @@ zfs-driver-image: zfs-driver
 ci:
 	@echo "--> Running ci test";
 	$(PWD)/ci/ci-test.sh
+
+.PHONY: sanity
+sanity:
+	@echo "--> Running ci test";
```
kmova, line 241: n_: running csi-sanity tests

ci/ci-test.sh
```diff
@@ -89,7 +89,7 @@ set +e
 
 echo "running ginkgo test case"
 
-ginkgo -v
+# ginkgo -v
```
kmova, line 92: n_ remove it.

pkg/driver/controller.go
```diff
@@ -126,17 +159,39 @@ func CreateZFSVolume(req *csi.CreateVolumeRequest) (string, error) {
...
+			}
+		} else {
+			if vol.Spec.Capacity != capacity {
+				return "", status.Errorf(codes.AlreadyExists,
+					"volume %s already present", volName)
+			}
+			return vol.Spec.OwnerNodeID, nil
+		}
+	}
+
```
payes, line 179: Should we add a return for errors other than NotFound.

pawanpraka1: this is Create request. We don't need to return any other error as we are creating the volume. The above check is added for sanity to check if volume is already present with the same capacity, then return successful response.

payes (approved):

pkg/driver/controller.go
```diff
@@ -169,12 +224,18 @@ func CreateZFSVolume(req *csi.CreateVolumeRequest) (string, error) {
...
 	return selected, nil
 }
 
 // CreateVolClone creates the clone from a volume
 func CreateVolClone(req *csi.CreateVolumeRequest, srcVol string) (string, error) {
-	volName := req.GetName()
+	volName := strings.ToLower(req.GetName())
```
kmova, line 238: With this change are there any backward compatiblity check that need to be verified. Like, creating a snapshot or clone in older release and after upgrading trying to create new snapshot or delete old snap or clone. 

If the ids were always - pvc-uuid.. may be it is not an issue. just checking here.

pawanpraka1: Yes, it was always pvc-uuid.

kmova (approved):

PR #234, opened 2020-11-10
feat(clone): add support for creating the Clone from volume as datasource

pawanpraka1:

 add support for creating the Clone from volume as datasource

**What this PR does?**:

We can now create the Clone from pvc directly

```
kind: PersistentVolumeClaim
apiVersion: v1
metadata:
  name: pvc-clone
spec:
  storageClassName: openebs-clone
  dataSource:
    name: pvc-vol
    kind: PersistentVolumeClaim
  accessModes:
    - ReadWriteOnce
  resources:
    requests:
      storage: 4Gi 
```
The ZFS_LocalPV driver will create one internal snapshot of the name
same as the new volume name and will create a clone out of it. Also,
while destroying the volume the driver will take care of deleting
the created snapshot for the clone.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -165,8 +165,56 @@ func CreateZFSVolume(req *csi.CreateVolumeRequest) (string, error) {
 	return selected, nil
 }
 
-// CreateZFSClone create a clone of zfs volume
-func CreateZFSClone(req *csi.CreateVolumeRequest, snapshot string) (string, error) {
+// CreateVolClone creates the clone from a volume
+func CreateVolClone(req *csi.CreateVolumeRequest, srcVol string) (string, error) {
+	volName := req.GetName()
+	parameters := req.GetParameters()
+	// lower case keys, cf CreateZFSVolume()
+	pool := helpers.GetInsensitiveParameter(&parameters, "poolname")
+	size := getRoundedCapacity(req.GetCapacityRange().RequiredBytes)
+	volsize := strconv.FormatInt(int64(size), 10)
+
+	vol, err := zfs.GetZFSVolume(srcVol)
+	if err != nil {
+		return "", status.Error(codes.NotFound, err.Error())
+	}
+
+	if vol.Spec.PoolName != pool {
+		return "", status.Errorf(codes.Internal,
+			"clone: different pool src pool %s dst pool %s",
+			vol.Spec.PoolName, pool)
+	}
+
+	if vol.Spec.Capacity != volsize {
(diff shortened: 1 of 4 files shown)


kmova:
@pawanpraka1 - what happens in the following case: 

- create pvc1
- create pvc2 ( clone of pvc1)
- delete pvc1 

Will pvc2 (and its PV) still exist?

pawanpraka1:
Yes @kmova, pvc2 and it's pv will be there. Also zfs volume created for pvc1 will also be there and it will be cleaned up automatically when we delete pvc2.

pkg/driver/controller.go
```diff
@@ -165,8 +165,56 @@ func CreateZFSVolume(req *csi.CreateVolumeRequest) (string, error) {
...
+			vol.Spec.PoolName, pool)
+	}
+
+	if vol.Spec.Capacity != volsize {
+		return "", status.Error(codes.Internal, "clone: volume size is not matching")
+	}
+
+	selected := vol.Spec.OwnerNodeID
+
+	labels := map[string]string{zfs.ZFSVolKey: vol.Name}
```
kmova, line 194: Can we make this a more explicit key - like "source-zfs-volume"?

pkg/zfs/zfs_util.go
```diff
@@ -580,6 +600,27 @@ func DestroyVolume(vol *apis.ZFSVolume) error {
...
+		// datasource is volume, delete the dependent snapshot
+		snap := &apis.ZFSSnapshot{}
+		snap.Name = vol.Name // snapname is same as volname
+		snap.Spec = vol.Spec
+		// add src vol name
+		snap.Labels = map[string]string{ZFSVolKey: srcVol}
+
+		klog.Infof("destroying snapshot %s@%s for the clone %s", srcVol, snap.Name, volume)
+
+		err := DestroySnapshot(snap)
```
kmova, line 614: Here is where the source volume needs to be cleaned up, in case this was the last clone/snapshot being deleted?

pawanpraka1: no, not for the last clone/snapshot being deleted, every clone will have a corresponsing snapshot. So when we are deleting the clone volume, we should delete the corresponding snapshot here.

kmova (approved):
this PR will be followed up documentation updates and better log management in case of error cases.

pawanpraka1:
@kmova here is the issue https://github.com/openebs/zfs-localpv/issues/236

PR #253, opened 2020-12-08
fix(doc): fixed the kustomization yaml file path in readme

w3aman:
Signed-off-by: Aman Gupta <aman.gupta@mayadata.io>

- Fix the kustomization yaml file path in readme


pawanpraka1 (approved):
Looks good.

PR #255, opened 2020-12-10
feat(image): moving to ubuntu bionic(18.04 LTS) docker image

pawanpraka1:
build is failing on ubuntu 19.10 as the ca-certificate package has been obsolete. Switching to 18.04 as now since we are doing root mount, we can use any ubuntu image.

### DIFF
--- a/buildscripts/zfs-driver/Dockerfile
+++ b/buildscripts/zfs-driver/Dockerfile
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu:19.10
+FROM ubuntu:18.04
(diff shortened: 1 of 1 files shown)


buildscripts/zfs-driver/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-FROM ubuntu:19.10
+FROM ubuntu:18.04
```
kmova, line 15: Why was it set to 19.10? was there some dependency that we needed earlier?

pawanpraka1: earlier there was libc issue, so it was set to the 19.10. but now with the root mount we don't need that.

kmova (approved):

PR #256, opened 2020-12-10
fix(sanity): fixing flaky sanity test case 

pawanpraka1:
flaky sanity fix and updated the docker image to 18.04 for github action, missed adding this in the PR https://github.com/openebs/zfs-localpv/pull/255.


prateekpandey14 (approved):
Lgtm, even we can think of moving to ubuntu 20.04 in coming future ?

pawanpraka1:
@prateekpandey14  we did that and faced mount related issue with 20,04 so reverted back to 19.10. If that is stable, we can move to it.

kmova (approved):

PR #260, opened 2020-12-15
fix(mount): fixing idempotency check for the mount path

pawanpraka1:

fixing idempotency check for the mount path

**What this PR does?**:

For shared mount, the idempotency check was not happening. So checking the mounts for shared volume also.

no

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/zfs/mount.go
+++ b/pkg/zfs/mount.go
@@ -144,24 +144,28 @@ func verifyMountRequest(vol *apis.ZFSVolume, mountpath string) (bool, error) {
 		return false, status.Errorf(codes.Internal, "verifyMount: GetVolumePath failed %s", err.Error())
 	}
 
-	// if it is not a shared volume, then make sure it is not mounted to more than one path
-	if vol.Spec.Shared != "yes" {
-		/*
-		 * This check is the famous *Wall Of North*
-		 * It will not let the volume to be mounted
-		 * at more than two places. The volume should
-		 * be unmounted before proceeding to the mount
-		 * operation.
-		 */
-		currentMounts, err := mnt.GetMounts(devicePath)
-		if err != nil {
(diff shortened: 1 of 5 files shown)


pkg/zfs/mount.go
```diff
@@ -144,24 +144,28 @@ func verifyMountRequest(vol *apis.ZFSVolume, mountpath string) (bool, error) {
...
+	 * operation.
+	 */
+	currentMounts, err := mnt.GetMounts(devicePath)
+	if err != nil {
+		klog.Errorf("can not get mounts for volume:%s dev %s err: %v",
+			vol.Name, devicePath, err.Error())
+		return false, status.Errorf(codes.Internal, "verifyMount: Getmounts failed %s", err.Error())
+	} else if len(currentMounts) >= 1 {
+		// if device is already mounted at the mount point, return successful
+		for _, mp := range currentMounts {
```
kmova, line 161: In this case, will it return true even if more than one node has mounted it? Should the count and shared check prior to this?

kmova: Thinking out loud. What should be the outcome with and without shared in the following cases? 

- not mounted 
- mounted at a different node 
- mounted at a different node and expected node 
- mounted on multiple nodes 
- mounted on expected node

pawanpraka1: This is a local mount only. The volume will.be mounted by multiple pods on the same node. The kubelet can ask to mount for the same path (idempotency).  So the scenario is 
1. Is shared is false, the. If it is not mounted at the same path, we should return error.
2. If shared is true, we should check if it is already mounted and return true. Other wise we should go ahead and mount again to the new path.

kmova (approved):

PR #262, opened 2020-12-17
fix(mount): creating directory with 0755 permission

pawanpraka1:


pkg/zfs/mount.go
```diff
@@ -254,16 +254,10 @@ func MountDataset(vol *apis.ZFSVolume, mount *MountInfo) error {
 
 // MountFilesystem mounts the disk to the specified path
 func MountFilesystem(vol *apis.ZFSVolume, mount *MountInfo) error {
-	if err := os.MkdirAll(mount.MountPath, 0000); err != nil {
+	if err := os.MkdirAll(mount.MountPath, 0755); err != nil {
```
kmova, line 257: n_: can we add a comment on why 0755 is being used as the default permissions?

pawanpraka1: added the comment @kmova

kmova (approved):

Issue #215, opened 2020-09-18
add capacity based scheduling in ZFS-LocalPV


pawanpraka1:
Currently the ZFS-LocalPV supports volume count based scheduling, where it creates the PV where less number of volumes are provisioned (https://github.com/openebs/zfs-localpv/blob/master/pkg/driver/scheduler.go#L70).  We need to add capacity based scheduling where the volume should be provisioned where more space is available in the ZFS Pool.

PR #266, opened 2020-12-28
feat(schd): adding capacity weighted scheduler

pawanpraka1:
The ZFS Driver will use capacity scheduler to pick a node
which has less capacity occupied by the volumes. Making this
as default scheduler as it is better than the volume count based
scheduling. We can use below storageclass to specify the scheduler
```yaml
apiVersion: storage.k8s.io/v1
kind: StorageClass
metadata:
 name: openebs-zfspv
allowVolumeExpansion: true
parameters:
 scheduler: "CapacityWeighted"
 poolname: "zfspv-pool"
```

Please Note that after the upgrade, there will be a change in the behavior.
If we are not using `scheduler` parameter in the storage class then after
the upgrade ZFS Driver will pick the node based on volume capacity weight
instead of the count.

fixes: https://github.com/openebs/zfs-localpv/issues/215

**Explanation**

ZFS Driver does not control the zfs pool, by checking volumes capacity on the node for a particular pool can get us close to balancing the volumes across the nodes. So let us say that we have three pools on 3 different nodes node1, node2 and node3, with initial volumes provisioned like this =>
```
node       -     node1    -     node2       -     node3 
volume     -     vol1     -     vol2        -    vol3
size       -     100G     -     5G          -    50G
```
Now, if we want to provision a volume a new volume, the driver will pick the node2 for provisioning the volume. It will keep on picking the node2 until the volume nodes are equally distributed capacity wise.

### DIFF
--- a/pkg/driver/schd_helper.go
+++ b/pkg/driver/schd_helper.go
@@ -19,6 +19,7 @@ package driver
 import (
 	"github.com/openebs/zfs-localpv/pkg/builder/volbuilder"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"strconv"
 
 	zfs "github.com/openebs/zfs-localpv/pkg/zfs"
 )
@@ -26,12 +27,15 @@ import (
 // scheduling algorithm constants
 const (
 	// pick the node where less volumes are provisioned for the given pool
-	// this will be the default scheduler when none provided
 	VolumeWeighted = "VolumeWeighted"
+
+	// pick the node where total provisioned volumes have occupied less capacity from the given pool
+	// this will be the default scheduler when none provided
+	CapacityWeighted = "CapacityWeighted"
 )
 
 // getVolumeWeightedMap goes through all the pools on all the nodes
-// and creats the node mapping of the volume for all the nodes.
+// and creates the node mapping of the volume for all the nodes.
 // It returns a map which has nodes as key and volumes present
 // on the nodes as corresponding value.
 func getVolumeWeightedMap(pool string) (map[string]int64, error) {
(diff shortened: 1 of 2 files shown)


pkg/driver/schd_helper.go
```diff
@@ -56,12 +60,44 @@ func getVolumeWeightedMap(pool string) (map[string]int64, error) {
...
+
+	if err != nil {
+		return nmap, err
+	}
+
+	// create the map of the volume capacity
+	// for the given pool
+	for _, zv := range zvlist.Items {
+		if zv.Spec.PoolName == pool {
+			volsize, err := strconv.ParseInt(zv.Spec.Capacity, 10, 64)
```
kmova, line 83: Is this going to get the actual used capacity or reserved capacity? 

For example, if the volumes are provisioned as follows on two nodes with 100GB pools:
- v1 = 50G => node 1 (this volume writes only 5mb of data)
- v2 = 50G => node 2 (this volume writes 1GB of data)

- v3 = 50G
- v4 = 50G

Will v3 and v4 go to node1 because it has occupied less capacity then node2?

pawanpraka1: it is based on pv size. So ,in the above case v3 and v4 will go to different nodes (node1 and node2).

praveengt: Kiran, for Thick provisioned volumes does the actual space usage even matters.

kmova: nope. for thick provisioned, it doesn't matter. using based on PV size is good.

kmova (approved):

PR #270, opened 2021-01-05
feat(zfspv): try volume creation on all the nodes

pawanpraka1:

pvc gets bound even zfs volume creation keeps on failing on the node. 

**What this PR does?**:

Currently scheduler returns one node and the provisioner uses that node to provision the volume. Once scheduler returns the node list in the priority order instead of returning just one node, the CSI controller can attempt the volume creation on all the nodes. Here, in this case if the volume is not created, the pvc will not be bound as the controller will wait to the volume to be ready or failed. Also, the the node agent will not reconcile for the creation of the volume, it will attempt to create once and fail if not successful. The csi provisioner will keep on requesting to create the volume in case of failure.

Depends on https://github.com/openebs/lib-csi/pull/2

**Known Issue?**:

if volume creation request in progress and we are deleting the pvc, CSI plugin does not get the delete call and if volume creation succeeds, there will be one stale volume in the system.

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -101,23 +101,6 @@ func getRoundedCapacity(size int64) int64 {
 	return ((size + Mi - 1) / Mi) * Mi
 }
 
-func waitForReadyVolume(volname string) error {
-	for true {
-		vol, err := zfs.GetZFSVolume(volname)
-		if err != nil {
-			return status.Errorf(codes.Internal,
-				"zfs: wait failed, not able to get the volume %s %s", volname, err.Error())
-		}
-
-		switch vol.Status.State {
-		case zfs.ZFSStatusReady:
-			return nil
-		}
-		time.Sleep(time.Second)
-	}
-	return nil
-}
-
 func waitForVolDestroy(volname string) error {
 	for true {
 		_, err := zfs.GetZFSVolume(volname)
@@ -129,6 +112,7 @@ func waitForVolDestroy(volname string) error {
 				"zfs: destroy wait failed, not able to get the volume %s %s", volname, err.Error())
 		}
 		time.Sleep(time.Second)
+		klog.Infof("waiting for volume to be destroyed %s", volname)
 	}
 	return nil
 }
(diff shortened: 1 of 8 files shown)


pkg/zfs/volume.go
```diff
@@ -80,6 +83,79 @@ func init() {
...
+			// delete the snapshot
+			if err := DeleteSnapshot(snapname); err != nil {
+				klog.Errorf(" not able to clean up the snapshot %s", snapname)
+			}
+
+			return fmt.Errorf("not able to create the snapshot")
+		}
+		time.Sleep(time.Second)
+	}
+	return fmt.Errorf("timeout, snapshot is being created")
```
kmova, line 156: q_: will this get executed? how does the timeout work?

pawanpraka1: still experimenting it,I have two design consideration
1. block forever till the volume is created
2. have a timeout and fail the volume creation if it is not created within the timeout period.

Have not finalized it yet as there are complications involved. It is still work in progress.

pawanpraka1:
cc: @iyashu @praveengt @abhranilc

pkg/driver/controller.go
```diff
@@ -235,13 +223,25 @@ func CreateZFSVolume(req *csi.CreateVolumeRequest) (string, error) {
...
 	}
 
-	err = zfs.ProvisionVolume(volObj)
+	klog.Infof("zfs: trying volume creation %s/%s on node %s", pool, volName, prfList)
+
+	// try volume creation sequentially on all nodes
+	for _, node := range prfList {
+		vol, _ := volbuilder.BuildFrom(volObj).WithOwnerNode(node).WithVolumeStatus(zfs.ZFSStatusPending).Build()
+
+		err = zfs.ProvisionVolume(vol)
```
kmova, line 232: @pawanpraka1 - overall the code seems to be good. 

In addition to the corner case mentioned in the PR (possibility of stale volumes), how about:
- cases, where the list returned, can get outdated because of parallel volume creation/deletion 
- validating the node itself is still intact while going through the list. How long do we wait before we determine that the current node selected is not suitable. 

I am now wondering if we should add an option where the provisioner can ask only for one node at a time and if the provisioning fails, return back so that CSI retries?

pawanpraka1: > cases, where the list returned, can get outdated because of parallel volume creation/deletion

That should not be a problem here as the goal is to avoid getting stuck on a single node. So, even in this case scheduling might not be perfect but volume will get provision.

> validating the node itself is still intact while going through the list. How long do we wait before we determine that the current node selected is not suitable.

Yes, I think we should add wait here. Will add that.

> I am now wondering if we should add an option where the provisioner can ask only for one node at a time and if the provisioning fails, return back so that CSI retries?

That option is already there and controlled by allowedtopologies, it will try on the nodes which satisfy the topology contraints.

kmova (approved):

Issue #302, opened 2021-04-01
Prevent deletion of ZFS dataset


centromere:
**Describe the problem/challenge you have**
When a ZFSVolume Kubernetes resource is deleted, its corresponding dataset on the host is also deleted. There is no way to disable this functionality.

**Describe the solution you'd like**
I would like a flag to be added to the ZFSVolume resource which indicates that the underlying dataset is not to be deleted when the resource is deleted.

**Environment:**
- ZFS-LocalPV version: 1.5.0
- Kubernetes version: 1.20.5
- Kubernetes installer & version: kubeadm
- Cloud provider or hardware configuration: Bare metal
- OS: Ubuntu 20.04

PR #303, opened 2021-04-03
feat(zfspv): remove finalizer that is owned by ZFS-LocalPV

pawanpraka1:
fixes: https://github.com/openebs/zfs-localpv/issues/302

User might want to add extra protection on the ZFSVolume CR and wants to add finalizers on it. The current code does not allow the user to setup the user's finalizers on the ZFSVolume CR.
 
**What this PR does?**:

We set the Finalizer to nil while handling the delete event, instead,
we should try to destroy the volume when there are no user finalizers
set. User might have added his own finalizers and we should not try to destroy
the volumes util those user finalizers are removed.

PR handles upgrade scenario.

### DIFF
--- a/pkg/zfs/volume.go
+++ b/pkg/zfs/volume.go
@@ -233,10 +233,6 @@ func UpdateZvolInfo(vol *apis.ZFSVolume, status string) error {
 	finalizers := []string{}
 	labels := map[string]string{ZFSNodeKey: NodeID}
 
-	if vol.Finalizers != nil {
-		return nil
-	}
-
 	switch status {
 	case ZFSStatusReady:
 		finalizers = append(finalizers, ZFSFinalizer)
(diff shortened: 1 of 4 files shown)


akhilerm (commented):
@pawanpraka1 We need to make this same change in LVM localpv right?

pkg/zfs/volume.go
```diff
@@ -358,3 +350,32 @@ func UpdateRestoreInfo(rstr *apis.ZFSRestore, status apis.ZFSRestoreStatus) erro
...
+	for _, fin := range finalizers {
+		if fin != ZFSFinalizer {
+			userFin = append(userFin, fin)
+		}
+	}
+	return userFin
+}
+
+// IsVolumeReady returns true if volume is Ready
+func IsVolumeReady(vol *apis.ZFSVolume) bool {
```
akhilerm, line 367: :+1:

pawanpraka1:
> We need to make this same change in LVM localpv right?

@akhilerm yes, it is good to have this change in LVM LocalPV repo also.

pkg/zfs/volume.go
```diff
@@ -290,10 +286,6 @@ func UpdateSnapInfo(snap *apis.ZFSSnapshot) error {
 	finalizers := []string{ZFSFinalizer}
 	labels := map[string]string{ZFSNodeKey: NodeID}
 
-	if snap.Finalizers != nil {
```
akhilerm, line 293: Why was this removed? If user has added a custom finalizer and snapshot is not ready, will this mark the snapshot CR as ready?

pawanpraka1: if this check is there, snapshot will never be ready, so we have to remove as it was assumed that no one will create the resources manually.

akhilerm (approved):
LGTM

PR #304, opened 2021-04-06
feat(migration): adding support to migrate the PV to a new node

pawanpraka1:

Here we are setting the node affinity on PV using nodename, so we can not migrate the pods to the new node as k8s scheduler will try to schedule on the old node only.

**What this PR does?**:

Here, we can now label the nodes with the unique values using the key `openebs.io/nodeid`. The ZFS LocalPV driver will pick the value from the nodes and set the affinity accordingly.

Now to migrate the PV to the other node, we have to move the disks to the other node and remove the old node from the cluster and set the same label on the new node using the same key and value, which will let k8s scheduler to schedule the pods to the that node.

**Refactoring?**:

We are using nodename as NodeID in code, refactored the code and made it NodeName to have better code readability. Also changed the ZFSVolume display string to `NodeID` from the strting `Node` to tell that it is the NodeID.

**Limitation?**:

we can only move the PVs to the new node, we can not move the PVs to the node which was already used in the cluster as there is only one allowed value for the custom key for setting the node label.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -88,7 +88,7 @@ func run(config *config.Config) {
 
 	klog.Infof("ZFS Driver Version :- %s - commit :- %s", version.Current(), version.GetGitCommit())
 	klog.Infof(
-		"DriverName: %s Plugin: %s EndPoint: %s NodeID: %s",
+		"DriverName: %s Plugin: %s EndPoint: %s Node Name: %s",
 		config.DriverName,
 		config.PluginType,
 		config.Endpoint,
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -55,7 +55,7 @@ func main() {
 	cmd.Flags().AddGoFlagSet(flag.CommandLine)
 
 	cmd.PersistentFlags().StringVar(
-		&config.NodeID, "nodeid", zfs.NodeID, "NodeID to identify the node running this driver",
+		&config.Nodename, "nodename", zfs.NodeID, "Nodename to identify the node running this driver",
 	)
 
 	cmd.PersistentFlags().StringVar(
@@ -92,7 +92,7 @@ func run(config *config.Config) {
 		config.DriverName,
 		config.PluginType,
 		config.Endpoint,
-		config.NodeID,
+		config.Nodename,
 	)
 
 	err := driver.New(config).Run()
--- a/deploy/yamls/zfs-driver.yaml
+++ b/deploy/yamls/zfs-driver.yaml
@@ -805,6 +805,8 @@ spec:
               value: "zfs-operator"
             - name: OPENEBS_IO_ENABLE_ANALYTICS
               value: "true"
+            - name: NODE_AFFINITY_KEY
+              value: kubernetes.io/hostname
           args :
             - "--endpoint=$(OPENEBS_CSI_ENDPOINT)"
             - "--plugin=$(OPENEBS_CONTROLLER_DRIVER)"
@@ -997,11 +999,11 @@ spec:
           image: openebs/zfs-driver:ci
           imagePullPolicy: IfNotPresent
           args:
-            - "--nodeid=$(OPENEBS_NODE_ID)"
+            - "--nodeid=$(OPENEBS_NODE_NAME)"
             - "--endpoint=$(OPENEBS_CSI_ENDPOINT)"
             - "--plugin=$(OPENEBS_NODE_DRIVER)"
           env:
-            - name: OPENEBS_NODE_ID
+            - name: OPENEBS_NODE_NAME
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
@@ -1011,6 +1013,8 @@ spec:
               value: agent
             - name: OPENEBS_NAMESPACE
               value: openebs
+            - name: NODE_AFFINITY_KEY
+              value: kubernetes.io/hostname
           volumeMounts:
             - name: plugin-dir
               mountPath: /plugin
--- a/deploy/yamls/zfs-driver.yaml
+++ b/deploy/yamls/zfs-driver.yaml
@@ -999,7 +999,7 @@ spec:
           image: openebs/zfs-driver:ci
           imagePullPolicy: IfNotPresent
           args:
-            - "--nodeid=$(OPENEBS_NODE_NAME)"
+            - "--nodename=$(OPENEBS_NODE_NAME)"
             - "--endpoint=$(OPENEBS_CSI_ENDPOINT)"
             - "--plugin=$(OPENEBS_NODE_DRIVER)"
           env:
--- a/deploy/yamls/zfs-driver.yaml
+++ b/deploy/yamls/zfs-driver.yaml
@@ -805,8 +805,6 @@ spec:
               value: "zfs-operator"
             - name: OPENEBS_IO_ENABLE_ANALYTICS
               value: "true"
-            - name: NODE_AFFINITY_KEY
-              value: kubernetes.io/hostname
           args :
             - "--endpoint=$(OPENEBS_CSI_ENDPOINT)"
             - "--plugin=$(OPENEBS_CONTROLLER_DRIVER)"
@@ -1013,8 +1011,6 @@ spec:
               value: agent
             - name: OPENEBS_NAMESPACE
               value: openebs
-            - name: NODE_AFFINITY_KEY
-              value: kubernetes.io/hostname
           volumeMounts:
             - name: plugin-dir
               mountPath: /plugin
--- a/deploy/zfs-operator.yaml
+++ b/deploy/zfs-operator.yaml
@@ -58,7 +58,7 @@ spec:
       type: string
     - description: Node where the volume is created
       jsonPath: .spec.ownerNodeID
-      name: Node
+      name: NodeID
       type: string
     - description: Size of the volume
       jsonPath: .spec.capacity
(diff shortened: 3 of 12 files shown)


cmd/main.go
```diff
@@ -88,7 +88,7 @@ func run(config *config.Config) {
 
 	klog.Infof("ZFS Driver Version :- %s - commit :- %s", version.Current(), version.GetGitCommit())
 	klog.Infof(
-		"DriverName: %s Plugin: %s EndPoint: %s NodeID: %s",
+		"DriverName: %s Plugin: %s EndPoint: %s Node Name: %s",
```
akhilerm, line 91: `NodeID` is used [here](https://github.com/openebs/zfs-localpv/pull/304/files#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R95)

deploy/yamls/zfs-driver.yaml
```diff
@@ -805,6 +805,8 @@ spec:
               value: "zfs-operator"
             - name: OPENEBS_IO_ENABLE_ANALYTICS
               value: "true"
+            - name: NODE_AFFINITY_KEY
+              value: kubernetes.io/hostname
```
akhilerm, line 809: Is this the label key that we are using? It is added to each node by kubernetes right? So how does the labelling a new node work when another node goes down?

cmd/main.go
pawanpraka1: it is config.NodeID, which is node name only.

deploy/yamls/zfs-driver.yaml
pawanpraka1: it is the default key used by the driver. User has to label and provide that key here. See the docs added.

deploy/zfs-operator.yaml
```diff
@@ -2204,11 +2206,11 @@ spec:
           image: openebs/zfs-driver:ci
           imagePullPolicy: IfNotPresent
           args:
-            - "--nodeid=$(OPENEBS_NODE_ID)"
+            - "--nodeid=$(OPENEBS_NODE_NAME)"
```
akhilerm, line 2209: Is it node name or node ID as the arg?

akhilerm (commented):
given some comments

deploy/yamls/zfs-driver.yaml
akhilerm: If the default key is used, PV migration cannot be done right?

pawanpraka1: Yep

deploy/zfs-operator.yaml
pawanpraka1: Yes, better to rename the arg here, will do that. Thanks

shubham14bajpai (approved):
LGTM

docs/faq.md
```diff
@@ -224,3 +224,34 @@ Then driver will find the nearest size in Mi, the size allocated will be ((1G +
 
 PVC size as zero in not a valid capacity. The minimum allocatable size for the ZFS-LocalPV driver is 1Mi, which means that if we are requesting 1 byte of storage space then 1Mi will be allocated for the volume.
 
+### 8. How to migrate PVs to the new node in case old node is not accessible
+
+The ZFS-LocalPV driver will set affinity on the PV to make the volume stick to the node so that pod gets scheduled to that node only where the volume is present. Now, the problem here is, when that node is not accesible due to some reason and we move the disks to a new node and import the pool there, the pods will not be scheduled to this node as k8s scheduler will be looking for that node only to schedule the pod.
+
+Now, with the release of 1.7.0, ZFS-LocalPV driver has the ability to use the user defined affinity for creating the PV. While deploying the ZFS-LocalPV driver, we should label all the nodes first using the key `openebs.io/nodeid` with some unique value
```
akhilerm, line 231:
```suggestion
From release 1.7.0 of ZFS-LocalPV, the driver has the ability to use the user defined affinity for creating the PV. While deploying the ZFS-LocalPV driver, first we should label all the nodes using the key `openebs.io/nodeid` with some unique value.
```

docs/faq.md
```diff
@@ -224,3 +224,34 @@ Then driver will find the nearest size in Mi, the size allocated will be ((1G +
...
+2. add a new node in the cluster
+3. move the disks to this new node
+4. import the zfs pools on the new nodes
+5. label the new node with same key and value. For example, if we have removed the node-3 from the cluster and added node-4 as new node, we have to label the node `node-4` and set the value to `custom-value-3` as shown below
+
+```
+$ kubectl label node node-4 openebs.io/nodeid=custom-value-3
+```
+
+Once the above steps are done, the pod should be able to run on this new node with all the data it has on the old node. Here, there is one limitation that we can only move the PVs to the new node, we can not move the PVs to the node which was already used in the cluster as there is only one allowed value for the custom key for setting the node label.
```
akhilerm, line 257:
```suggestion
Once the above steps are done, the pod should be able to run on this new node with all the data it has on the old node. Here, there is one limitation that we can only move the PVs to the new node, we can not move the PVs to the node which were already used in the cluster as there is only one allowed value for the custom key for setting the node label.
```
This line is a little bit confusing. `we can move PVs, or cannot move PVs?`

pkg/zfs/volume.go
```diff
@@ -70,19 +71,45 @@ var (
...
+		}
+		nodename := os.Getenv("OPENEBS_NODE_NAME")
+		if nodename == "" {
+			klog.Fatalf("OPENEBS_NODE_NAME environment variable not set")
+		}
+		if NodeID, err = GetNodeID(nodename); err != nil {
+			klog.Fatalf("GetNodeID failed for node=%s err: %s", nodename, err.Error())
+		}
+		klog.Infof("zfs: node(%s) has node affinity %s=%s", nodename, ZFSTopologyKey, NodeID)
+	} else if os.Getenv("OPENEBS_CONTROLLER_DRIVER") != "" {
```
akhilerm, line 90:
```suggestion
	} else if os.Getenv("OPENEBS_NAMESPACE") != "" {
```

akhilerm (commented):
@pawanpraka1 Given some comments

pkg/zfs/volume.go
shubham14bajpai: There is another if below it

pawanpraka1: @akhilerm, we need to check for namespace only if is a controller.

docs/faq.md
pawanpraka1: we can move the pv to the new node.
We can not move the pv to the existing(old) node.

akhilerm: oh. ohkay. In that case, this suggestion is not required. will change it.

docs/faq.md
```diff
@@ -224,3 +224,34 @@ Then driver will find the nearest size in Mi, the size allocated will be ((1G +
...
+2. add a new node in the cluster
+3. move the disks to this new node
+4. import the zfs pools on the new nodes
+5. label the new node with same key and value. For example, if we have removed the node-3 from the cluster and added node-4 as new node, we have to label the node `node-4` and set the value to `custom-value-3` as shown below
+
+```
+$ kubectl label node node-4 openebs.io/nodeid=custom-value-3
+```
+
+Once the above steps are done, the pod should be able to run on this new node with all the data it has on the old node. Here, there is one limitation that we can only move the PVs to the new node, we can not move the PVs to the node which were already used in the cluster as there is only one allowed value for the custom key for setting the node label.
```
akhilerm, line 257:
```suggestion
Once the above steps are done, the pod should be able to run on this new node with all the data it has on the old node. Here, there is one limitation that we can only move the PVs to the new node, we can not move the PVs to the node which was already used in the cluster as there is only one allowed value for the custom key for setting the node label.
```

akhilerm (approved):
LGTM

kmova:
Few high level questions. 

=> Is this feature tied to a fixed label called "openebs.io/nodeid" ? Or can the user decide a custom label set by their infra scipts like "coolplatform.io/node-id"?

=> The PR has some changes related to replacing NODEID with NODENAME. Can you update a note in the PR description on why this refactoring was done.

pawanpraka1:
@kmova 


> => Is this feature tied to a fixed label called "openebs.io/nodeid" ? Or can the user decide a custom label set by their infra tools "coolplatform.io/node-id"?

The earlier implementation was doing that only where users have to label the nodes with the custom key and then they have to mention the key the operator yaml. It seems like a lot of work for the users and does not look very user friendly. Later I made the change to fix the key (https://github.com/openebs/zfs-localpv/pull/304/commits/dce4b0c6e3f9aca5c69df863099bc9db803e70f9) so that user does not need to do anything else. We just need to lable the nodes with the fixed key.

> => The PR has some changes related to replacing NODEID with NODENAME. Can you update a note in the PR description on why this refactoring was done.

done

kmova:
Had a chat with @pawanpraka1 regarding the customization of the label. 

In the current implementation, the driver will automatically set the node labels using node name as a unique identifier. This approach is backward compatible and users don't need to take any action. Going ahead with merging this solution. 

This will be followed up by further discussions and design around - making the unique label configurable. We have to decide whether it is a good idea to make this configurable at install or storage class (if this can be done). Based on the feedback on the current implementation we will have to see if we can design a solution that can also be used for migrating volumes to an existing node.

kmova (approved):

Issue #291, opened 2021-02-16
utilize zfs replication to migrate PVs between nodes


artw:
zfs replication is pretty unique tool in the block storage world of OSS, why not utilize it for something awesome like that.

Some annotation to PVC, that would trigger a snapshot/replicate/delete source/ modify or recreate PV object.

pawanpraka1:
@artw you mean asynchronous replication? Why not use the velero and create the schedule to take the periodic backup?

artw:
> @artw you mean asynchronous replication? Why not use the velero and create the schedule to take the periodic backup? 
> 
> 
> 
> Check this to migrate the pv to the other node https://github.com/openebs/zfs-localpv/blob/master/docs/backup-restore.md#restore-on-a-different-node.

I do, and it works quite well, thanks.

It would be much more effective to move the data between nodes directly utilizing the openebs-zfs-node daemonset. 

To think of it further, you could even do it regulary to have HA PVs mirrored between all (or some) nodes 🤔

pawanpraka1:
@artw, sorry, I missed this thread. So basically, we need a data migration framework from one node to other without using velero. This can be done easily, as matter of the fact, velero already uses node daemonsets to backup/restore the data. We just need one operator which can help us to migrate the data from one node to other. This feature is in our roadmap.

mazocode:
@pawanpraka1 It would be awesome to have something like syncoid integrated. This would make synchronization and moving PVs  between nodes a breeze. We are using kubevirt which does not support live migration of local PV volumes but with something like syncoid it would require a quick reboot only. We're currently doing this for backups (snapshot replication of the PVs to dedicated backup servers).

almereyda:
There is PR about the subject, which is not yet linked from here, nor does itself link to here:

- #336

It could possibly be considered a follow-up to @pawanpraka1 previous comment above.

avishnu:
This will be picked up for discussion after the current backlog list is tackled. Community contributions will be definitely welcome.

almereyda:
Incus has a fairly established Go implementation of migrating ZFS datasets between Incus hosts. Eventually it is of interest for research to some here.

<details><summary>Details</summary>
<p>

- [How to move or copy storage volumes - Incus documentation](https://linuxcontainers.org/incus/docs/main/howto/storage_move_volume/)
- [How to move existing Incus instances between servers - Incus documentation](https://linuxcontainers.org/incus/docs/main/howto/move_instances/)
[...]

PR #336, opened 2021-05-20
chore(design): adding pv migration proposal

pawanpraka1:


design/pv-migration.md
```diff
@@ -0,0 +1,86 @@
...
+
+### Migrator
+
+ZFS POOL name should be same across all the nodes for ZFS-LocalPV. So, we have to rename the ZFS POOLs if we are moving it to any existing node. We need a Migrator workflow to update the POOL name in the ZFSVolume object. This will find all the volumes present in a ZFS POOL on that node and updates the ZFSVolume object with the correct PoolName.
+
+**Note:** We can not edit PV volumeAttributes with the new pool name as it is immutable field.
+
+The migrator will look for all the volumes start with the name "pvc-" and will look for corresponding ZFSVolume object and will update it. In case of admin has imported the existing volumes where it is possible that volume name might not start with the name "pvc-", the PoolName needs to be updated manually as the volume was created manually by the admin.
```
mittachaitu, line 52: To figure out `zvols` managed by CSI-driver can we set some customer ZFS property? So that it will reduce some manual steps.

design/pv-migration.md
```diff
@@ -0,0 +1,86 @@
...
+$ kubectl get node pawan-node-1 --show-labels
+NAME           STATUS   ROLES    AGE    VERSION   LABELS
+node-1   Ready    worker   351d   v1.17.4   beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,kubernetes.io/arch=amd64,kubernetes.io/hostname=node-1,kubernetes.io/os=linux,node-role.kubernetes.io/worker=true,openebs.io/nodeid=node1,openebs.io/nodename=node-1,openebs.io/pool1=true,openebs.io/pool2=true
+```
+- If we are moving the pool1 from node1 to node2, then there are two cases here :-
+
+#### 1. if node2 is a fresh node
+
+- we can simply import the pool and restart the ZFS-LocalPV driver to make it aware of that pool to set the corresponding node topology
+- the ZFS-LocalPV driver will look for `openebs.io/pool1=true` and removes the label from the nodes where pool is not present
```
mittachaitu, line 68: How can we ensure it will unlabel the correct node in case multiple nodes have the same pool name?

design/pv-migration.md
```diff
@@ -0,0 +1,86 @@
...
+
+The problem with the above approach is we can not move the volumes to any existing node as we can set only one label on the node for a given key. So, for any existing node in a cluster, label `openebs.io/nodeid` will already be set so, we can not update it to a new value as the pods already running on it will not able to get scheduled here.
+
+
+## Proposal
+
+### Keys Per ZPOOL
+
+We are proposing to have a key dedicated to ZFS POOL. This key will be used by the ZFS-LocalPV driver to set the label on the nodes where it is present. In this way we can allow the ZFS POOLs to move from any node to any other node as the key is tied to the ZFS POOL as opposed to keeping it per node. We are proposing to have a `openebs.io/<poolname>=true` label on the node where the pool is present. Assuming admins do not have large number of pools on a node, there will be not much label set on a node.
```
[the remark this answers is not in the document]
pawanpraka1: Yes, that is correct. It is supposed to be openebs.io/<pool guid>. Will fix the doc.

design/pv-migration.md
pawanpraka1: It is openebs.io/pool-guid. So node agent knows which pools it own. And can take the decision.

design/pv-migration.md
pawanpraka1: That manual step is still needed after that as volume is imported by admin manually.

mittachaitu: Yes, that is true admin needs to manually import the volume into CSI-Driver and it should be one-time activity. IMO for the sake of completeness, it is something like If the pool disk is moved to a different node then our CSI-Driver should take care of all the volumes which were previously managed by it irrespective of whether it is manually added by admin (or) created by K8s. 

> In case of admin has imported the existing volumes where it is possible that volume name might not start with the name "pvc-", the PoolName needs to be updated manually as the volume was created manually by the admin. 

To avoid the above manual step, Once the volume is handover to the CSI-Driver, we can set some custom `zfs property` (zfs set <openebs.io:is_managed=true>) indicating volume is managed by OpenEBS CSI-Driver and it will prevent the above additional manual steps when the disk is migrated to a different node.

pawanpraka1: @mittachaitu I would not recommend to set anything on the zpool as we don't touch anything there and would like to keep it that way. Now coming to this point, we already have ZFSVolume CR, using that we can know that which volumes are managed by ZFSLocalPV driver.

mittachaitu: Ok, I am trying to propose to set a property on zfs/zvol volume dataset, not on the pool. But yeah, other way around is also possible i.e managing based on the ZFSVolume CR so that we can that manual step.

pawanpraka1:
cc: @iyashu Please take a look.

design/pv-migration.md
```diff
@@ -0,0 +1,86 @@
...
+
+The problem with the above approach is we can not move the volumes to any existing node as we can set only one label on the node for a given key. So, for any existing node in a cluster, label `openebs.io/nodeid` will already be set so, we can not update it to a new value as the pods already running on it will not able to get scheduled here.
+
+
+## Proposal
+
+### Keys Per ZPOOL
+
+We are proposing to have a key dedicated to ZFS POOL. This key will be used by the ZFS-LocalPV driver to set the label on the nodes where it is present. In this way we can allow the ZFS POOLs to move from any node to any other node as the key is tied to the ZFS POOL as opposed to keeping it per node. We are proposing to have a `guid.openebs.io/<pool-guid>=true` label on the node where the pool is present. Assuming admins do not have large number of pools on a node, there will be not much label set on a node.
```
mittachaitu, line 44: How about following the naming convention `guid.zfspool.openebs.io/<pool-guid>=true`. Open for discussion on a naming convention

mittachaitu (approved):
LGTM

design/pv-migration.md
akhilerm: `guid.zfs.openebs.io/<pool-guid>=true`. can also be considered, since we already have `zfs.openebs.io`

design/pv-migration.md
```diff
@@ -0,0 +1,87 @@
...
+$ kubectl get node pawan-node-1 --show-labels
+NAME           STATUS   ROLES    AGE    VERSION   LABELS
+node-1   Ready    worker   351d   v1.17.4   beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,kubernetes.io/arch=amd64,kubernetes.io/hostname=node-1,kubernetes.io/os=linux,node-role.kubernetes.io/worker=true,openebs.io/nodeid=node1,openebs.io/nodename=node-1,guid.openebs.io/14820954593456176137=true,guid.openebs.io/16291571091328403547=true
+```
+- If we are moving the pool1 from node1 to node2, then there are two cases here :-
+
+#### 1. if node2 is a fresh node
+
+- we can simply import the pool and restart the ZFS-LocalPV driver to make it aware of that pool to set the corresponding node topology
+- the ZFS-LocalPV driver will look for `guid.openebs.io/14820954593456176137=true` and will remove the label from the nodes where pool is not present
```
akhilerm, line 68: Is this done by the controller plugin or the node plugin?

design/pv-migration.md
```diff
@@ -0,0 +1,87 @@
...
+- the ZFS-LocalPV driver will update the new node with `guid.openebs.io/14820954593456176137=true` label
+- the migrator will look for ZFSVolume resource and update the OwnerNodeID with the new node id for all the volumes.
+- the k8s scheduler will be able to see the new label and should schedule the pods to this new node.
```
akhilerm, line 71: Does the user need to delete the old pod on the old node, which is no longer Running,? or will it be automatically scheduled to the new node?

design/pv-migration.md
```diff
@@ -0,0 +1,87 @@
...
+
+#### 2. if node2 is existing node and Pool of the same name is present there
+
+- here we need import the pool with the different name and restart the ZFS-LocalPV driver to make it aware of that pool to set the corresponding node topology
+- the ZFS-LocalPV driver will look for `guid.openebs.io/14820954593456176137=true` and will remove the label from the nodes where the pool is not present
+- the ZFS-LocalPV driver will update the new node with `guid.openebs.io/14820954593456176137=true` label
+- the migrator will look for ZFSVolume resource and update the PoolName and OwnerNodeID for all the volumes.
```
akhilerm, line 78: Is this migrator a separate job? as in the openebs/upgrade, openebs/migrate jobs?

akhilerm (commented):
@pawanpraka1 Have a few questions on the design

design/pv-migration.md
pawanpraka1: it can be seperator job also, it is one time activity which can be done at the driver registration time.

design/pv-migration.md
pawanpraka1: no, the pod will automatically scheduled to the new node.

design/pv-migration.md
pawanpraka1: it is done by node agent.

design/pv-migration.md
pawanpraka1: thinking of making it `guid.zpool.openebs.io/<pool-guid>=true` as it specifically says that it is the guid of zpool. Will wait for others to comment on this one.

prateekpandey14: `guid.zpool.openebs.io/<pool-guid>=true` make more sense to me.

design/pv-migration.md
```diff
@@ -0,0 +1,87 @@
...
+
+- we can simply import the pool and restart the ZFS-LocalPV driver to make it aware of that pool to set the corresponding node topology
+- the ZFS-LocalPV driver will look for `guid.openebs.io/14820954593456176137=true` and will remove the label from the nodes where pool is not present
+- the ZFS-LocalPV driver will update the new node with `guid.openebs.io/14820954593456176137=true` label
+- the migrator will look for ZFSVolume resource and update the OwnerNodeID with the new node id for all the volumes.
+- the k8s scheduler will be able to see the new label and should schedule the pods to this new node.
+
+#### 2. if node2 is existing node and Pool of the same name is present there
+
+- here we need import the pool with the different name and restart the ZFS-LocalPV driver to make it aware of that pool to set the corresponding node topology
```
prateekpandey14, line 75: nit: Here we need to import ?

design/pv-migration.md
prateekpandey14: Does app pod rescheduled instantly or do we need to wait for pod timeout (~300sec default toleration) ?

prateekpandey14 (approved):
lgtm

design/pv-migration.md
pawanpraka1: it will let k8s scheduler to take that decesion which will happen after pod timeout.

akhilerm (approved):
LGTM

design/pv-migration.md
```diff
@@ -0,0 +1,87 @@
...
+- here we need to import the pool with the different name and restart the ZFS-LocalPV driver to make it aware of that pool to set the corresponding node topology
+- the ZFS-LocalPV driver will look for `guid.openebs.io/14820954593456176137=true` and will remove the label from the nodes where the pool is not present
+- the ZFS-LocalPV driver will update the new node with `guid.openebs.io/14820954593456176137=true` label
+- the migrator will look for ZFSVolume resource and update the PoolName and OwnerNodeID for all the volumes.
+- the k8s scheduler will be able to see the new label and should schedule the pods to this new node.
+
+## Implementation Plan
+
+### Phase 1
+1. Implement replacement with fresh node
```
kmova, line 84: Do we first need to support StorageClass having a zpool name pattern? prior to implementing this approach?

design/pv-migration.md
kmova: `guid.zpool.openebs.io/<pool-guid>=true` is fine. 

But if we need to maintain consistency we can also use `zpool-guid.zfs.openebs.io/<pool-guid>=true`

Also, considering how the LVM will make use of this field can also help standardize the parameter.

design/pv-migration.md
```diff
@@ -0,0 +1,87 @@
...
+NAME           STATUS   ROLES    AGE    VERSION   LABELS
+node-1   Ready    worker   351d   v1.17.4   beta.kubernetes.io/arch=amd64,beta.kubernetes.io/os=linux,kubernetes.io/arch=amd64,kubernetes.io/hostname=node-1,kubernetes.io/os=linux,node-role.kubernetes.io/worker=true,openebs.io/nodeid=node1,openebs.io/nodename=node-1,guid.openebs.io/14820954593456176137=true,guid.openebs.io/16291571091328403547=true
+```
+- If we are moving the pool1 from node1 to node2, then there are two cases here :-
+
+#### 1. if node2 is a fresh node
+
+- we can simply import the pool and restart the ZFS-LocalPV driver to make it aware of that pool to set the corresponding node topology
+- the ZFS-LocalPV driver will look for `guid.openebs.io/14820954593456176137=true` and will remove the label from the nodes where pool is not present
+- the ZFS-LocalPV driver will update the new node with `guid.openebs.io/14820954593456176137=true` label
```
kmova, line 69: q_: How to avoid a race conditions where two nodes in the cluster will have the same label? Also consider the cases like:

* Node is shutdown or in not ready state and the pools have been moved to new node.

design/pv-migration.md
pawanpraka1: no need. That is altogether different feature.

design/pv-migration.md
```diff
@@ -0,0 +1,87 @@
...
+ZFS POOL name should be same across all the nodes for ZFS-LocalPV. So, we have to rename the ZFS POOLs if we are moving it to any existing node. We need a Migrator workflow to update the POOL name in the ZFSVolume object. This will find all the volumes present in a ZFS POOL on that node and updates the ZFSVolume object with the correct PoolName.
+
+**Note:** We can not edit PV volumeAttributes with the new pool name as it is immutable field.
+
+The migrator will look for all the volumes for all the pools present on the node and will look for corresponding ZFSVolume object and will update it with the correct poolname and ownernodeId field.
+
+### Workflow
+
+- user will setup all the nodes and setup the ZFS pool on each of those nodes.
+- the ZFS-LocalPV CSI driver will look for all the pools on the node and will set the `guid.openebs.io/<pool-guid>=true` label for all ZFS POOLs that is present on that node. Let's say node-1 has two pools(say pool1 with guid as 14820954593456176137 and pool2 with guid as 16291571091328403547) present then the labels will be like this :
```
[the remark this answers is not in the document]
pawanpraka1: there are cloud providers which have some limit like gcp has limit of 64.

design/pv-migration.md
pawanpraka1: previous step handles that, it has to make sure there are no node with the same label.

design/pv-migration.md
kmova: ok. then this feature will only work by moving the pool to new node without any pools on it.

design/pv-migration.md
kmova: in the previous steps - ZFS - Local PV driver - is it the node driver running on the old node or new node?

pawanpraka1: Yes, the node driver will remove/set the label.

design/pv-migration.md
pawanpraka1: It can have pools. Why do you ask that question, could you elaborate?

design/pv-migration.md
kmova: in uninstall, cases - will the labels be cleared from nodes.

pawanpraka1: do we need that, at the installation time, node label will be handle/moved accordingly.

almereyda:
Is there still something missing from this proposal, which blocks it from being merged?

Else it could be a first step towards an implementation of #291. Merging it would signal that this feature is still considered, while the discussion is stalled, that could be continued in pair with a reference implementation.

sinhaashish:
@pawanpraka1  It would be great if you contribute this feature(code) to this repo. We are our hands full of other openebs engines work at the moment, and i don't see we will be able to code for this improvement.
Please contribute.

PR #339, opened 2021-06-04
feat(charts): set default fstype to ext4 in csi-provisioner

shovanmaity:
Set default fstype to ext4 in csi-provisioner. This will be helpful when fsType is not mention in storageclass.


pawanpraka1 (approved):
looks good. Thanks @shovanmaity .

PR #342, opened 2021-06-11
chore(workflow): replace ci tag with regex in release workflow

akhilerm:
- replace ci tag with regex in sed
- use github actions user to commit changes instead of GITHUB_ACTOR

Signed-off-by: Akhil Mohan <akhil.mohan@mayadata.io>


pawanpraka1 (approved):
looks good.

PR #353, opened 2021-06-25
fix(crash): reuse the err variable

pawanpraka1:
fixes: https://github.com/openebs/zfs-localpv/issues/351

**What this PR does?**:

The error variable is being redeclared, so this is not
accessible outside of the loop and the provisioner is crashing
while return in case of error as it tries to access err.Error
and err is nil outside of the loop.

The controller will not crash unless there is grpc timeout or controller is not able to provision the volume on any one of the nodes.

no


avishnu (approved):
Lgtm

PR #366, opened 2021-07-27
refact(e2e): load key's passphrase in dataset after node-power-failure test case

w3aman:
Signed-off-by: w3aman <aman.gupta@mayadata.io>

- This PR refactors the node-power failure infra-chaos test. It adds those manual step where we have to load key's passphrase into dataset on the node after node failure cases.

- For docker-restart test-case if we try to stop docker services on node with the command something like: 
` systemctl stop docker.service`

it can give such warning
```
Warning: Stopping docker.service, but it can still be activated by:
  docker.socket
```
And then docker services will still be running.
So we can stop docker via socket and can execute following command:
` systemctl stop docker.socket`

### DIFF
--- a/e2e-tests/experiments/infra-chaos/node_failure/run_e2e_test.yml
+++ b/e2e-tests/experiments/infra-chaos/node_failure/run_e2e_test.yml
@@ -82,6 +82,9 @@ spec:
           - name: ZPOOL_NAME
             value: ''
 
+          - name: ZPOOL_ENCRYPTION_PASSWORD
+            value: 'test1234'
+
           - name: ESX_PASSWORD
             valueFrom:
               secretKeyRef:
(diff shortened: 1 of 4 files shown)


e2e-tests/experiments/infra-chaos/node_failure/run_e2e_test.yml
```diff
@@ -82,6 +82,9 @@ spec:
           - name: ZPOOL_NAME
             value: ''
 
+          - name: ZPOOL_ENCRYPTION_PASSWORD
+            value: 'test1234'
```
nsathyaseelan, line 86: @w3aman  is this value needed here in runner file?

nsathyaseelan (changes requested):

e2e-tests/experiments/infra-chaos/node_failure/run_e2e_test.yml
w3aman: In case of encrypted zpool, after node-failure we have to again load this key into datasets so we need this value

nsathyaseelan (approved):
lgtm

PR #368, opened 2021-08-03
refact(e2e-crd): update zfs e2e-crds apiversion to v1

w3aman:
Signed-off-by: w3aman <aman.gupta@mayadata.io>

- Update zfs  e2e-crds apiversion to v1


nsathyaseelan (approved):
lgtm

PR #393, opened 2021-10-02
fix (localpv): CSIStorageClass shows zero if SC "poolname" includes dataset path

### LINKED ISSUE #392: GetCapacity reports zero when using non-root dataset

I created a StorageClass with the poolname "rpool/LOCALPV", and volumeBindingMode: WaitForFirstConsumer, then tried to start pods with PVCs using this SC.

CSIStorageCapacity objects containing non-zero "capacity" field should have been created.

**The output of the following commands will help us better understand what's going on**:

The capacities can be checked quickly:

`kubectl get -A CSIStorageCapacity -o custom-columns=NODE:nodeTopology.matchLabels.kubernetes\\.io/hostname,NAME:metadata.name,CAPACITY:capacity`

Logs doesn't contain anything helpful here.

In pkg/driver/controller.go there's the function called "GetCapacity" that reports node storage capacity. This function returns 0 when the "poolname" attribute of the StorageClass contains anything but a ZFS pool name, e.g. "rpool".

However the documentation states that I can use a child dataset too, e.g. "rpool/LOCALPV".

If I create a storage class with the above path and volumeBindingMode: WaitForFirstConsumer, kube-scheduler will report that it cannot schedule the pods due to insufficient storage capacity.

**Environment:**
- ZFS-LocalPV version: 1.9.3
- Kubernetes version (use `kubectl version`): 1.22.2
- Kubernetes installer & version: kubeadm 1.22.2
- Cloud provider or hardware configuration: 3 x Raspberry Pi 4 8G RAM, also a desktop with i10900K & 64G RAM
- OS (e.g. from `/etc/os-release`): Ubuntu "Focal" 20.04.3 in both cases

**Fix**

PR coming soon.

netom:
Signed-off-by: Fábián Tamás László <giganetom@gmail.com>

It fixes #392 "GetCapacity reports zero when using non-root dataset"

**What this PR does?**:

Amends pkg/driver/controller.go to return the root dataset's capacity when using a child dataset as the pool name.

I'm aware that this is not the ideal solution, as the child dataset may offer less capacitry than the whole pool, but this is most definitely better than returning a straigh zero like it's now.

The situation could further be improved if the Node could return the capacity of a child dataset.

No, this is just a simple fix in the capacity calculation.

I've manually verified that the storage capacity returns the pool's capacity when a child dataset is used, and also when the "poolname" is really just the pool's name.

This is just a small, standalone fix, not part of anything greater.

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -877,6 +877,21 @@ func (cs *controller) GetCapacity(
 	params := req.GetParameters()
 
 	poolParam := helpers.GetInsensitiveParameter(&params, "poolname")
+
+	// The "poolname" parameter can either be the name of a ZFS pool
+	// (e.g. "zpool"), or a path to a child dataset (e.g. "zpool/k8s/localpv").
+	//
+	// We parse the "poolname" parameter so the name of the ZFS pool and the
+	// path to the dataset is available separately.
+	//
+	// The dataset path is not used now. It could be used later to query the
+	// capacity of the child dataset, which could be smaller than the capacity
+	// of the whole pool.
+	//
+	// This is necessary because capacity calculation currently only works with
+	// ZFS pool names. This is why it always returns the capacitry of the whole
+	// pool, even if the child dataset given as the "poolname" parameter has a
+	// smaller capacity than the whole pool.
 	poolParamPool, _ := func() (string, string) {
 		poolParamSliced := strings.SplitN(poolParam, "/", 2)
 		if len(poolParamSliced) == 2 {
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -873,8 +873,18 @@ func (cs *controller) GetCapacity(
 	}
 
 	zfsNodesCache := cs.zfsNodeInformer.GetIndexer()
+
 	params := req.GetParameters()
+
 	poolParam := helpers.GetInsensitiveParameter(&params, "poolname")
+	poolParamPool, _ := func() (string, string) {
+		poolParamSliced := strings.SplitN(poolParam, "/", 2)
+		if len(poolParamSliced) == 2 {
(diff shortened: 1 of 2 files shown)


pkg/driver/controller.go
```diff
@@ -873,8 +873,18 @@ func (cs *controller) GetCapacity(
 	}
 
 	zfsNodesCache := cs.zfsNodeInformer.GetIndexer()
+
 	params := req.GetParameters()
+
 	poolParam := helpers.GetInsensitiveParameter(&params, "poolname")
+	poolParamPool, _ := func() (string, string) {
```
pawanpraka1: Can we add a comment explaining why need to parse the pool name.

pkg/driver/controller.go
```diff
@@ -892,7 +902,7 @@ func (cs *controller) GetCapacity(
 		// See https://github.com/kubernetes/enhancements/tree/master/keps/sig-storage/1472-storage-capacity-tracking#available-capacity-vs-maximum-volume-size &
 		// https://github.com/container-storage-interface/spec/issues/432 for more details
 		for _, zpool := range zfsNode.Pools {
-			if zpool.Name != poolParam {
+			if zpool.Name != poolParamPool {
```
pawanpraka1: Can we rename poolParamPool to poolname?

pawanpraka1 (commented):
Looks good. Have given some minor comments. Thanks for this PR @netom.

pkg/driver/controller.go
netom: Sure, no problem. I'll update the PR quick.

pkg/driver/controller.go
netom: Of course. Will do that shortly.

pawanpraka1 (approved):
looks good.

PR #395, opened 2021-10-08
fix(provisioning): register topologyKeys from driver env

shubham14bajpai:
Signed-off-by: shubham <shubham14bajpai@gmail.com>

Right now, all custom node Labels will be registered when restart the ZFS-LocalPV Driver, CSI node topologyKeys will keep all those keys even if node labels has been changed. When custom node Labels has been deleted, ZFS LocalPV CSI driver will not schedule the PV to the node, restart the ZFS-LocalPV Driver daemon set is a must requirement.

This PR fixes the above issue by adding a `ALLOWED_TOPOLOGIES` to the zfs-node daemonset which will take only a given set of custom keys required for provisioning.
```yaml
env:
  - name: OPENEBS_NODE_ID
    valueFrom:
      fieldRef:
        fieldPath: spec.nodeName
  - name: OPENEBS_CSI_ENDPOINT
    value: unix:///plugin/csi.sock
  - name: OPENEBS_NODE_DRIVER
    value: agent
  - name: OPENEBS_NAMESPACE
    value: openebs
  - name: ALLOWED_TOPOLOGIES
    value: "openebs.io/rack"
```

**What this PR does?**:

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests


pkg/driver/agent.go
```diff
@@ -245,6 +244,18 @@ func (ns *node) NodeGetInfo(
...
 
+	// support topologykeys from env ALLOWED_TOPOLOGIES
+	allowedTopologies := os.Getenv("ALLOWED_TOPOLOGIES")
+	allowedKeys := strings.Split(allowedTopologies, ",")
+
+	for _, key := range allowedKeys {
+		if key != "" {
+			if value, ok := node.Labels[key]; ok {
+				topology[key] = value
+			}
```
mittachaitu: Let's add a warning if the configured label is not available on nodes... but as per docs of [topology](https://kubernetes-csi.github.io/docs/topology.html#implementing-topology-in-your-csi-driver) it should be another way around like if a driver passes topology information if that key:value is not present then kubelet will add it on a node. 

image

shubham14bajpai: Done

mittachaitu (approved):
LGTM

docs/storageclasses.md
```diff
@@ -189,6 +189,57 @@ allowedTopologies:
...
+  - apiVersion: v1
+    kind: Node
+    name: k8s-node-1
+    uid: fe268f4b-d9a9-490a-a999-8cde20c4dadb
+  resourceVersion: "4586341"
+  selfLink: /apis/storage.k8s.io/v1/csinodes/k8s-node-1
+  uid: 522c2110-9d75-4bca-9879-098eb8b44e5d
+spec:
+  drivers:
+  - name: local.csi.openebs.io
```
pawanpraka1: zfs.csi.openebs.io

pawanpraka1 (commented):
looks good. Have One minor comment.

docs/storageclasses.md
shubham14bajpai: Done :+1:

pawanpraka1 (approved):
looks good.

pawanpraka1:
this will be a breaking change if someone was already utilizing the old behavior. Thinking out loud, should we have default list as "All", which will add all the node labels as topology key and if someone wants to change that, he can update the list to the desired keys. What do you think @shubham14bajpai @kmova .

shubham14bajpai:
> this will be a breaking change if someone was already utilizing the old behavior. Thinking out loud, should we have default list as "All", which will add all the node labels as topology key and if someone wants to change that, he can update the list to the desired keys. What do you think @shubham14bajpai @kmova .

Yes in that case we can set the env as empty string/"All" by default and add all the labels.

pawanpraka1:
@shubham14bajpai can we go ahead and add "All" as the default value for the ALLOWED_TOPOLOGIES.

mittachaitu:
> > this will be a breaking change if someone was already utilizing the old behavior. Thinking out loud, should we have default list as "All", which will add all the node labels as topology key and if someone wants to change that, he can update the list to the desired keys. What do you think @shubham14bajpai @kmova .
> 
> Yes in that case we can set the env as empty string/"All" by default and add all the labels.

IMO, let's not set the env by default from OpenEBS side mainly to support backward compatibility, if env is empty/not set we can push all node labels same as current behavior. If it is configured by user we can honor it as is.

shubham14bajpai:
> > > this will be a breaking change if someone was already utilizing the old behavior. Thinking out loud, should we have default list as "All", which will add all the node labels as topology key and if someone wants to change that, he can update the list to the desired keys. What do you think @shubham14bajpai @kmova .
> > 
> > 
> > Yes in that case we can set the env as empty string/"All" by default and add all the labels.
> 
> IMO, let's not set the env by default from OpenEBS side mainly to support backward compatibility, if env is empty/not set we can push all node labels same as current behavior. If it is configured by user we can honor it as is.

I guess having the env present with "All" will help the user get to know about the new feature even if they miss the release notes and will still not break backward compatibility.

deploy/yamls/zfs-driver.yaml
```diff
@@ -1053,6 +1053,8 @@ spec:
               value: agent
             - name: OPENEBS_NAMESPACE
               value: openebs
+            - name: ALLOWED_TOPOLOGIES
+              value: "All"
```
pawanpraka1, line 1061: can we add a comment(with example) saying, we can add comma separated keys here.

docs/faq.md
```diff
@@ -91,7 +91,12 @@ spec:
 
 ### 6. How to add custom topology key
 
-To add custom topology key, we can label all the nodes with the required key and value :-
+To add custom topology key:
+* Label the nodes with the required key and value.
+* Set env variables in the ZFS driver daemonset yaml(openebs-zfs-node), if already deployed, you can edit the daemonSet directly.
```
pawanpraka1: let us add the details about the default value where we need to start the daemonset to pick the new label.

pawanpraka1 (approved):
looks good.

PR #406, opened 2022-03-22
fix(destroy): making ZFS-LocalPV aware of pool import

### LINKED ISSUE #398: filesystems not removed after zpool was offline
A machine of ours that is running the zfs-localpv driver recently had an issue with the zpool that took some time to investigate. In the meantime, no productions pods were scheduled on this machine, but it was online. An operator running in the cluster regularly creates and removes snapshots and kept doing so during the outage.

Creating new snapshots obviously failed and the driver kept trying to ensure this for as long as the outage took. The cleaning of the snapshots however succeeded in the perspective of the driver:

```
I1108 15:08:55.087960       1 zfs_util.go:683] deleted snapshot pvc-24bf1c9a-5858-4e81-be77-822d42bcbc55@snapshot-16b1d680-57b1-4b70-8e96-d69c755e14b7
E1108 15:08:55.103201       1 zfs_util.go:667] destroy: snapshot pvc-24bf1c9a-5858-4e81-be77-822d42bcbc55 is not present, error: exit status 1
```
While this looks like and is logged like an error is in fact the expected result: the driver must ensure the snapshot is no longer there and it isn't. Job done!

In our case, as soon as the zpool came online again, there was a huge difference between the volumesnapshots in kubernetes and the ZFS snapshots on disk. The desirable situation would be that the driver would detect not only the ZFS snapshot was not there but also the rest of the zpool and therefor considered the snapshot not deleted (yet).

While fetching the version our ZFS-LocalPV deployment I saw there has been a lot of releases in between, so I'll investigate upgrading my setup anyway.

**Environment:**
- ZFS-LocalPV version: 1.3.0
- Kubernetes version (use `kubectl version`): 1.18.9
- Kubernetes installer & version: manual
- Cloud provider or hardware configuration: bare-metal
- OS (e.g. from `/etc/os-release`): Ubuntu 20.04

pawanpraka1:
fixes: https://github.com/openebs/zfs-localpv/issues/398

If we are destroying a volume/snapshot and pool is not
imported, the ZFS-LocalPV driver assumes that the volume/snapshot
is not there and deletes the corresponding zfsvolume and zfssnap CR.

Here we are checking the parent dataset presence before attempting
to delete the volume or snapshot.

### DIFF
--- a/pkg/zfs/zfs_util.go
+++ b/pkg/zfs/zfs_util.go
@@ -621,6 +621,15 @@ func SetVolumeProp(vol *apis.ZFSVolume) error {
 // DestroyVolume deletes the zfs volume
 func DestroyVolume(vol *apis.ZFSVolume) error {
 	volume := vol.Spec.PoolName + "/" + vol.Name
+	parentDataset := vol.Spec.PoolName
+
+	// check if parent dataset is present or not before attempting to delete the volume
+	if err := getVolume(parentDataset); err != nil {
+		klog.Errorf(
+			"destroy: parent dataset %v is not present, error: %s", parentDataset, err.Error(),
+		)
+		return err
+	}
 
 	if err := getVolume(volume); err != nil {
 		klog.Errorf(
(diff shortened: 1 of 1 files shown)


akhilerm (approved):
LGTM

PR #478, opened 2023-09-20
fix(api-version): update apiversion to v1 for snapshot and snapshotclass

w3aman:
- update apiversion to v1 for volume snapshot and volume snapshot class in e2e-tests


sinhaashish (approved):

hrudaya21 (approved):

avishnu (approved):
lgtm

PR #512, opened 2024-03-16
refactor(charts): move CRDs into a subchart 

niladrih:
Changes:
- bumps chart version, appVersion, image tag.
- moves CRDs into a subchart and adds templating to enable/disable them, while maintaining install order. Install order matters for CRDs in case there are CRs inside the chart which may require the CRDs.
- Add templating to CRDs which disables their installation if they already exist. Because the CRDs are inside the template directory of the subchart, they will carry the release-name label from helm. This makes them difficult to patch into an existing CRD in the cluster. The install/upgrade errors out in these cases. The templating will make sure the CRDs will be installed only if they are not present.

PRs with similar change:
- https://github.com/openebs/mayastor-extensions/pull/414
- https://github.com/openebs/mayastor-extensions/pull/438
- https://github.com/openebs/lvm-localpv/pull/284


Abhinandan-Purkait (approved):

Abhinandan-Purkait:
Has been picked in #513

PR #513, opened 2024-03-20
chore: change zfs-controller to a deployment from statefulset

Abhinandan-Purkait:
### Changes in the PR

This changes the app type of zfs-controller from a statefulset to a deployment.


niladrih:
@Abhinandan-Purkait -- while you're making this change, could you also check in these ones as well:
- Codecov CI fix
- Go mod security fix by bumping protobuf (google) to 1.33.0

Thanks.

niladrih (approved):

PR #519, opened 2024-03-27
fix: crds getting removed on helm upgrades

Abhinandan-Purkait:
#### Why do we need this PR?

It was seen that on upgrade to 1.5.x the CRDs get removed causing the removal of CRs. This would prevent it.
It also adds annotations to let helm keep the CRDs on uninstall, which is configurable.

### DIFF
--- a/deploy/helm/charts/Chart.yaml
+++ b/deploy/helm/charts/Chart.yaml
@@ -17,3 +17,4 @@ sources:
 dependencies:
   - name: crds
     version: "2.5.1"
+    condition: crds.enabled
--- a/deploy/helm/charts/Chart.yaml
+++ b/deploy/helm/charts/Chart.yaml
@@ -16,5 +16,5 @@ sources:
 - https://github.com/openebs/zfs-localpv
 dependencies:
   - name: crds
-    version: "2.5.1"
+    version: 2.5.1
     condition: crds.enabled
--- a/deploy/helm/charts/Chart.yaml
+++ b/deploy/helm/charts/Chart.yaml
@@ -1,7 +1,7 @@
 apiVersion: v2
 name: zfs-localpv
 description: Helm chart for CSI Driver for dynamic provisioning of ZFS Persistent Local Volumes. For instructions on how to use this helm chart, see - https://openebs.github.io/zfs-localpv/
-version: 2.5.0
+version: 2.5.1
 appVersion: 2.5.0
 icon: https://raw.githubusercontent.com/cncf/artwork/master/projects/openebs/icon/color/openebs-icon-color.png
 home: https://openebs.io/
@@ -16,4 +16,4 @@ sources:
 - https://github.com/openebs/zfs-localpv
 dependencies:
   - name: crds
-    version: "2.5.0"
+    version: "2.5.1"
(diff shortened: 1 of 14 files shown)


deploy/helm/charts/charts/crds/templates/csi-volume-snapshot-content.yaml
```diff
@@ -1,12 +1,11 @@
 {{- if .Values.csi.volumeSnapshots.enabled -}}
-{{- $crdName := "volumesnapshotcontents.snapshot.storage.k8s.io" -}}
-{{- if (include "crdIsAbsent" (list $crdName)) -}}
 apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
     api-approved.kubernetes.io: https://github.com/kubernetes-csi/external-snapshotter/pull/814
     controller-gen.kubebuilder.io/version: v0.11.3
+    {{- include "crds.extraAnnotations" .Values.csi.volumeSnapshots | nindent 4 }}
```
avishnu, line 8: should it be volumeSnapshotContents?

avishnu: I guess it's correct.

avishnu (approved):
lgtm

tiagolobocastro (approved):

niladrih (approved):

deploy/helm/charts/Chart.yaml
```diff
@@ -16,4 +16,5 @@ sources:
 - https://github.com/openebs/zfs-localpv
 dependencies:
   - name: crds
-    version: "2.5.0"
+    version: "2.5.1"
```
niladrih: nit: this shouldn't have the double quotes because we're typically not using quotes in chart versions.

deploy/helm/charts/Chart.yaml
```diff
@@ -16,4 +16,5 @@ sources:
 - https://github.com/openebs/zfs-localpv
 dependencies:
   - name: crds
-    version: "2.5.0"
+    version: 2.5.1
+    condition: crds.enabled
```
tiagolobocastro, line 20: do we need to document that this has to be disabled when upgrading?

tiagolobocastro (approved):

deploy/helm/charts/Chart.yaml
avishnu: I think it's already present in the doc website. https://openebs.io/docs/main/user-guides/upgrade

niladrih (approved):

Abhinandan-Purkait:
> and this now leads to it bugging out when reinstalling in a cluster or installing on a cluster that previously used the crds folder.
> 
> Honestly: Does anyone here actually test these helm PR's, outside just a single initial install? But even so, this behavior is already pretty clear by just reading the damn template... So what is going wrong here?

For reinstalling or a fresh install with crds existing, users can use a simple set flag `--set crds.csi.volumeSnapshots.enabled=false` to them helm command.

I don't think that was too hard to figure out from the PR, anyway. 

So I don't anything is going wrong here, we provided options, now users need to use what fits their use case.

If you still feel we don't know what we are doing please feel free to open an issue and you can contribute, we would be happy to review your PR.

wolfgangwalther:
> and this now leads to it bugging out when reinstalling in a cluster or installing on a cluster that previously used the crds folder.

"bugging out" as in "fails to install the chart, but doesn't destroy anything"?

Cool, I can very happily live with that - this is what I expected when I need to change some configuration.

Issue #423, opened 2022-08-17
Option to choose `refquota` zfs field to apply quotas


tahajahangir:
**Describe the problem/challenge you have**
Currently, PVC sizes are applied using zfs `quota` field. This way, taking snapshot of a volume, decreases the available space of the PVC. I think this is not the desired behavior for many users.

**Describe the solution you'd like**
A simple solution is using `refquota` instead of `quota`.
A better solution is adding an option (operator-wide, per StorageClass or per PVC) to apply limits with `refquota`.
Another solution is provide a way (e.g. with annotations) to apply `quota` and `refquota` limits independently. For example, we can use the `capacity` property of PVC for `refquota`, and provide an annotation for limiting the zfs `quota`.

*Environment:**
- Not important

rdegez:
We have the exact same issue and I wonder how many people are actualy using snapshots in production without this!

Abhinandan-Purkait:
@sabzsystem I see a PR was raised to address this? Why was this closed?

We can take this up as a roadmap item. Thanks

niladrih:
Continuing the discussion here @cinapm.

I have one issue with #542: I think the CSI driver shouldn't be pinned to the quota/refquota choice it made at deploy time. It should be configurable via a storageClass parameter. Admins could choose if the provisioner uses quota/refquota at provision based on the StorageClass configuration.

I'm not a ZFS pro, but does the quota affect volume expansion?

@tiagolobocastro @avishnu @dsavitskiy you thoughts?

Abhinandan-Purkait:
The functionality has been added as part of PR #542 . Will try to add a test for the change before we close this issue.

sinhaashish:
Closing this issue as it has beeen resolved using these PRs https://github.com/openebs/zfs-localpv/pull/542  and https://github.com/openebs/zfs-localpv/pull/608

PR #542, opened 2024-06-13
feat(zfs-localpv): add option for choosing between refquota and quota

cinapm:
resolve this issue: [https://github.com/openebs/zfs-localpv/issues/423](https://github.com/openebs/zfs-localpv/issues/423)

**What this PR does?**: 
this PR adds an option for choosing between refquota or quota and uses this value as zfs command options

**Checklist:**
- [X] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -221,7 +221,7 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 	schld := parameters["scheduler"]
 	fstype := parameters["fstype"]
 	shared := parameters["shared"]
-	qoutatype := parameters["qoutatype"]
+	quotatype := parameters["quotatype"]
 
 	vtype := zfs.GetVolumeType(fstype)
 
@@ -280,7 +280,7 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 		WithVolumeType(vtype).
 		WithVolumeStatus(zfs.ZFSStatusPending).
 		WithFsType(fstype).
-		WithQuotaType(qoutatype).
+		WithQuotaType(quotatype).
 		WithShared(shared).
 		WithCompression(compression).Build()
 
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -221,6 +221,7 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 	schld := parameters["scheduler"]
 	fstype := parameters["fstype"]
 	shared := parameters["shared"]
+	qoutatype := parameters["qoutatype"]
 
 	vtype := zfs.GetVolumeType(fstype)
 
@@ -279,6 +280,7 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 		WithVolumeType(vtype).
 		WithVolumeStatus(zfs.ZFSStatusPending).
 		WithFsType(fstype).
+		WithQuotaType(qoutatype).
 		WithShared(shared).
 		WithCompression(compression).Build()
 
--- a/pkg/zfs/zfs_util.go
+++ b/pkg/zfs/zfs_util.go
@@ -153,10 +153,10 @@ func buildCloneCreateArgs(vol *apis.ZFSVolume) []string {
 		if vol.Spec.ThinProvision == "no" {
 			reservationProperty := ""
 			switch vol.Spec.QuotaType {
-			case "quota":
-				reservationProperty = "reservation=" + vol.Spec.Capacity
 			case "refquota":
 				reservationProperty = "refreservation=" + vol.Spec.Capacity
+			default:
+				reservationProperty = "reservation=" + vol.Spec.Capacity
 			}
 			ZFSVolArg = append(ZFSVolArg, "-o", reservationProperty)
 		}
@@ -232,10 +232,10 @@ func buildDatasetCreateArgs(vol *apis.ZFSVolume) []string {
 	if vol.Spec.ThinProvision == "no" {
 		reservationProperty := ""
 		switch vol.Spec.QuotaType {
-		case "quota":
-			reservationProperty = "reservation=" + vol.Spec.Capacity
 		case "refquota":
 			reservationProperty = "refreservation=" + vol.Spec.Capacity
+		default:
+			reservationProperty = "reservation=" + vol.Spec.Capacity
 		}
 		ZFSVolArg = append(ZFSVolArg, "-o", reservationProperty)
 	}
--- a/pkg/zfs/zfs_util.go
+++ b/pkg/zfs/zfs_util.go
@@ -143,7 +143,7 @@ func buildCloneCreateArgs(vol *apis.ZFSVolume) []string {
 
 	if vol.Spec.VolumeType == VolTypeDataset {
 		if len(vol.Spec.Capacity) != 0 {
-			quotaProperty := "quota=" + vol.Spec.Capacity
+			quotaProperty := vol.Spec.QuotaType + "=" + vol.Spec.Capacity
 			ZFSVolArg = append(ZFSVolArg, "-o", quotaProperty)
 		}
 		if len(vol.Spec.RecordSize) != 0 {
@@ -216,7 +216,7 @@ func buildDatasetCreateArgs(vol *apis.ZFSVolume) []string {
 	ZFSVolArg = append(ZFSVolArg, ZFSCreateArg)
 
 	if len(vol.Spec.Capacity) != 0 {
-		quotaProperty := "quota=" + vol.Spec.Capacity
+		quotaProperty := vol.Spec.QuotaType + "=" + vol.Spec.Capacity
 		ZFSVolArg = append(ZFSVolArg, "-o", quotaProperty)
 	}
 	if len(vol.Spec.RecordSize) != 0 {
@@ -292,7 +292,7 @@ func buildVolumeResizeArgs(vol *apis.ZFSVolume) []string {
 	ZFSVolArg = append(ZFSVolArg, ZFSSetArg)
 
 	if vol.Spec.VolumeType == VolTypeDataset {
-		quotaProperty := "quota=" + vol.Spec.Capacity
+		quotaProperty := vol.Spec.QuotaType + "=" + vol.Spec.Capacity
 		ZFSVolArg = append(ZFSVolArg, quotaProperty)
 	} else {
 		volsizeProperty := "volsize=" + vol.Spec.Capacity
@@ -350,7 +350,7 @@ func buildVolumeRestoreArgs(rstr *apis.ZFSRestore) ([]string, error) {
 
 	if rstr.VolSpec.VolumeType == VolTypeDataset {
 		if len(rstr.VolSpec.Capacity) != 0 {
-			ZFSRecvParam += " -o quota=" + rstr.VolSpec.Capacity
+			ZFSRecvParam += " -o " + rstr.VolSpec.QuotaType + "=" + rstr.VolSpec.Capacity
 		}
 		if len(rstr.VolSpec.RecordSize) != 0 {
 			ZFSRecvParam += " -o recordsize=" + rstr.VolSpec.RecordSize
--- a/pkg/zfs/zfs_util.go
+++ b/pkg/zfs/zfs_util.go
@@ -151,14 +151,7 @@ func buildCloneCreateArgs(vol *apis.ZFSVolume) []string {
 			ZFSVolArg = append(ZFSVolArg, "-o", recordsizeProperty)
 		}
 		if vol.Spec.ThinProvision == "no" {
-			reservationProperty := ""
-			switch vol.Spec.QuotaType {
-			case "refquota":
-				reservationProperty = "refreservation=" + vol.Spec.Capacity
-			default:
-				reservationProperty = "reservation=" + vol.Spec.Capacity
-			}
-			ZFSVolArg = append(ZFSVolArg, "-o", reservationProperty)
+			ZFSVolArg = append(ZFSVolArg, "-o", reservationProperty(vol.Spec.QuotaType, vol.Spec.Capacity))
 		}
 		ZFSVolArg = append(ZFSVolArg, "-o", "mountpoint=legacy")
 	}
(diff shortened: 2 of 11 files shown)


cmd/main.go
```diff
@@ -74,18 +74,26 @@ func main() {
...
-func run(config *config.Config) {
+func run(config *configs.Config) {
 	if config.Version == "" {
 		config.Version = version.Current()
 	}
 
+	if configs.QuotaType != configs.Quota && configs.QuotaType != configs.RefQuota {
+		log.Fatalln(fmt.Errorf("quota-type should be quota or refquota"))
+	}
+
```
Abhinandan-Purkait: I think instead of erroring out if that option is not passed it would be better if we set it to `configs.Quota` otherwise this needs to be passed down in upgrade, also kind of changes the default behaviour. cc @tiagolobocastro @niladrih Any thoughts?

pkg/config/config.go
```diff
@@ -43,6 +43,13 @@ type Config struct {
 	Nodename string
 }
 
+const (
+	Quota    = "quota"
+	RefQuota = "refquota"
+)
```
Abhinandan-Purkait: Please add comments explaining them.

cmd/main.go
```diff
@@ -19,10 +19,10 @@ package main
 import (
 	"flag"
 	"fmt"
+	configs "github.com/openebs/zfs-localpv/pkg/config"
```
Abhinandan-Purkait: nit: Any reason for changing it to configs? The package name is config.
Also please run `goimports` to sort the imports.

cmd/main.go
cinapm: Additionally, the input option should be either quota or refquota and not any other values

pkg/config/config.go
cinapm: i have added comments for quota types @Abhinandan-Purkait

cmd/main.go
cinapm: I have run imports and now it is fixed @Abhinandan-Purkait

cinapm: I have changed to configs because I want to use the QuotaType variable in the config package globally and it has a conflict with config variable name @Abhinandan-Purkait

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
Thanks @cinapm. Looks good to me. I will let other folks review this. We will target this for this upcoming release.

dsharma-dc:
Have we tested this change by trying to apply a quota, and then a refquota underneath that dataset?
Add a simple test as well perhaps.

niladrih (commented):
This configuration option should be toggle-able through a storageclass.

Abhinandan-Purkait:
@cinapm Can you please let us know your plans on this PR? I would request you to consider the suggestion made by @niladrih here.

cinapm:
> This configuration option should be toggle-able through a storageclass.

i have added option to add quotatype parameter on storageclass, the default value is quota

cinapm:
> @cinapm Can you please let us know your plans on this PR? I would request you to consider the suggestion made by @niladrih here.

instead of using config for local-pv, I have changed the controller to accept and use quotatype parameter on storageclass

Abhinandan-Purkait:
@cinapm Thanks, we will review this. The team needs to decide on this because this changes the CRDs and changing anything on the CRDs would need a CRD version bump which now raises the question of upgrades and migration. cc @niladrih @avishnu @tiagolobocastro

abuisine:
I would strongly suggest to verify the impact on the **thick** provisioning side of things :
with a `thinprovision: "no"` the property `reservation` is set, it would probably be mandatory to switch to `refreservation` at the same time that `refquota` is used.

Let me explain the potential problem with the typical usage of snapshots :
- snapshots are considered in `quota` but are not when using `refquota`
- snapshots are considered in `reservation` but are not when using `refreservation`

You could end up with a non functionnal **thick** provisioning while mixing `refquota` with `reservation`

Let me take an exemple with a workload that requires 10GB of free space, with **thick** provisioning :
- a PersitentVolume of 10GB
- `refquota` and `reservation` being set (through `thinprovision: "no"`)
- with `refquota`, you can create an unlimited amount of GB of snapshots
- with `reservation`, snapshots are considered as consuming space
- this would mean that with 3GB of space being used and 9GB of snapshots, there would be 7GB of free space from a zfs FS standpoint, and that `reservation` would be exceeded by 2GB
- so 2GB would be **thin** provisioned whereas the configuration is set to **thick** provisioning

Using `refreservation` while setting `refquota` and `thinprovision: "no"` would solve the issue.

abuisine:
Also, I would like to confirm the huge value of this PR : without `refquota` and `refreservation` the usage of snapshots becomes almost impossible as they are deduced from free space.
From an operational standpoint, it basically means that backups can impact production.
We never used snapshots with this CSI for this reason.

This PR (with `refreservation`) would enable safe usage of snapshots in production.

cinapm:
> I would strongly suggest to verify the impact on the **thick** provisioning side of things : with a `thinprovision: "no"` the property `reservation` is set, it would probably be mandatory to switch to `refreservation` at the same time that `refquota` is used.
> 
> Let me explain the potential problem with the typical usage of snapshots :
> 
> * snapshots are considered in `quota` but are not when using `refquota`
> * snapshots are considered in `reservation` but are not when using `refreservation`
> 
> You could end up with a non functionnal **thick** provisioning while mixing `refquota` with `reservation`
> 
> Let me take an exemple with a workload that requires 10GB of free space, with **thick** provisioning :
> 
> * a PersitentVolume of 10GB
> * `refquota` and `reservation` being set (through `thinprovision: "no"`)
> * with `refquota`, you can create an unlimited amount of GB of snapshots
> * with `reservation`, snapshots are considered as consuming space
> * this would mean that with 3GB of space being used and 9GB of snapshots, there would be 7GB of free space from a zfs FS standpoint, and that `reservation` would be exceeded by 2GB
> * so 2GB would be **thin** provisioned whereas the configuration is set to **thick** provisioning
> 
> Using `refreservation` while setting `refquota` and `thinprovision: "no"` would solve the issue.

Your point is very valid, and based on the points you mentioned, I have made the changes. now when there is option `thinprovision: "no"` , the `reservation` will be set for `quota` and `refreservation` will be set for `refquota`

pkg/driver/controller.go
```diff
@@ -237,6 +237,7 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 	schld := parameters["scheduler"]
 	fstype := parameters["fstype"]
 	shared := parameters["shared"]
+	qoutatype := parameters["qoutatype"]
```
sinhaashish: qoutatype -> quotatype

pkg/driver/controller.go
```diff
@@ -295,6 +296,7 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 		WithVolumeType(vtype).
 		WithVolumeStatus(zfs.ZFSStatusPending).
 		WithFsType(fstype).
+		WithQuotaType(qoutatype).
```
sinhaashish: qoutatype -> quotatype

pkg/zfs/zfs_util.go
```diff
@@ -143,15 +143,21 @@ func buildCloneCreateArgs(vol *apis.ZFSVolume) []string {
...
 			ZFSVolArg = append(ZFSVolArg, "-o", quotaProperty)
 		}
 		if len(vol.Spec.RecordSize) != 0 {
 			recordsizeProperty := "recordsize=" + vol.Spec.RecordSize
 			ZFSVolArg = append(ZFSVolArg, "-o", recordsizeProperty)
 		}
 		if vol.Spec.ThinProvision == "no" {
-			reservationProperty := "reservation=" + vol.Spec.Capacity
+			reservationProperty := ""
+			switch vol.Spec.QuotaType {
```
sinhaashish: If `quota` is teh default then it should be handled in default,.

pkg/driver/controller.go
cinapm: i have fixed this typo

pkg/driver/controller.go
cinapm: i have fixed this typo

pkg/zfs/zfs_util.go
cinapm: can you explain me more? i can't get what do you mean

cinapm: @sinhaashish

Abhinandan-Purkait (changes requested):
@cinapm A couple of issues here I see you have changed the `v1/zfsvolume.go` file and also the crd and operator yamls. The CRDs are supposed to be generated from the `v1/zfsvolume.go` file and are not supposed to be changed manually. It would be good if you undo the yaml changes and do it via the script. Also, since you have changed the `v1/zfsvolume.go` we also need to do the kubegen again.

```
make bootstrap
make kubegen
make manifests
```
One more thing, I see you changed the zfs-restore and zfs-snapshot crds as well, but I don't see any equivalent changes on their corresponding go files. So if you want to change them as well, you need to change the go files first and run the above scripts.

cinapm:
> make manifests
@Abhinandan-Purkait 

i didn't change them manually, first of all I changed `v1/zfsvolume.go` and after that I have used following commands:
`make bootstrap
make manifests`

the `make manifests` command has changed zfs-restore and zfs-snapshot crds automatically based on changes in `v1/zfsvolume.go`

Abhinandan-Purkait:
> > make manifests
> > @Abhinandan-Purkait
> 
> i didn't change them manually, first of all I changed `v1/zfsvolume.go` and after that I have used following commands: `make bootstrap make manifests`
> 
> the `make manifests` command has changed zfs-restore and zfs-snapshot crds automatically based on changes in `v1/zfsvolume.go`

Okay, I see the embedded volSpec in the restores and snapshot structures. Okay. It would help if you rebased your branch, there were some changes merged on the deepcopy methods. Also please run `make kubegen` although it won't change anything in your case, but it's worth checking.

You also need to copy the crd changes into the `helm chart` crd templates, as that's how crds get installed. Check https://github.com/openebs/zfs-localpv/tree/develop/deploy/helm/charts/charts/crds/templates

The other changes look good to me. It might be worth testing upgrade once before we take this in. Example, installing the previous version of zfs and upgrading it to this version and see the crd changes. Thanks

pkg/zfs/zfs_util.go
```diff
@@ -216,15 +222,21 @@ func buildDatasetCreateArgs(vol *apis.ZFSVolume) []string {
...
 	}
 	if vol.Spec.ThinProvision == "no" {
-		reservationProperty := "reservation=" + vol.Spec.Capacity
+		reservationProperty := ""
+		switch vol.Spec.QuotaType {
+		case "quota":
+			reservationProperty = "reservation=" + vol.Spec.Capacity
+		case "refquota":
+			reservationProperty = "refreservation=" + vol.Spec.Capacity
+		}
```
Abhinandan-Purkait: How about simplifying this by using a method and calling it here. That way we only need to modify the method to change something and not at all places?

```
// get the reservation property based on the quota type
func reservationProperty(quotaType string, capacity string) string {
	var reservationProperties = map[string]string{
		"quota":    "reservation=",
		"refquota": "refreservation=",
	}
	return reservationProperties[quotaType] + capacity
}
```

```
ZFSVolArg = append(ZFSVolArg, "-o", reservationProperty(vol.Spec.QuotaType, vol.Spec.Capacity))
```

pkg/zfs/zfs_util.go
sinhaashish: Since quota is default why to handle it as a different case. Keep it as default . 
Something like this
```
switch vol.Spec.QuotaType {
  case "refquota":
      reservationProperty = "refreservation=" + vol.Spec.Capacity
  default:
      reservationProperty = "reservation=" + vol.Spec.Capacity
}
```

cinapm: @sinhaashish I have resolved it just as you said

cinapm:
> > > make manifests
> > > @Abhinandan-Purkait
> > 
> > 
> > i didn't change them manually, first of all I changed `v1/zfsvolume.go` and after that I have used following commands: `make bootstrap make manifests`
> > the `make manifests` command has changed zfs-restore and zfs-snapshot crds automatically based on changes in `v1/zfsvolume.go`
> 
> Okay, I see the embedded volSpec in the restores and snapshot structures. Okay. It would help if you rebased your branch, there were some changes merged on the deepcopy methods. Also please run `make kubegen` although it won't change anything in your case, but it's worth checking.
> 
> You also need to copy the crd changes into the `helm chart` crd templates, as that's how crds get installed. Check https://github.com/openebs/zfs-localpv/tree/develop/deploy/helm/charts/charts/crds/templates
> 
> The other changes look good to me. It might be worth testing upgrade once before we take this in. Example, installing the previous version of zfs and upgrading it to this version and see the crd changes. Thanks

@Abhinandan-Purkait 

as you said, first of all, I have rebased my branch and after that I added crds changes into the corresponding helm manifests

pkg/zfs/zfs_util.go
cinapm: @Abhinandan-Purkait 
I have changed this just as you said

Abhinandan-Purkait (approved):
LGTM. Test the upgrades before this can be taken in.

sinhaashish (approved):

sinhaashish:
@cinapm  It will be good if you add the test case to check the `refquota` and `refreservation`
Something on this lines is alredy present here https://github.com/openebs/zfs-localpv/blob/2f938970d5bb03322e0fbdc779453c639ef6545c/tests/utils.go#L131

You can find the details about this is this issue #560 
The conclusion is here https://github.com/openebs/zfs-localpv/issues/560#issuecomment-2232535073

Abhinandan-Purkait:
@cinapm Please rebase and run `make manifests` a lot of changes went in on the testing and ci side of things. Thanks

Abhinandan-Purkait:
@cinapm I see there isn't any activity here for a while, is there any blocker that you are facing that you need help with?

cinapm:
> @cinapm I see there isn't any activity here for a while, is there any blocker that you are facing that you need help with?

I have rebased the develop branch and also run `make manifests`

cinapm:
> @cinapm Please rebase and run `make manifests` a lot of changes went in on the testing and ci side of things. Thanks

@Abhinandan-Purkait 
I have resolved them

Abhinandan-Purkait (approved):
LGTM

Abhinandan-Purkait:
@niladrih @tiagolobocastro @sinhaashish  PTAL

sinhaashish:
I am approving it , but better add the test here [here](https://github.com/openebs/zfs-localpv/blob/8d4686899d817582493a75a928ffe432eddcff49/tests/utils.go#L131)

sinhaashish (approved):

sinhaashish:
@cinapm Just making sure there’s nothing pending on your side. If everything looks good, we can go ahead and merge it

cinapm:
> @cinapm Just making sure there’s nothing pending on your side. If everything looks good, we can go ahead and merge it

@sinhaashish  
there is nothing on my side, everything is ok

Abhinandan-Purkait:
@cinapm Thanks. Merging the PR.

PR #548, opened 2024-06-24
feat(analytics): add heartbeat pinger

niladrih:
Ref:
- https://github.com/openebs/google-analytics-4/pull/8
- https://github.com/openebs/dynamic-localpv-provisioner/pull/188
- https://github.com/openebs/lvm-localpv/pull/318

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -1,19 +1,3 @@
-/*
-Copyright © 2019 The OpenEBS Authors
-
-Licensed under the Apache License, Version 2.0 (the "License");
-you may not use this file except in compliance with the License.
-You may obtain a copy of the License at
-
-    http://www.apache.org/licenses/LICENSE-2.0
-
-Unless required by applicable law or agreed to in writing, software
-distributed under the License is distributed on an "AS IS" BASIS,
-WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
-See the License for the specific language governing permissions and
-limitations under the License.
-*/
-
 package driver
 
 import (
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -22,33 +22,32 @@ import (
 	"strings"
 	"time"
 
-	k8sapi "github.com/openebs/lib-csi/pkg/client/k8s"
-	zfsapi "github.com/openebs/zfs-localpv/pkg/apis/openebs.io/zfs/v1"
-	clientset "github.com/openebs/zfs-localpv/pkg/generated/clientset/internalclientset"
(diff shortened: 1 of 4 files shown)


pkg/driver/controller.go
```diff
@@ -159,8 +160,7 @@ func sendEventOrIgnore(pvcName, pvName, capacity, method string) {
 		analytics.New().CommonBuild(DefaultCASType).ApplicationBuilder().
 			SetVolumeName(pvName).
 			SetVolumeClaimName(pvcName).
-			SetLabel(analytics.EventLabelCapacity).
```
avishnu, line 162: This bit is removed... why?

pkg/driver/controller.go
```diff
@@ -133,7 +133,8 @@ func (cs *controller) init() error {
 	if zfs.GoogleAnalyticsEnabled == "true" {
 		analytics.RegisterVersionGetter(version.GetVersionDetails)
 		analytics.New().CommonBuild(DefaultCASType).InstallBuilder(true).Send()
-		go analytics.PingCheck(DefaultCASType, Ping)
+		go analytics.PingCheck(DefaultCASType, Ping, false)
```
avishnu, line 120: What are true and false for?

pkg/driver/controller.go
niladrih: The v0.1.0 version of the google-analytics-4 module used fields Category, Action, Label and Value to contain parameters for node-count (during install event), replica-count and capacity (during provision and deprovision events). These containers are inconvenient to look up on the Google Analytics dashboard.

v0.2.0 has replaced these containers in favour of new fields node_count, replica_count, vol_capacity. These contain the same values as before but they are visible alongside the usual metrics... Node OS, Engine Version, K8s version, etc.

Ref:
https://github.com/openebs/google-analytics-4/pull/8

pkg/driver/controller.go
niladrih: [This](https://github.com/openebs/google-analytics-4/blob/develop/usage/ping.go#L39) is the definition for the PingCheck function after the changes of https://github.com/openebs/google-analytics-4/pull/8.

The true/false is to trigger an immediate ping. With heartbeat, we'd want it trigger immediately and then trigger as per the ticker schedule, so that gets a `true`. With ping, we'd want it to avoid pinging immediately and just stick to the ticker schedule, so it gets a false. It's largely a lazy extension of the already existing PingCheck API.
```go
if pingImmediately {
	// Ping immediately.
	pingSender.Send()
}
```

Abhinandan-Purkait (approved):

avishnu (approved):

PR #556, opened 2024-07-12
test(features): add missing bdd

sinhaashish:
Add missing bdd

### DIFF
--- a/tests/bdd/zfs.feature
+++ b/tests/bdd/zfs.feature
@@ -32,7 +32,7 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
 
   Scenario Outline: Volume property change 
     Given a storage class is created with fsType as <fsType>
-    And pvc is created referencing this storage class with a deployment using the same pvc
+    And pvc is created referencing this storage class and a deployment using the same pvc
     When the zfsvolume properties like <properties> are updated
     Then the zfsvolume properties must reflect the updated value
     Examples:
@@ -49,7 +49,7 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
 
   Scenario Outline: Volume data size change
     Given a storage class is created with fsType as <fsType>
-    And pvc is created referencing this storage class with a deployment using the same pvc
+    And pvc is created referencing this storage class and a deployment using the same pvc
     When the zfsvolume volume type is <volume_type> and record size is updated to <record_size> and volume block size to <block_size> size is updated
     Then the volume type is <volume_type> must have the record sise as  <record_size> and volume block size as <block_size> size
     Examples:
@@ -65,7 +65,7 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
 
   Scenario Outline: Create a snapshot from a pvc
     Given a storage class is created with fsType as <fsType>
-    And pvc is created referencing this storage class with a deployment using the same pvc
+    And pvc is created referencing this storage class and a deployment using the same pvc
     And a zfsvolume is created and used by the application
     When the snapshot create command for the given pvc is run in the same namespace
     Then the snapshot must be created for the corresponding pvc with status as ready to use
@@ -78,7 +78,7 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
     
   Scenario Outline: Create a clone from a <source>
     Given a storage class is created with fsType as <fsType>
-    And pvc is created referencing this storage class with a deployment using the same pvc
+    And pvc is created referencing this storage class and a deployment using the same pvc
     And a zfsvolume is created and used by the application
     And a snapshot is be created for the corresponding pvc 
     When the clone create command is run for this snapshot and a deployment is created to use this clone
@@ -96,7 +96,7 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
 
   Scenario Outline: Resize of pvc
     Given a storage class is created with fsType as <fsType>
-    And pvc is created referencing the this storage class with a deployment using the same pvc
+    And pvc is created referencing the this storage class and a deployment using the same pvc
     When the size of pvc is update to a <new_capacity>
     Then the pvc must be of size <new_capacity> 
     Examples:
@@ -112,16 +112,6 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
     When a deployment is created using the same pvc
     Then zfsvolume must be created and used by the application
 
-  Scenario Outline: Volume property change for a raw block volume
-    Given a storage class is created without any fstype
-    And a pvc with volumeMode as Bolck is created referencing this storage class
-    When a deployment is created using the same pvc
-    When the zfsvolume properties <properties> are updated
-    Then the zfsvolume properties must reflect the changed value for the <properties>
-    Examples:
-      | properties     |
-      |  compression   |
-      |  dedup         |
 
   Scenario Outline: Volume data size change for a raw block volume
     Given a storage class is created without any fstype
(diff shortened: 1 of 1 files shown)


tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
+    Given a storage class is created with fsType as <fsType>
+    And pvc is created referencing this same storage class with zfsvolume in Ready state
+    And a deployment is created using the same pvc which is in running state
+    When resize volume requested is by updating the PVC yaml
+    Then corresponding pvc must be resized
+    Examples:
+      | fsType |
+      |  zfs   |
+      |  ext4  |
+      |  xfs   |
```
Abhinandan-Purkait, line 142: btrfs as well?

tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
\ No newline at end of file
+    
+############################################################################################################
+################## The below bdd need to be implemented in ci tests ########################################
+############################################################################################################
+
+
+  Scenario Outline: Online volume resize
+    Given a storage class is created with fsType as <fsType>
+    And pvc is created referencing this same storage class with zfsvolume in Ready state
```
Abhinandan-Purkait: what does `zfsvolume in Ready state` mean here in context of creation?

tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
+    And a deployment is created using the same pvc which is in running state
+    When resize volume requested is by updating the PVC yaml
```
Abhinandan-Purkait: PVC yaml --> PVC resource

tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
+    Then corresponding pvc must be resized
```
Abhinandan-Purkait, line 137: We need steps to validate in the lowest layer as well, i.e. the dataset

tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
+    Examples:
+      | fsType |
+      |  zfs   |
+      |  ext4  |
+      |  xfs   |
+
+  Scenario Outline: Create a snapshot from a pvc for raw block volume
+    Given a storage class is created without any fstype
```
Abhinandan-Purkait: There is no fstype needed for raw block right, let's not mention it?

tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
+    And a pvc with volumeMode as Bolck is created referencing this storage class
+    And a deployment is created using the same pvc
+    When the snapshot create command for the given pvc is run in the same namespace
+    Then the snapshot must be created for the corresponding pvc with status as ready to use
```
Abhinandan-Purkait: Need step to validate the dataset.

tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
+
+    
+  Scenario Outline: Create a clone from a for raw block volume
+    Given a storage class is created without any fstype
```
Abhinandan-Purkait: fstype not needed here

tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
+    And a pvc with volumeMode as Bolck is created referencing this storage class
+    And a deployment is created using the same pvc
+    And a snapshot is be created for the corresponding pvc 
+    When the clone create command is run for this snapshot and a deployment is created to use this clone
+    Then the clone must be created from the <source> and the created deployment must use the cloned volume
```
Abhinandan-Purkait: <source> --> source?

Abhinandan-Purkait: Also you probably want to change the scenario in such a way that we consider both clone from snapshot and clone from volume. Maybe create two different scenarios if it's not possible to put it under 1

tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
+
+  Scenario Outline: Thin provisioning
+    Given a storage class is created with fsType as <fsType>
+    And pvc is created referencing this storage class with a deployment using the same pvc
+    And a zfsvolume is created and used by the application
+    When thinprovision value  is "yes" in the storage class
```
Abhinandan-Purkait: This should be a part of line 177, the storage class.

tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
+    Then the volume must be provisioned even if the ZPOOL does not have the enough capacity
```
Abhinandan-Purkait, line 149: Specify the capacities as well, otherwise it's difficult to understand. Example:- zpool 1GB, volume created 10GB

tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
+      |  zfs   |
+      |  ext4  |
+      |  xfs   |
+      |  btrfs |
+
+  Scenario Outline: Shared volume
+    Given a storage class is created with fsType as <fsType>
+    And pvc is created referencing this storage class with a deployment using the same pvc
+    And a zfsvolume is created and used by the application
+    When shared set to "yes" in storage class
```
Abhinandan-Purkait: Should be a part of line 190, properties are predefined not discovered later.

tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
+    Then the LocalPV-ZFS Driver will allow the volumes to be mounted by more than one pods
```
Abhinandan-Purkait: Change this to rather point out that two pods were deployed on same node and they both are running

tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
+      |  zfs   |
+      |  ext4  |
+      |  xfs   |
+      |  btrfs |
+
+  Scenario Outline: Supported compression algorithm
+    Given a storage class is created with fsType as zfs, ext4, xfs, btrfs
+    And pvc is created referencing this storage class with a deployment using the same pvc
+    When the zfsvolume compression is present 
+    Then the system must support the <compression_type> algorithm
```
Abhinandan-Purkait: move the <compression_type> as a part of sc param defined in line 203.

tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
+      | fsType |
+      |  zfs   |
+      |  ext4  |
+      |  xfs   |
+      |  btrfs |
+
+  Scenario Outline: Supported compression algorithm
+    Given a storage class is created with fsType as zfs, ext4, xfs, btrfs
+    And pvc is created referencing this storage class with a deployment using the same pvc
+    When the zfsvolume compression is present
```
Abhinandan-Purkait, line 160: What do we mean by this?

tests/bdd/zfs.feature
Abhinandan-Purkait: We should not say system must support, we just need to see if the volume is created successfully or not. Please change it accordingly

tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
+
+############################################################################################################
+################## The below bdd need to be implemented in e2e tests ########################################
+############################################################################################################
+
+  Scenario Outline: Allowed topologies 
+    Given a storage class with allowedTopologies is created with fsType as <fsType>
+    And pvc is created referencing this storage class with a deployment using the same pvc
+    And a zfsvolume is created and used by the application
+    When the topology for specific zpool is defined in the storage class
```
Abhinandan-Purkait, line 234: We don't need this stage? It's already defined in line 247

tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
+      |  zfs   |
+      |  ext4  |
+      |  xfs   |
+      |  btrfs |
+
+   Scenario Outline: Import Existing Volumes to LocalPV-ZFS
+    Given a storage class is created with fsType as <fsType>
+    And pvc is created referencing this storage class with a deployment using the same pvc
+    And a zfsvolume is created and used by the application
+    When a node fails, with all the volumes volumes and their data intact as disks are still in good shape
```
Abhinandan-Purkait: Let's not put that `with all the volumes volumes and their data intact as disks are still in good shape` how can the test verify that?

tests/bdd/zfs.feature
```diff
@@ -139,4 +139,133 @@ Feature: Validate volume provisioning for fsType zfs, ext4, xfs, btrfs
...
+    Then the disks can be moved to the other node and import the pool there
```
Abhinandan-Purkait: When the disks is attached to a different node
Then the zfs pool should be imported on that pool
And the zfsvolume resource should reflect that on creation of a volume referencing the pool on that node
And the application should keep running

Now checking the data loss part might be difficult, we might need some real data base etc to validate that probably.

tests/bdd/zfs.feature
sinhaashish: btrfs does not support online volume resize

tests/bdd/zfs.feature
sinhaashish: removed,

tests/bdd/zfs.feature
sinhaashish: Updated it in `Given`

tests/bdd/zfs.feature
sinhaashish: that the deployment is using the refreced pvc

tests/bdd/zfs.feature
sinhaashish: what do you mean by this, its a raw block

Abhinandan-Purkait (approved):

PR #557, opened 2024-07-12
ci: enhance pull_request workflow

Abhinandan-Purkait:
# Workflow in detail

1. A user creates a PR to develop or release/x.y branch.
2. The concerned GitHub action workflow kicks in on the PR.
3. First, it checks out and sets up the runner.
4. It then runs the lint. Shellcheck and helm lint. 
5. Verify the chart version, it should be of the format x.y+1.0-develop. If it’s anything else, the workflow fails. From now on any change to the helm chart in the PR would not need a helm version change. Users are not supposed to change the chart version in their PRs, even if it’s a helm change.
6. Extract the chart version from the chart. This tag would be used for building the image locally.
7. Build the images using the code in the branch and tag it with x.y+1.0-develop
8. Set up the minikube cluster.
9. Execute the CI tests by installing the helm chart from the deploy/helm/charts directory. Set the image pull policy to Never since we are supposed to use the images built in the previous step.
10. Execute the CI tests. We are to use the helm chart here.
11. If the test passes the CI succeeds else it fails.


.github/workflows/pull_request.yml
```diff
@@ -1,38 +1,25 @@
...
       - 'docs/**'
       - 'design/**'
       - 'changelogs/**'
       - '*.md'
+      - 'MAINTAINERS'
     branches:
-      # on pull requests to develop and release branches
       - develop
-      - 'v*'
+      - release/*
```
niladrih: This filter is invalid without the single quotes.

nit: should we narrow the filter down? [Look here](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#filter-pattern-cheat-sheet).
```suggestion
      - 'release/v[0-9]+.[0-9]+'
```

.github/workflows/pull_request.yml
```diff
@@ -41,9 +28,30 @@ jobs:
...
+      - uses: cachix/install-nix-action@v22
+      - name: Check if the chart follows a valid semver version
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' scripts/nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          branch_name="${{ github.event.pull_request.base.ref }}"
+          nix-shell --pure --run "./scripts/validate-version.sh --branch $branch_name" ./scripts/shell.nix
 
+      - name: Set up chart-testing
+        uses: helm/chart-testing-action@v2.6.1
```
niladrih: nit: Conflicting dependency sourcing. Should we source our dependencies only from one of the sources (i.e. nix-shell or github actions plugins)?

.github/workflows/pull_request.yml
```diff
@@ -71,18 +79,20 @@ jobs:
...
           cache: false
+        
+      - name: Build images locally
+        run: make zfs-driver-image || exit 1;
+
+      - name: Setup zfs pools
+        run: |
+          sudo apt-get install -y zfsutils-linux
+          truncate -s 100G /tmp/disk.img
+          sudo zpool create zfspv-pool `sudo losetup -f /tmp/disk.img --show`
```
niladrih: Should this be behind the nix-shell?

.github/workflows/pull_request.yml
```diff
@@ -99,50 +109,18 @@ jobs:
           echo "KUBECONFIG=$HOME/.kube/config" >> $GITHUB_ENV
           echo "OPENEBS_NAMESPACE=openebs" >> $GITHUB_ENV
 
-      - name: Build images locally
-        run: make zfs-driver-image || exit 1;
```
niladrih, line 103: It feels like this image that is being used in the CI tests. Are we building from the local changes?

.github/workflows/pull_request.yml
Abhinandan-Purkait: Why should this be behind nix shell?

.github/workflows/pull_request.yml
Abhinandan-Purkait: Yes, we build locally and test because this is a PR, there is no push happening here.

.github/workflows/pull_request.yml
Abhinandan-Purkait: Well, ideally yes, but using gh-actions at some places gives us some more than just the binary or the tool.

.github/workflows/pull_request.yml
niladrih: `apt-get` is tied to debian/ubuntu. And we are adding the nix-shell to pull dependencies. Could it sit beside semver and yq, and get pulled in together?

.github/workflows/pull_request.yml
niladrih: Oh, I see, this one is moved to line 87. I thought it's being removed.

.github/workflows/pull_request.yml
Abhinandan-Purkait: But the tests and image build is happening outside nix-shell, unless we want to put everything behind nix-shell, which I would not prefer as that would need more changes.

.github/workflows/pull_request.yml
Abhinandan-Purkait: I changed it to happen via nix-shell.

scripts/shell.nix
```diff
@@ -0,0 +1,8 @@
+with (import <nixpkgs> { });
```
niladrih: Not sure, but I think this doesn't actually use the sources.json. @tiagolobocastro might be able to tell.

scripts/validate-version.sh
```diff
@@ -0,0 +1,107 @@
...
+}
+
+if [ "$BRANCH_NAME" == "develop" ]; then
+  if [[ "$CHART_VERSION" != *"-develop" ]]; then
+    die "Chart version must include '-develop' for develop branch"
+  fi
+  if [[ "$APP_VERSION" != *"-develop" ]]; then
+    die "App version must include '-develop' for develop branch"
+  fi
+elif [[ "$BRANCH_NAME" =~ ^release/[0-9]+\.[0-9]+$ ]]; then
```
niladrih:
```suggestion
elif [[ "$BRANCH_NAME" =~ ^(release/[0-9]+\.[0-9]+)$ ]]; then
```

scripts/shell.nix
Abhinandan-Purkait: Yes it does, check in the workflow where we do prepopulate.

.github/workflows/pull_request.yml
```diff
@@ -99,50 +109,18 @@ jobs:
...
-
-      - name: Build
-        uses: docker/build-push-action@v5
-        with:
-          context: .
-          file: ./buildscripts/zfs-driver/zfs-driver.Dockerfile
-          push: false
-          platforms: linux/amd64, linux/arm64
-          tags: |
-            openebs/zfs-driver:ci
```
niladrih: This workflow tests the multi-arch build. Some CPU instructions may not translate well to other architectures. This workflow tests builds for linux-arm64 as well as linux/am64. Also it tests the multi-stage build. The build before the integration tests is a simpler single-stage build. It's quicker, and there the main intention is to test the code behaviour.

We shouldn't remove this, I think.

scripts/shell.nix
```diff
@@ -0,0 +1,9 @@
...
+mkShell {
+  name = "scripts-shell";
+  buildInputs = [
+    semver-tool
+    yq-go
+    chart-testing
```
niladrih: The github action plugin installs 3 tools, might be worth pulling in all 3. [Look here](https://github.com/helm/chart-testing-action/blob/main/action.yml).

```suggestion
    chart-testing
    yamllint
    yamale
```

.github/workflows/pull_request.yml
Abhinandan-Purkait: Okay, but why is the tag needed here?

scripts/validate-version.sh
```diff
@@ -0,0 +1,107 @@
...
+
+Options:
+  --branch <branch name>                    Name of the branch on which this workflow is running.
+
+Examples:
+  $(basename "$0") --branch develop
+EOF
+}
+
+die() {
```
niladrih: Duplicate of the die function up above?

scripts/shell.nix
Abhinandan-Purkait: Well, it works without them. Check the latest commit's run.

scripts/validate-version.sh
Abhinandan-Purkait: Ohh, I initially used that but then used die and forgot to remove it.

.github/workflows/pull_request.yml
niladrih: Tag isn't needed.

.github/workflows/pull_request.yml
niladrih: Fair enough. Makes sense.

.github/workflows/pull_request.yml
Abhinandan-Purkait: Thanks for pointing out this was an oversight. I didn't actually plan to remove this.

.github/workflows/pull_request.yml
Abhinandan-Purkait: @niladrih Don't think we need the v here right? The names would be release/10.11 and not release/v10.11 ?

niladrih: Yeah, the v should not be there.

pchandra19 (approved):
Verify the chart version, it should be of the format x.y+1.0-develop. If it’s anything else, the workflow fails. From now on any change to the helm chart in the PR would not need a helm version change. Users are not supposed to change the chart version in their PRs, even if it’s a helm change.

Just a few question on above : You mentioned the format should be "x.y+1.0-develop" 
a) That means we will not be releasing any patch versions ?
b) Why do we have to even increment y (y+1) ?

Abhinandan-Purkait:
> Just a few question on above : You mentioned the format should be "x.y+1.0-develop" 
> a) That means we will not be releasing any patch versions ? 

Yes we would be releasing patch versions.

>b) Why do we have to even increment y (y+1) ?

Once a release is made, develop becomes the vNext. The patch releases are done from release/x.y branches. Now whenever a release let's say `x.y.z` is made, the develop would now become `x.y+1-develop`  because we are working towards bumping the minor version and the release/x.y becomes `x.y.z+1-prerelease` because in release/x.y we are working towards bumping a patch version.

scripts/validate-version.sh
```diff
@@ -0,0 +1,102 @@
...
```
niladrih: There's two die functions on this script.

scripts/validate-version.sh
```diff
@@ -0,0 +1,102 @@
...
+      ;;
+    -h|--help)
+      help
+      exit 0
+      ;;
+    *)
+      help
+      die "Unknown option: $1"
+      ;;
+  esac
```
niladrih: The `shift 2` works here because the other options are halting options. If this script is extended with newer options, this might get in the way. Might be better to separate the shift out.
```suggestion
      shift
      ;;
    -h|--help)
      help
      exit 0
      ;;
    *)
      help
      die "Unknown option: $1"
      ;;
  esac
  shift
```

scripts/validate-version.sh
```diff
@@ -0,0 +1,102 @@
...
+done
+
+# Extract major and minor version from the branch name
+extract_major_minor() {
+  echo "$1" | awk -F/ '{print $2}'
+}
+
+if [ "$BRANCH_NAME" == "develop" ]; then
```
niladrih:
```suggestion
if [ "$BRANCH_NAME" = "develop" ]; then
```

or 

```suggestion
if [[ "$BRANCH_NAME" == "develop" ]]; then
```

scripts/validate-version.sh
niladrih, line 1: The name could be better IMO. This script is related to the helm chart. The name should include 'helm' or 'chart' in some form. Or maybe it could live inside a directory specifically meant to store helm stuff.

.github/workflows/pull_request.yml
```diff
@@ -1,38 +1,25 @@
...
       - 'docs/**'
       - 'design/**'
       - 'changelogs/**'
       - '*.md'
+      - 'MAINTAINERS'
     branches:
-      # on pull requests to develop and release branches
       - develop
-      - 'v*'
+      - 'release/[0-9]+.[0-9]+'
```
tiagolobocastro: why not just `release/**` ?

scripts/validate-version.sh
```diff
@@ -0,0 +1,102 @@
+#!/bin/bash
```
niladrih: Let's also use `#!/usr/bin/env bash` like we are doing in the other script, just to keep things consistent?

.github/workflows/pull_request.yml
```diff
@@ -41,9 +28,27 @@ jobs:
           reporter: github-pr-review
           path: '.'
           pattern: '*.sh'
+          
+      - uses: cachix/install-nix-action@v22
+      - name: Check if the chart follows a valid semver version
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' scripts/nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
```
tiagolobocastro: I'd suggest setting up the nix in separate run altogether so it's easy to read and copy&paste to other actions as needed

.github/workflows/pull_request.yml
```diff
@@ -41,9 +28,27 @@ jobs:
...
+          branch_name="${{ github.event.pull_request.base.ref }}"
+          nix-shell --pure --run "./scripts/validate-version.sh --branch $branch_name" ./scripts/shell.nix
 
-
-  unit-test:
+      - name: Run chart-testing (list-changed)
+        id: list-changed
+        run: |
+          changed=$(nix-shell --pure --run "ct list-changed --config ct.yaml" ./scripts/shell.nix)
```
tiagolobocastro: what does this changed signify here?

.github/workflows/pull_request.yml
```diff
@@ -71,18 +76,20 @@ jobs:
...
         with:
           go-version: 1.19.9
           cache: false
+        
+      - name: Build images locally
+        run: make zfs-driver-image || exit 1;
+
+      - name: Setup zfs pools
+        run: |
+          sudo apt-get install -y zfsutils-linux
```
tiagolobocastro: use zfs from nix-shell?

.github/workflows/pull_request.yml
```diff
@@ -99,17 +106,14 @@ jobs:
...
-
       - name: bootstrap
         run: make bootstrap
 
       - name: Running tests
         run: |
           make ci
           make sanity 
-
+
```
tiagolobocastro: whitespace
```suggestion

```

Makefile
```diff
@@ -54,7 +54,7 @@ endif
 
 
 ifeq (${IMAGE_TAG}, )
-  IMAGE_TAG = ci
+  IMAGE_TAG = 2.7.0-develop
```
tiagolobocastro: Should we have a file on the root where we define the version? Or perhaps simply read the version from the chart?

deploy/helm/charts/Chart.yaml
```diff
@@ -16,5 +16,5 @@ sources:
 - https://github.com/openebs/zfs-localpv
 dependencies:
   - name: crds
-    version: 2.6.0
+    version: 2.7.0-develop
```
tiagolobocastro, line 19: Do we still need to keep bumping this one btw?

deploy/helm/charts/values.yaml
```diff
@@ -2,7 +2,7 @@
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 release:
-  version: "2.6.0"
+  version: "2.7.0-develop"
```
tiagolobocastro: Interesting, why is this used, can we not access the version from helm?

scripts/nix/sources.json
```diff
@@ -0,0 +1,14 @@
+{
+    "nixpkgs": {
+        "branch": "release-22.11",
```
tiagolobocastro: this is ancient btw, why not 24.05?

scripts/shell.nix
tiagolobocastro: this is not needed for the CI as per the changes I see, though it would be needed if we want to use it locally, otherwise we'd all get our own versions

scripts/validate-version.sh
Abhinandan-Purkait: Any suggestions?

Abhinandan-Purkait: `validate-chart-version.sh` ?

scripts/validate-version.sh
```diff
@@ -0,0 +1,102 @@
...
+}
+
+if [ "$BRANCH_NAME" == "develop" ]; then
+  if [[ "$CHART_VERSION" != *"-develop" ]]; then
+    die "Chart version must include '-develop' for develop branch"
+  fi
+  if [[ "$APP_VERSION" != *"-develop" ]]; then
+    die "App version must include '-develop' for develop branch"
+  fi
+elif [[ "$BRANCH_NAME" =~ ^(release/[0-9]+\.[0-9]+)$ ]]; then
```
tiagolobocastro: we're not taking into all full spec, ex: release/2.8.1

tests/tests.test
tiagolobocastro, line 1: what is this file?

scripts/validate-version.sh
niladrih: Sure, sounds fine to me.

.github/workflows/pull_request.yml
Abhinandan-Purkait: That's what I did initially, @niladrih suggested this way.

scripts/validate-version.sh
niladrih: This pattern is unreachable based on the current design.

.github/workflows/pull_request.yml
Abhinandan-Purkait: You mean seperate step in the job?

tests/tests.test
niladrih: This is a compiled version of the bdd tests. This was put in (on accident) in a much older PR. This PR removes this and adds this glob to the gitignore.

.github/workflows/pull_request.yml
Abhinandan-Purkait: It just checks if the chart has changed or not, so that we can conditionally run the lint, might not be needed. Seems I missed adding the condition in the lint stage.

deploy/helm/charts/Chart.yaml
Abhinandan-Purkait: I am not sure what to do here. I just bumped as that's what we have been doing.

tests/tests.test
Abhinandan-Purkait: Ginko generates this, seems like it was checked in at some point.

scripts/validate-version.sh
Abhinandan-Purkait: You mean remove the + at the end?

.github/workflows/pull_request.yml
niladrih: It's like that to narrow it down, it's fine to use `release/*` or `release/**` as well. 👍

scripts/nix/sources.json
Abhinandan-Purkait: There is no specific reason, I used this because that's what we seem to use at mayastor-extensions.

.github/workflows/pull_request.yml
Abhinandan-Purkait: As in like the prepoplulate step.

Makefile
Abhinandan-Purkait: We can read it from the chart that should be doable. But would need yq for that.

.github/workflows/pull_request.yml
tiagolobocastro: Yes

Makefile
Abhinandan-Purkait: That means the tests scripts would now run in nix-shell?

.github/workflows/pull_request.yml
tiagolobocastro: I think having it a catch all will help find mistakes early, example if we create branch in not the format we expect

.github/workflows/pull_request.yml
tiagolobocastro: Any downside to always just running the lint?

Makefile
tiagolobocastro: You could still run them outside nix-shell, as long as yq is present?

deploy/helm/charts/Chart.yaml
tiagolobocastro: Yeah that's fine, it's more of a general question. CC @niladrih

scripts/validate-version.sh
tiagolobocastro: No, I mean, if we ever have a branch in the release/x.y.z format, this script say unknown right?
But if we don't want to support that then it's fine

.github/workflows/pull_request.yml
Abhinandan-Purkait: I don't think so, I will remove this step.

Makefile
Abhinandan-Purkait: Buy yq is coming from nix-shell if you see. That's the problem. Can we do something about that?

scripts/shell.nix
Abhinandan-Purkait: Have changed it

scripts/nix/sources.json
Abhinandan-Purkait: Changed to 24.05

ci/ci-test.sh
```diff
@@ -1,58 +1,42 @@
...
 
 set -e
 
-ZFS_OPERATOR=deploy/zfs-operator.yaml
 SNAP_CLASS=deploy/sample/zfssnapclass.yaml
-
 TEST_DIR="tests"
 
-
 # Prepare env for runnging BDD tests
```
dsharma-dc: nitpick: typo - `running` . Not introduced in this change though.

scripts/nix/sources.json
dsharma-dc: I see 23.05 currently. typo?

niladrih (changes requested):
The chart-release workflow might run as soon as this PR is merged, due to the helm chart change. We might want to disable that workflow. Temporarily removing the `on` YAML body (with a comment) could be enough.

scripts/nix/sources.json
Abhinandan-Purkait: No, we later decided to not go with the latest, as that might not have some packages that might be for our code.

Abhinandan-Purkait:
> The chart-release workflow might run as soon as this PR is merged, due to the helm chart change. We might want to disable that workflow. Temporarily removing the `on` YAML body (with a comment) could be enough.

Done. Also wondering would it be a good idea to disable the build.yml workflow.

niladrih:
> > The chart-release workflow might run as soon as this PR is merged, due to the helm chart change. We might want to disable that workflow. Temporarily removing the `on` YAML body (with a comment) could be enough.
> 
> Done. Also wondering would it be a good idea to disable the build.yml workflow.

IMO we should let the build workflow run because there's WIP PRs and contributions coming in and contributors might want to work with their built develop branch images.

Makefile
```diff
@@ -53,8 +53,12 @@ ifeq (${DBUILD_SITE_URL}, )
 endif
 
 
+# Set the path to the Chart.yaml file
+ROOT_DIR:=$(dir $(realpath $(firstword $(MAKEFILE_LIST))))
+CHART_YAML:=${ROOT_DIR}/deploy/helm/charts/Chart.yaml
+
 ifeq (${IMAGE_TAG}, )
-  IMAGE_TAG = ci
+  IMAGE_TAG := ${shell yq e '.version' $(CHART_YAML)}
```
niladrih: This closely ties this makefile to yq, and maybe even the nix-shell. This will break some workflows. Also, I think, building this image locally shouldn't require specialised YAML parsing tools. It makes sense in our CI context, because we make use of `yq` extensively in our scripts outside of this.

How about this way of doing it?
```console
IMAGE_TAG := $(shell awk -F': ' '/^version:/ {print $$2}' $(CHART_YAML))
```

Also, should this be appVersion instead?

Abhinandan-Purkait: Well, as we decided appVersion and chart version would now be tied, shouldn't matter. Also why do you think it would be an issue to add yq in the makefile? We are anyhow using yq in other script. All workflows will now work this way.

Abhinandan-Purkait: Also, using yq is less prone to errors I feel.

niladrih: Because, a common workflow is... change code, then build the image with your changes. `yq` is not a common tool to have around. It should require `go` and `docker buildx`, kind of odd to require `yq`.

You're right, that this isn't exact, but close enough, I feel.

Abhinandan-Purkait: Alright.

niladrih (approved):

tiagolobocastro (approved):

PR #564, opened 2024-07-17
chore: replace CRD with auto-generated copy

niladrih:


Abhinandan-Purkait (approved):
LGTM. We will take this in once the CI changes are done.

pkg/apis/openebs.io/zfs/v1/zz_generated.deepcopy.go
```diff
@@ -29,6 +29,7 @@ import (
 func (in *Pool) DeepCopyInto(out *Pool) {
 	*out = *in
 	out.Free = in.Free.DeepCopy()
+	out.Used = in.Used.DeepCopy()
```
tiagolobocastro, line 32: ouch, could this have caused any bugs?

niladrih: Not really. It'd require another thread to be working on this exact resource at the same time.

PR #565, opened 2024-07-17
ci: enhance build.yml workflow

Abhinandan-Purkait:
### Workflow in detail

- A user PR gets merged to the develop or release/x.y branch.
- The concerned GitHub action workflow kicks in on the PR.
- First, it checks out and sets up the runner.
- It then runs the lint. Lints include helm lint. 
- Verify the chart version, it should be of the format x.y+1.0-develop or x.y.z-prerelease. If it’s anything else, the workflow fails. 
- Extract the chart version from the chart. This tag would be used for building the image locally.
- Build the images using the code in the branch and tag it with x.y+1.0-develop or x.y.z-prerelease.
- Set up the minikube cluster.
- Execute the CI tests by installing the helm chart from the deploy/helm/charts directory. Set the image pull policy to Never since we are supposed to use the images built in the previous step.
- Execute the CI tests. We are to use the helm chart here.
- If the test passes the CI succeeds else it fails.
- Push the images to the docker hub registry ?
- Package the helm chart and push it to the gh-pages ?

### NOTE
This PR removes the `ansible` step, that used to create image for e2e-test. e2e-test for openebs engines are now supposed to be carried out in the https://github.com/openebs/openebs-e2e repo going forward.

PR #566, opened 2024-07-17
ci: enhance build.yml workflow

Abhinandan-Purkait:
### Workflow in detail

- A user PR gets merged to the develop or release/x.y branch.
- The concerned GitHub action workflow kicks in on the PR.
- First, it checks out and sets up the runner.
- It then runs the lint. Lints include helm lint. 
- Verify the chart version, it should be of the format x.y+1.0-develop or x.y.z-prerelease. If it’s anything else, the workflow fails. 
- Extract the chart version from the chart. This tag would be used for building the image locally.
- Build the images using the code in the branch and tag it with x.y+1.0-develop or x.y.z-prerelease.
- Set up the minikube cluster.
- Execute the CI tests by installing the helm chart from the deploy/helm/charts directory. Set the image pull policy to Never since we are supposed to use the images built in the previous step.
- Execute the CI tests. We are to use the helm chart here.
- If the test passes the CI succeeds else it fails.
- Push the images to the docker hub registry ?
- Package the helm chart and push it to the gh-pages ?


Abhinandan-Purkait:
Duplicate #565

PR #567, opened 2024-07-19
ci: add branch preparation CI changes

Abhinandan-Purkait:
## Detailed workflow

- We create the release/x.y branch on the repo from develop.
- This concerned GitHub workflow kicks in.
- The workflow extracts the x.y from the branch name i.e. release/x.y.
- The workflow updates the chart version to x.y.0-prerelease and creates a PR against the release/x.y branch.
- We are supposed to merge the PRs manually. Once the above PR is merged the PR merge workflow for the release/x.y kicks  in.

This also adds the script that is capable of updating release and develop branches after tag is cut.

### DIFF
--- a/.github/workflows/branch_preparation.yml
+++ b/.github/workflows/branch_preparation.yml
@@ -1,0 +1,131 @@
+name: Branch Preparation
+
+on:
+  push:
+    branches:
+      - 'release/**'
+    tags:
+      - 'v[0-9]+.[0-9]+.[0-9]+**'
+
+jobs:
+  update_release_branch_after_release:
+    runs-on: ubuntu-latest
+    if: ${{ github.ref_type == 'tag' }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: cachix/install-nix-action@v22
+      - name: Pre-populate nix-shell
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          nix-shell --pure --run "echo" ./shell.nix
+
+      - name: Test the chart version updater script
+        run: |
+            nix-shell --pure --run "./scripts/test-update-chart-version.sh" ./shell.nix
+
+      - name: Modify the chart version based on the tag
+        run: |
+          tag=${{ github.ref_name }}
+          echo "BASE=$(nix-shell --pure --run "./scripts/update-chart-version.sh --tag $tag" ./shell.nix)" >> $GITHUB_ENV
+
+      - name: Create Pull Request to release
+        if:  ${{ env.BASE }}
+        id: cpr
+        uses: peter-evans/create-pull-request@v5
+        with:
+          base: ${{ env.BASE }}
+          commit-message: "chore(ci): update helm chart versions and/or git submodules"
+          committer: GitHub <noreply@github.com>
+          author: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
+          title: "Prepare release branch after release ${{ github.ref_name }}"
+          labels: |
+              update-release-branch
+              automated-pr
+          draft: false
+          signoff: true
+          branch: "create-pull-request/patch-${{ env.BASE }}"
+          token: ${{ secrets.GITHUB_TOKEN }}
+ 
+  update_develop_branch_on_release_branch_creation:
+    runs-on: ubuntu-latest
+    if: ${{ github.ref_type == 'branch' }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - run: |
+          git checkout develop
+
+      - uses: cachix/install-nix-action@v22
+      - name: Pre-populate nix-shell
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          nix-shell --pure --run "echo" ./shell.nix
+
+      - name: Test the chart version updater script
+        run: |
+            nix-shell --pure --run "./scripts/test-update-chart-version.sh" ./shell.nix
+
+      - name: Modify the chart version based on the branch name for develop
+        run: |
+            branch_name=${{ github.ref_name }}
+            nix-shell --pure --run "./scripts/update-chart-version.sh --branch $branch_name --type develop" ./shell.nix
+
+      - name: Create Pull Request to develop
+        id: cpr
+        uses: peter-evans/create-pull-request@v5
+        with:
+          base: develop
+          commit-message: "chore(ci): update helm chart versions and/or git submodules"
+          committer: GitHub <noreply@github.com>
+          author: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
+          title: "Prepare develop branch on ${{ github.ref_name }} creation"
+          labels: |
+            update-develop-branch
+            automated-pr
+          draft: false
+          signoff: true
+          branch: "create-pull-request/patch-develop"
+          token: ${{ secrets.GITHUB_TOKEN }}
+
+  prepare_release_branch_on_creation:
+    runs-on: ubuntu-latest
+    if: ${{ github.ref_type == 'branch' }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: cachix/install-nix-action@v22
+      - name: Pre-populate nix-shell
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          nix-shell --pure --run "echo" ./shell.nix
+
+      - name: Test the chart version updater script
+        run: |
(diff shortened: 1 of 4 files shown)


.github/workflows/branch_preparation.yml
```diff
@@ -0,0 +1,116 @@
...
+jobs:
+  prepare_develop_branch_after_release:
+    runs-on: ubuntu-latest
+    if: ${{ github.ref_type == 'tag' }}
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - run: |
+          git checkout develop
```
tiagolobocastro: I wonder if here we should trigger an action from the develop code itself.
Let's image what happens if the structure of the repo changes, then these steps here might not work on develop?
Can we instead trigger an action from the develop branch instead?

.github/workflows/branch_preparation.yml
```diff
@@ -0,0 +1,116 @@
...
+      - uses: cachix/install-nix-action@v22
+      - name: Pre-populate nix-shell
+        run: |
+          export NIX_PATH=nixpkgs=$(jq '.nixpkgs.url' nix/sources.json -r)
+          echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
+          nix-shell --pure --run "echo" ./shell.nix
+      - name: Check if the chart is publishable
+        run: |
+            tag=${{ github.ref_name }}
+            echo "BASE=$(nix-shell --pure --run "./scripts/update-chart-version.sh --tag $tag --type develop" ./shell.nix)" >> $GITHUB_ENV
```
tiagolobocastro: What happens if this is a patch release? In this case we don't want to modify develop, is this taken care of?
Maybe add some simple test cases for the update script to make it clear what is meant to happen for different releases

.github/workflows/branch_preparation.yml
```diff
@@ -0,0 +1,116 @@
...
+          committer: GitHub <noreply@github.com>
+          author: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
+          title: "Prepare develop branch after release ${{ github.ref_name }}"
+          labels: |
+            prepare-develop-branch
+            automated-pr
+          draft: false
+          signoff: true
+          reviewers: |
+            Abhinandan-Purkait
```
tiagolobocastro: Better to leave as the default? You can add yourself as a reviewer to the zfs repo

scripts/update-chart-version.sh
```diff
@@ -0,0 +1,140 @@
...
+}
+
+create_version_from_release_branch() {
+  if [[ "$BRANCH_NAME" =~ ^(release/[0-9]+\.[0-9]+)$ ]]; then
+    local EXTRACTED_VERSION=$(echo "$BRANCH_NAME" | grep -oP '(?<=release/)\d+\.\d+')
+    CURRENT_CHART_VERSION=$(yq e '.version' "$CHART_YAML")
+    if [[ "$CURRENT_CHART_VERSION" == *"-develop" ]]; then
+      VERSION="${EXTRACTED_VERSION}.0-prerelease"
+    elif [[ "$CURRENT_CHART_VERSION" == *"-prerelease" ]]; then
+      VERSION=$CURRENT_CHART_VERSION
```
tiagolobocastro: here we should leave it unset?

scripts/update-chart-version.sh
```diff
@@ -0,0 +1,140 @@
...
+  create_version_from_tag
+else
+  help
+  die "Either --branch or --tag and --type must be specified."
+fi
+
+if [[ -n $VERSION ]]; then
+  update_chart_yaml "$VERSION" "$VERSION"
+else
+  die "Failed to update the chart versions"
```
tiagolobocastro: Do we always expect version to be set?

.github/workflows/branch_preparation.yml
Abhinandan-Purkait: We are checking out the develop here that means the script is running from the develop branch so it should work?
The fetch-depth: 0 will pull in all changes at that point in time, so when we checkout develop, this would run from develop code.

.github/workflows/branch_preparation.yml
Abhinandan-Purkait: Yes, if it's a patch release then the minor version bump for the develop will still be same, so will be no op for develop.

scripts/update-chart-version.sh
Abhinandan-Purkait: No, I have made the scripts to fail if it's unset.

scripts/update-chart-version.sh
Abhinandan-Purkait: Yes, the branches in the script always set the versions, even if it would result in a no op.

.github/workflows/branch_preparation.yml
Abhinandan-Purkait: I have added the tests.

.github/workflows/branch_preparation.yml
tiagolobocastro: Yes but the steps here are defined in the action, example if you rename `./scripts/test-update-chart-version.sh` then the job would fail...
Anyway don't know how simple it is to "call" another workflow on-demand, so probably fine as is

scripts/test-update-chart-version.sh
```diff
@@ -0,0 +1,53 @@
...
+  else
+    echo "FAIL"
+    echo "Expected: $expected_output"
+    echo "Got: $output"
+  fi
+  echo "----------------------------------------"
+}
+
+# Define test cases
+run_test "Test 1: Generate version from release branch name" \
```
tiagolobocastro: Does the test cater for when we are not on develop branch?

scripts/test-update-chart-version.sh
```diff
@@ -0,0 +1,53 @@
...
+         "1.2.0-prerelease" \
+         --branch "release/1.2" --dry-run
+
+run_test "Test 2: Invalid branch name" \
+         "Invalid branch name format. Expected 'release/x.y' only" \
+         --branch "feature/1.2" --dry-run
+
+run_test "Test 3: Valid tag with type release [first release]" \
+         "1.2.1-prerelease" \
+         --tag "v1.2.0" --type "release" --dry-run
```
tiagolobocastro: hmm what happens if we tag an rc?

scripts/test-update-chart-version.sh
Abhinandan-Purkait: Yes, if you see the caller of this test, it happens on all jobs.

.github/workflows/branch_preparation.yml
Abhinandan-Purkait: Yeah, this probably won't change very often, so should be fine?

scripts/test-update-chart-version.sh
Abhinandan-Purkait: I haven't added to handle rc at the moment. Just adding the framework now. Need to think about rc.

scripts/test-update-chart-version.sh
tiagolobocastro: What I meant is, we shouldn't test with current Chart in workspace, we should add overrides and test for all cases, ex: develop and release, if you see what I mean?

tiagolobocastro (approved):

scripts/test-update-chart-version.sh
Abhinandan-Purkait: Done. Added test as well.

scripts/test-update-chart-version.sh
Abhinandan-Purkait: Okay. I added the test to run when we checkout develop as well. Would that be enough? Or you think we should add in the other workflows as well?

.github/workflows/branch_preparation.yml
```diff
@@ -0,0 +1,131 @@
...
+            nix-shell --pure --run "./scripts/test-update-chart-version.sh" ./shell.nix
+
+      - name: Modify the chart version based on the tag
+        run: |
+          tag=${{ github.ref_name }}
+          echo "BASE=$(nix-shell --pure --run "./scripts/update-chart-version.sh --tag $tag" ./shell.nix)" >> $GITHUB_ENV
+
+      - name: Create Pull Request to release
+        if:  ${{ env.BASE }}
+        id: cpr
```
tiagolobocastro, line 35: Are we missing the automated merge here?

.github/workflows/branch_preparation.yml
```diff
@@ -0,0 +1,131 @@
...
+        uses: peter-evans/create-pull-request@v5
+        with:
+          base: develop
+          commit-message: "chore(ci): update helm chart versions and/or git submodules"
+          committer: GitHub <noreply@github.com>
+          author: ${{ github.actor }} <${{ github.actor }}@users.noreply.github.com>
+          title: "Prepare develop branch on ${{ github.ref_name }} creation"
+          labels: |
+            prepare-develop-branch
```
tiagolobocastro:
```suggestion
            update-develop-branch
```

.github/workflows/branch_preparation.yml
Abhinandan-Purkait: No, it's intentional, it would be good to have a glance before merging?

tiagolobocastro: Up to you I guess, for mayastor we automated it - still goes via the PR checks and all

Abhinandan-Purkait: Yeah, I saw that I can add it, but it would needs the secrets and stuffs to be configured for this repo I think.

PR #572, opened 2024-07-25
test: add thin provision and volume type test

sinhaashish:


tests/utils.go
```diff
@@ -108,15 +107,39 @@ func IsPVCDeletedEventually(pvcName string) bool {
...
+		parameters["reservation"] = defaultReservation
+		if parameters["thinprovision"] == "no" {
+			parameters["reservation"] = capacity
 		}
+	} else {
+		parameters["quota"] = "-"
+		parameters["reservation"] = defaultReservation
 	}
+	delete(parameters, "thinprovision")
 }
```
Abhinandan-Purkait: We are not supposed to set the parameters manually, it would happen from the code.

Abhinandan-Purkait (approved):

PR #576, opened 2024-08-06
fix(controller): provision zfs volume if zfs volume already exists

AChangFeng:
- When the ZFSVolume is actually ready but the PV has not been created due to timeout or controller crash, the status of ZFSVolume will remain in Pending, which prevents the PV from being created and keeps the PVC in Pending as well.

**What this PR does?**:
- Clear finalizers of ZFSVolume in zfs.ProvisionVolume()
- When volume.ZVController.syncZV() is called, it will update the status of the ready ZFSVolume to the actual status instead of keep calling zfs.SetVolumeProp()

No

- This is an intermittent issue, and it is difficult to reproduce the scenarios manually

No

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/zfs/volume.go
+++ b/pkg/zfs/volume.go
@@ -402,8 +402,8 @@ func UpdateRestoreInfo(rstr *apis.ZFSRestore, status apis.ZFSRestoreStatus) erro
 }
 
 // GetUserFinalizers returns all the finalizers present on the ZFSVolume object
-// execpt the one owned by ZFS node daemonset. We also need to ignore the foregroundDeletion
-// finalizer as this will be present becasue of the foreground cascading deletion
+// except the one owned by ZFS node daemonset. We also need to ignore the foregroundDeletion
+// finalizer as this will be present because of the foreground cascading deletion
 func GetUserFinalizers(finalizers []string) []string {
 	var userFin []string
 	for _, fin := range finalizers {
@@ -417,17 +417,21 @@ func GetUserFinalizers(finalizers []string) []string {
 
 // IsVolumeReady returns true if volume is Ready
 func IsVolumeReady(vol *apis.ZFSVolume) bool {
-
-	if vol.Status.State == ZFSStatusReady {
-		return true
+	// The status was added to ZFSVolume since v0.8.0
+	if vol.Status.State != "" {
+		// For newer volumes created after v0.8.0, the status is sufficient to determine if the volume is ready
+		// If we check the finalizer to ensure the volume is Ready while the status is Pending or Failed
+		// the volume may never become Ready again when the controller provisions the volume again due to a timeout or controller crash
+		return vol.Status.State == ZFSStatusReady
 	}
 
-	// For older volumes, there was no Status field
+	// For older volumes created before v0.8.0, there was no Status field
 	// so checking the node finalizer to make sure volume is Ready
 	for _, fin := range vol.Finalizers {
 		if fin == ZFSFinalizer {
 			return true
 		}
 	}
+
 	return false
 }
--- a/pkg/zfs/volume_test.go
+++ b/pkg/zfs/volume_test.go
@@ -1,0 +1,40 @@
+package zfs
+
+import (
+	"testing"
+
+	apis "github.com/openebs/zfs-localpv/pkg/apis/openebs.io/zfs/v1"
+)
+
+func TestIsVolumeReady(t *testing.T) {
+	volGen := func(finalizer string, state string) *apis.ZFSVolume {
+		vol := &apis.ZFSVolume{}
+		if finalizer != "" {
+			vol.Finalizers = append(vol.Finalizers, finalizer)
+		}
+		if state != "" {
+			vol.Status.State = state
+		}
+		return vol
+	}
+	tests := []struct {
+		name         string
+		volFinalizer string
+		volState     string
+		want         bool
+	}{
+		{"Older volume is ready", ZFSFinalizer, "", true},
+		{"Older volume is not ready", "", "", false},
+		{"Newer volume is pending", "", ZFSStatusPending, false},
+		{"Newer volume is pending with finalizer", ZFSFinalizer, ZFSStatusPending, false},
+		{"Newer volume is ready with finalizer", ZFSFinalizer, ZFSStatusReady, true},
+		{"Newer volume is failed", "", ZFSStatusFailed, false},
+	}
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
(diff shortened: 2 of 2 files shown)


Abhinandan-Purkait:
@AChangFeng Can you give some more context on how updating the finalizers help here?

AChangFeng:
> @AChangFeng Can you give some more context on how updating the finalizers help here?

- If we don't reset finalizer, zfs.IsVolumeReady() always return true, this leading zfs.ZVController.syncZV()  keep calling zfs.SetVolumeProp()
- When we reset finalizer, zfs.IsVolumeReady() will return false, this leading zfs.ZVControlelr.syncZV() call zfs.UpdateZvolInfo()

Abhinandan-Purkait:
> > @AChangFeng Can you give some more context on how updating the finalizers help here?
> 
> * If we don't reset finalizer, zfs.IsVolumeReady() always return true, this leading zfs.ZVController.syncZV()  keep calling zfs.SetVolumeProp()
> * When we reset finalizer, zfs.IsVolumeReady() will return false, this leading zfs.ZVControlelr.syncZV() call zfs.UpdateZvolInfo()

Isn't that bit of the code for older volumes that did not have the status? Shouldn't hit if the zfsvolume having a status is ready?

AChangFeng:
> > > @AChangFeng Can you give some more context on how updating the finalizers help here?
> > 
> > 
> > 
> > * If we don't reset finalizer, zfs.IsVolumeReady() always return true, this leading zfs.ZVController.syncZV()  keep calling zfs.SetVolumeProp()
> > * When we reset finalizer, zfs.IsVolumeReady() will return false, this leading zfs.ZVControlelr.syncZV() call zfs.UpdateZvolInfo()
> 
> Isn't that bit of the code for older volumes that did not have the status? Shouldn't hit if the zfsvolume having a status is ready?

Think about this corner case: 
1. zfs.ProvisionVolume() is called by controller to provision a zfs volume
2. agent create a zfs volume then update status of ZFSVolume to Ready and append zfs.ZFSFinalizer to ZFSVolume
3. agent create zfs volume too slow so that controller call zfs.ProvisionVolume again
4. zfs.GetZFSVolume() call in zfs.ProvisionVolume return ZFSVolume and nil
5. do not reset finalizers of ZFSVolume, but set status of ZFSVolume to pending
6. agent will never update status of ZFSVolume to Ready or Failed because ZFSVolume already has zfs.ZFSFinalizer
7. zfs.checkVolCreation() call in zfs.ProvisionVolume() keep return fmt.Errorf("zfs: context deadline reached")
8. PV will never be provisioned

Abhinandan-Purkait:
> Think about this corner case:
> 
> 1. zfs.ProvisionVolume() is called by controller to provision a zfs volume
> 2. agent create a zfs volume then update status of ZFSVolume to Ready and append zfs.ZFSFinalizer to ZFSVolume
> 3. agent create zfs volume too slow so that controller call zfs.ProvisionVolume again
> 4. zfs.GetZFSVolume() call in zfs.ProvisionVolume return ZFSVolume and nil
> 5. do not reset finalizers of ZFSVolume, but set status of ZFSVolume to pending
> 6. agent will never update status of ZFSVolume to Ready or Failed because ZFSVolume already has zfs.ZFSFinalizer
> 7. zfs.checkVolCreation() call in zfs.ProvisionVolume() keep return fmt.Errorf("zfs: context deadline reached")
> 8. PV will never be provisioned

Yes, I agree to the flow you suggested here. I was wondering rather than resetting the finalizer, what if we fix the `IsVolumeReady` method, to check finalizer if and only if the status is empty. If a status is a valid value and is `Ready` return true if not return false, we don't even go to the finalizer check section. Does this sound viable to you?

I believe the issue happens because we do the check using finalizer even if we have the status, which should have been just enough otherwise to figure out if the volume is ready.

AChangFeng:
> > Think about this corner case:
> > 
> > 1. zfs.ProvisionVolume() is called by controller to provision a zfs volume
> > 2. agent create a zfs volume then update status of ZFSVolume to Ready and append zfs.ZFSFinalizer to ZFSVolume
> > 3. agent create zfs volume too slow so that controller call zfs.ProvisionVolume again
> > 4. zfs.GetZFSVolume() call in zfs.ProvisionVolume return ZFSVolume and nil
> > 5. do not reset finalizers of ZFSVolume, but set status of ZFSVolume to pending
> > 6. agent will never update status of ZFSVolume to Ready or Failed because ZFSVolume already has zfs.ZFSFinalizer
> > 7. zfs.checkVolCreation() call in zfs.ProvisionVolume() keep return fmt.Errorf("zfs: context deadline reached")
> > 8. PV will never be provisioned
> 
> Yes, I agree to the flow you suggested here. I was wondering rather than resetting the finalizer, what if we fix the `IsVolumeReady` method, to check finalizer if and only if the status is empty. If a status is a valid value and is `Ready` return true if not return false, we don't even go to the finalizer check section. Does this sound viable to you?
> 
> I believe the issue happens because we do the check using finalizer even if we have the status, which should have been just enough otherwise to figure out if the volume is ready.

LGTM. I have fixed the `IsVolumeReady` method with unit tests covered.

pkg/zfs/volume.go
```diff
@@ -417,9 +417,9 @@ func GetUserFinalizers(finalizers []string) []string {
 
 // IsVolumeReady returns true if volume is Ready
 func IsVolumeReady(vol *apis.ZFSVolume) bool {
-
-	if vol.Status.State == ZFSStatusReady {
-		return true
+	if vol.Status.State != "" {
```
tiagolobocastro, line 421: Would be good to add comment here explaining why this is needed

tiagolobocastro (approved):

pkg/zfs/volume.go
AChangFeng: I have added a comment explaining why this is necessary. Please let me know if any further details are needed.

Abhinandan-Purkait (approved):
LGTM

pkg/zfs/volume.go
tiagolobocastro: Thanks, maybe clarify newer to mention the version from which the status was added

AChangFeng: Thank you for the suggestion. I have clarified the version from which the status was added to provide better context.

tiagolobocastro (approved):

Abhinandan-Purkait:
@AChangFeng Are we good to merge?

AChangFeng:
> @AChangFeng Are we good to merge?

Yes, everything looks good to me.

AChangFeng:
> @AChangFeng Are we good to merge?

Yes, everything looks good to me.

PR #581, opened 2024-08-30
fix(ci): add ginkgo binary to nix shell

niladrih:
CI runs for PRs 578, 580 are failing with
```
/home/runner/work/zfs-localpv/zfs-localpv/ci/ci-test.sh: line 79: ginkgo: command not found
```
This adds the ginkgo CLI to the nix-shell and pins it to version used in the go.mod using niv dependencies.


niladrih:
This is not required, the issue was related to mismatched versions in the version of ginkgo used to write the tests and the ginkgo binary.

PR #588, opened 2024-09-17
fix(controller): Fix race of csi controller calls on the same volume

Lucaber:

When a CreateVolume requests takes longer then 10 seconds the request runs into a client-side timeout and gets retied by kubelet. While processing the new request, the old request is still being canceled on the server-side and the created ZFSVolume CR gets deleted.

1|2
-|-
CreateVolume call|
ZFSVolume CR created|
waiting for volume creation|
client timeout + context cancel|
| |CreateVolume call
| | ZFSVolume CR already exists
ZFSVolume CR deleted|
Return ERR|Return OK without creating a ZFSVolume

```
I0916 15:01:49.227721       1 grpc.go:72] GRPC call: /csi.v1.Controller/CreateVolume requests {"accessibility_requirements":{"preferred":[{"segments":{"beta.kubernetes.io/arch":"amd64","beta.kubernetes.io/os":"linux","kubernetes.io/arch":"amd64","kubernetes.io/hostname":"i04-1-b-node-i041b-0-0","kubernetes.io/os":"linux","m3.services/cluster":"i04-1-b","m3.services/worker-group":"i041b-0","node.systems.mittwald.cloud/placement-group":"i041b-0","openebs.io/nodeid":"i04-1-b-node-i041b-0-0","openebs.io/nodename":"i04-1-b-node-i041b-0-0"}}],"requisite":[{"segments":{"beta.kubernetes.io/arch":"amd64","beta.kubernetes.io/os":"linux","kubernetes.io/arch":"amd64","kubernetes.io/hostname":"i04-1-b-node-i041b-0-0","kubernetes.io/os":"linux","m3.services/cluster":"i04-1-b","m3.services/worker-group":"i041b-0","node.systems.mittwald.cloud/placement-group":"i041b-0","openebs.io/nodeid":"i04-1-b-node-i041b-0-0","openebs.io/nodename":"i04-1-b-node-i041b-0-0"}}]},"capacity_range":{"required_bytes":5368709120},"name":"pvc-eb11c446-f26c-4333-b3da-4ecb9c5b2457","parameters":{"compression":"off","csi.storage.k8s.io/pv/name":"pvc-eb11c446-f26c-4333-b3da-4ecb9c5b2457","csi.storage.k8s.io/pvc/name":"test-pvc0","csi.storage.k8s.io/pvc/namespace":"default","dedup":"off","fstype":"zfs","poolname":"kluster-pool/i04-1-b-128k","recordsize":"128k","shared":"yes","thinprovision":"yes"},"volume_capabilities":[{"AccessType":{"Mount":{"fs_type":"zfs"}},"access_mode":{"mode":1}}]}
I0916 15:01:58.853180       1 controller.go:289] zfs: trying volume creation kluster-pool/i04-1-b-128k/pvc-eb11c446-f26c-4333-b3da-4ecb9c5b2457 on node [i04-1-b-node-i041b-0-0]
I0916 15:01:59.104207       1 volume.go:139] zfs: waiting for volume kluster-pool/i04-1-b-128k/pvc-eb11c446-f26c-4333-b3da-4ecb9c5b2457 to be created on nodeid i04-1-b-node-i041b-0-0
I0916 15:02:00.105273       1 volume.go:170] zfs: volume kluster-pool/i04-1-b-128k/pvc-eb11c446-f26c-4333-b3da-4ecb9c5b2457 provisioning failed on nodeid i04-1-b-node-i041b-0-0 err: zfs: context deadline reached
I0916 15:02:00.235113       1 grpc.go:72] GRPC call: /csi.v1.Controller/CreateVolume requests {"accessibility_requirements":{"preferred":[{"segments":{"beta.kubernetes.io/arch":"amd64","beta.kubernetes.io/os":"linux","kubernetes.io/arch":"amd64","kubernetes.io/hostname":"i04-1-b-node-i041b-0-0","kubernetes.io/os":"linux","m3.services/cluster":"i04-1-b","m3.services/worker-group":"i041b-0","node.systems.mittwald.cloud/placement-group":"i041b-0","openebs.io/nodeid":"i04-1-b-node-i041b-0-0","openebs.io/nodename":"i04-1-b-node-i041b-0-0"}}],"requisite":[{"segments":{"beta.kubernetes.io/arch":"amd64","beta.kubernetes.io/os":"linux","kubernetes.io/arch":"amd64","kubernetes.io/hostname":"i04-1-b-node-i041b-0-0","kubernetes.io/os":"linux","m3.services/cluster":"i04-1-b","m3.services/worker-group":"i041b-0","node.systems.mittwald.cloud/placement-group":"i041b-0","openebs.io/nodeid":"i04-1-b-node-i041b-0-0","openebs.io/nodename":"i04-1-b-node-i041b-0-0"}}]},"capacity_range":{"required_bytes":5368709120},"name":"pvc-eb11c446-f26c-4333-b3da-4ecb9c5b2457","parameters":{"compression":"off","csi.storage.k8s.io/pv/name":"pvc-eb11c446-f26c-4333-b3da-4ecb9c5b2457","csi.storage.k8s.io/pvc/name":"test-pvc0","csi.storage.k8s.io/pvc/namespace":"default","dedup":"off","fstype":"zfs","poolname":"kluster-pool/i04-1-b-128k","recordsize":"128k","shared":"yes","thinprovision":"yes"},"volume_capabilities":[{"AccessType":{"Mount":{"fs_type":"zfs"}},"access_mode":{"mode":1}}]}
I0916 15:02:00.449153       1 volume.go:221] zfs: deleted the volume pvc-eb11c446-f26c-4333-b3da-4ecb9c5b2457
I0916 15:02:00.449299       1 controller.go:466] created the volume kluster-pool/i04-1-b-128k/pvc-eb11c446-f26c-4333-b3da-4ecb9c5b2457 on node i04-1-b-node-i041b-0-0
I0916 15:02:00.553778       1 grpc.go:81] GRPC response: {"volume":{"accessible_topology":[{"segments":{"openebs.io/nodeid":"i04-1-b-node-i041b-0-0"}}],"capacity_bytes":5368709120,"volume_context":{"openebs.io/cas-type":"localpv-zfs","openebs.io/poolname":"kluster-pool/i04-1-b-128k"},"volume_id":"pvc-eb11c446-f26c-4333-b3da-4ecb9c5b2457"}}
```

**What this PR does?**: Adds a per volume mutex to csi calls to prevent simultaneous requests

**Does this PR require any upgrade changes?**: No

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
-	"sync"
 	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
@@ -65,23 +64,7 @@ type controller struct {
 	k8sNodeInformer cache.SharedIndexInformer
 	zfsNodeInformer cache.SharedIndexInformer
 
-	volMutexes sync.Map
-}
-
-func (cs *controller) LockVolume(volume string) func() {
-	value, _ := cs.volMutexes.LoadOrStore(volume, &sync.Mutex{})
-	mtx := value.(*sync.Mutex)
-	mtx.Lock()
-	return func() { mtx.Unlock() }
-}
-
-func (cs *controller) LockVolumeWithSnapshot(volume string, snapshot string) func() {
-	unlockVol := cs.LockVolume(volume)
-	unlockSnap := cs.LockVolume(snapshot)
-	return func() {
-		unlockVol()
-		unlockSnap()
-	}
+	volumeLock *volumeLock
 }
 
 // NewController returns a new instance
@@ -90,6 +73,7 @@ func NewController(d *CSIDriver) csi.ControllerServer {
 	ctrl := &controller{
 		driver:       d,
 		capabilities: newControllerCapabilities(),
+		volumeLock:   newVolumeLock(),
 	}
 	if err := ctrl.init(); err != nil {
 		klog.Fatalf("init controller: %v", err)
@@ -469,7 +453,7 @@ func (cs *controller) CreateVolume(
 	contentSource := req.GetVolumeContentSource()
 	pvcName := helpers.GetInsensitiveParameter(&parameters, "csi.storage.k8s.io/pvc/name")
 
-	unlock := cs.LockVolume(volName)
+	unlock := cs.volumeLock.LockVolume(volName)
 	defer unlock()
 
 	if contentSource != nil && contentSource.GetSnapshot() != nil {
@@ -515,7 +499,7 @@ func (cs *controller) DeleteVolume(
 	}
 
 	volumeID := strings.ToLower(req.GetVolumeId())
-	unlock := cs.LockVolume(volumeID)
+	unlock := cs.volumeLock.LockVolume(volumeID)
 	defer unlock()
 
 	// verify if the volume has already been deleted
@@ -635,7 +619,7 @@ func (cs *controller) ControllerExpandVolume(
 			"ControllerExpandVolume: no volumeID provided",
 		)
 	}
-	unlock := cs.LockVolume(volumeID)
+	unlock := cs.volumeLock.LockVolume(volumeID)
 	defer unlock()
 
 	/* round off the new size */
@@ -733,7 +717,7 @@ func (cs *controller) CreateSnapshot(
 	if err != nil {
 		return nil, err
 	}
-	unlock := cs.LockVolumeWithSnapshot(volumeID, snapName)
+	unlock := cs.volumeLock.LockVolumeWithSnapshot(volumeID, snapName)
 	defer unlock()
 
 	snapTimeStamp := time.Now().Unix()
@@ -831,7 +815,7 @@ func (cs *controller) DeleteSnapshot(
 		// should succeed when an invalid snapshot id is used
 		return &csi.DeleteSnapshotResponse{}, nil
 	}
-	unlock := cs.LockVolumeWithSnapshot(snapshotID[0], snapshotID[1])
+	unlock := cs.volumeLock.LockVolumeWithSnapshot(snapshotID[0], snapshotID[1])
 	defer unlock()
 	if err := zfs.DeleteSnapshot(snapshotID[1]); err != nil {
 		return nil, status.Errorf(
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/container-storage-interface/spec/lib/go/csi"
@@ -63,6 +64,15 @@ type controller struct {
 
 	k8sNodeInformer cache.SharedIndexInformer
 	zfsNodeInformer cache.SharedIndexInformer
+
+	volMutexes sync.Map
+}
+
+func (cs *controller) LockVolume(volume string) func() {
+	value, _ := cs.volMutexes.LoadOrStore(volume, &sync.Mutex{})
+	mtx := value.(*sync.Mutex)
+	mtx.Lock()
+	return func() { mtx.Unlock() }
 }
 
 // NewController returns a new instance
@@ -450,6 +460,9 @@ func (cs *controller) CreateVolume(
 	contentSource := req.GetVolumeContentSource()
 	pvcName := helpers.GetInsensitiveParameter(&parameters, "csi.storage.k8s.io/pvc/name")
 
+	unlock := cs.LockVolume(volName)
+	defer unlock()
+
 	if contentSource != nil && contentSource.GetSnapshot() != nil {
 		snapshotID := contentSource.GetSnapshot().GetSnapshotId()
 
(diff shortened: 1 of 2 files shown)


niladrih:
Hmm... this is upsetting. I suppose all of the tests on our Go repos need to use the `-race` flag on CI pipeline.
I've created this ticket: https://github.com/openebs/openebs/issues/3784

Abhinandan-Purkait (approved):

pkg/driver/controller.go
```diff
@@ -705,6 +722,10 @@ func (cs *controller) CreateSnapshot(
 	if err != nil {
 		return nil, err
 	}
+	unlockVol := cs.LockVolume(volumeID)
```
tiagolobocastro: ok for now but if we have more uses of the volume&snapshot, might be worth having a function to lock, ensuring the locks are always taken in the same order to avoid deadlock

tiagolobocastro (approved):
Thanks!
I wonder if we're missing this in other places...

pkg/driver/controller.go
```diff
@@ -63,6 +64,15 @@ type controller struct {
 
 	k8sNodeInformer cache.SharedIndexInformer
 	zfsNodeInformer cache.SharedIndexInformer
+
+	volMutexes sync.Map
+}
+
+func (cs *controller) LockVolume(volume string) func() {
+	value, _ := cs.volMutexes.LoadOrStore(volume, &sync.Mutex{})
```
tiagolobocastro: Just wondering, are entries automatically GC'd?
Otherwise this may grow "forever" ?

Lucaber:
> Hmm... this is upsetting. I suppose all of the tests on our Go repos need to use the `-race` flag on CI pipeline. I've created this ticket: [openebs/openebs#3784](https://github.com/openebs/openebs/issues/3784)

This isn't a go data race, `-race` can't detect this issue.
To reproduce this issue I created 50 PVCs with `WaitForFirstConsumer` and then started a single pod with all 50 volumes to create a high load on both kubelet and openebs.

pkg/driver/controller.go
Lucaber: No, they are not getting GC'd. The Mutex can only be removed from the Map if there are no other references to the Mutex. Im not sure if this relatively small leak (the volume string key is probably larger than the mutex) is worth the additional complexity.

pkg/driver/controller.go
Lucaber: Good point, added a new function `LockVolumeWithSnapshot`

pkg/driver/controller.go
Abhinandan-Purkait: Well, it might grow to large extents? Also what about deleleted volume, the entries for them will also stay in the map right?

tiagolobocastro: Seems most drivers on kubernetes-csi actually use the same exact implementation: https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/common/volume_lock.go
Which handles it by not using `N` locks but rather keep a list of volumes and deleting the volume entry.

Lucaber: They are handling it a bit different, instead of waiting for the lock, the csi call errors instantly with `An operation with the given Volume ID %s already exists` https://github.com/kubernetes-sigs/gcp-compute-persistent-disk-csi-driver/blob/master/pkg/gce-pd-csi-driver/node.go#L140

tiagolobocastro: Yeah guess we'd need to do the same here as well. Perhaps that's also a better approach, letting the caller retry.
Should we use this new way instead? @Lucaber @Abhinandan-Purkait

Lucaber: I have looked at a few different Implementations. TopoLVM also uses a single map to add and remove keys but using a sync.Cond that locks the second concurrent call like my implementation: https://github.com/topolvm/topolvm/blob/main/internal/driver/lock.go This seems to be the best way to handle this.

Alternatively we could also use the keymutex implementation by kubernetes that used a few mutexes which are being shared between volumes. But this would require special handling to ensure that the snapshot and parent volume are using different mutexes (to prevent deadlocking). https://github.com/kubernetes/utils/blob/master/keymutex/hashed.go

Abhinandan-Purkait: The topolvm one looks reasonable, wdyt?

tiagolobocastro (approved):
Thanks, lgtm, though I still somehow wonder if returning an error would be a safer approach.

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
@Lucaber Please rebase your branch. @tiagolobocastro Are we good to take this in now?

tiagolobocastro:
> @Lucaber Please rebase your branch. @tiagolobocastro Are we good to take this in now?

Yeah I had approved already

Abhinandan-Purkait:
@Lucaber Mark the conversations as resolved if you don't think there is anything more to do before we can merge this.

Abhinandan-Purkait:
Merging this once the CI passes. Thanks for the contribution @Lucaber

PR #599, opened 2024-11-05
ci: add Fossa CLI workflow

Abhinandan-Purkait:
- Adds the fossa cli workflow.

### DIFF
--- a/.github/workflows/fossa.yml
+++ b/.github/workflows/fossa.yml
@@ -1,0 +1,17 @@
+name: Fossa CLI
+on:
+  push:
+    branches:
+      - 'develop'
+      - 'release/**'
+
+jobs:
+  fossa-scan:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          submodules: recursive
+      - uses: fossas/fossa-action@v1.4.0
+        with:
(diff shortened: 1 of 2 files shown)


.github/workflows/fossa.yml
```diff
@@ -0,0 +1,17 @@
+name: Fossa CLI
+on:
+  push:
+    branches:
+      - 'develop'
+      - 'release/**'
```
niladrih, line 6: Should these run on the pull request workflow, so that license issue could be caught early.
Not sure how this works, assuming this scans the code and dependencies for supported licenses, which are input somehow?

Abhinandan-Purkait: The fossa cli github action if for pushing the analytics to the fossa server and not catching the issues realtime, the calculation happens on the server I believe. So it cannot be a part of PR workflow.

tiagolobocastro: There's a different github app which can do this per PR, IIRC it's an app not an action, so probably doesn't need any yml action changes, just needs to be configured as such

tiagolobocastro (approved):

niladrih (approved):

Abhinandan-Purkait:
@tiagolobocastro The project url for zfs-localpv seems to be incorrect in the fossa portal.

tiagolobocastro (approved):

PR #605, opened 2024-12-18
fix(chart): move app=componentName label out of csi-node matchLabels

### LINKED ISSUE #593: Develop chart modifies csi-node daemonset MatchLabels
Upgrade to 2.7.0-develop chart from 4.1.0 and 4.0.1 failed with  
```
Upgrade "openebs" failed: cannot patch "openebs-zfs-localpv-node" with kind DaemonSet: DaemonSet.apps "openebs-zfs-localpv-node" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app":"openebs-zfs-node", "name":"openebs-zfs-node", "release":"openebs"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```

The below label was added to MatchLabels section which is immutable
"app":"openebs-zfs-node"

niladrih:
The label `app: openebs-zfs-node` was missing on the csi-node components. However, adding it to the selector labels section breaks upgrade from previous versions.

This need not live in matchLabels, and could simply live in componentLabels

Fixes https://github.com/openebs/zfs-localpv/issues/593

### DIFF
--- a/deploy/helm/charts/templates/_helpers.tpl
+++ b/deploy/helm/charts/templates/_helpers.tpl
@@ -98,7 +98,6 @@ Create labels for openebs zfs-localpv controller
 Create match labels for openebs zfs-localpv node daemon
 */}}
 {{- define "zfslocalpv.zfsNode.matchLabels" -}}
-app: {{ .Values.zfsNode.componentName | quote }}
 name: {{ .Values.zfsNode.componentName | quote }}{{ if or (not (hasKey .Values "enableHelmMetaLabels")) .Values.enableHelmMetaLabels }}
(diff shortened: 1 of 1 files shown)


tiagolobocastro:
Should our tests have caught this, are we still lacking in the testing?

Abhinandan-Purkait (approved):

sinhaashish (approved):

niladrih:
> Should our tests have caught this, are we still lacking in the testing?

There is a gap in testing for upgrade scenarios.

PR #607, opened 2024-12-20
fix(controller): prevent zfs volume cr deletion if snapshot exists

sinhaashish:
## Pull Request template

Please, go through these steps before you submit a PR.

**Why is this PR required? What issue does it fix?**:
Upon pvc deletion the zfs snapshot also gets deleted. this can be verified using ssh into the node and running the command ` zfs list -t all`

**What this PR does?**:
It stops deletion of CR if snapshot exists

**Does this PR require any upgrade changes?**:
no

**If the changes in this PR are manually verified, list down the scenarios covered:**:

**Any additional information for your reviewer?** :
_Mention if this PR is part of any design or a continuation of previous PRs_

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] PR Title follows the convention of  `<type>(<scope>): <subject>`
- [ ] Has the change log section been updated?
- [ ] Commit has unit tests
- [ ] Commit has integration tests
- [ ] (Optional) Are upgrade changes included in this PR? If not, mention the issue/PR to track:
- [ ] (Optional) If documentation changes are required, which issue on https://github.com/openebs/website is used to track them:

PR #613, opened 2025-01-21
fix(controller): prevent zfs volume cr deletion if snapshot exists

sinhaashish:
## Pull Request template

Upon pvc deletion the zfs snapshot also gets deleted. this can be verified using ssh into the node and running the command  zfs list -t all

**What this PR does?**:

**Checklist:**
- [X] Fixes #<issue number>
- [ ] Commit has unit tests
- [X] Commit has integration tests

### DIFF
--- a/pkg/builder/volbuilder/build.go
+++ b/pkg/builder/volbuilder/build.go
@@ -21,6 +21,9 @@ import (
 	apis "github.com/openebs/zfs-localpv/pkg/apis/openebs.io/zfs/v1"
 )
 
+// MarkForDeletionAnnotation is the annotation key
+const MarkForDeletionAnnotation string = "openebs.io/marked-for-deletion"
+
 // Builder is the builder object for ZFSVolume
 type Builder struct {
 	volume *ZFSVolume
@@ -151,6 +154,15 @@ func (b *Builder) WithRecordSize(rs string) *Builder {
 // WithVolBlockSize sets the volblocksize of ZFSVolume
 func (b *Builder) WithVolBlockSize(bs string) *Builder {
 	b.volume.Object.Spec.VolBlockSize = bs
+	return b
+}
+
+// WithAnnotation sets the annotation of ZFSVolume
+func (b *Builder) WithAnnotation() *Builder {
+	if b.volume.Object.Annotations == nil {
+		b.volume.Object.Annotations = make(map[string]string)
+	}
+	b.volume.Object.Annotations[MarkForDeletionAnnotation] = "true"
 	return b
 }
 
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -503,14 +503,14 @@ func (cs *controller) DeleteVolume(
 	defer unlock()
 
 	// verify if the volume has already been deleted
-	vol, err := zfs.GetVolume(volumeID)
+	vol, err := zfs.GetZFSVolume(volumeID)
 	if vol != nil && vol.DeletionTimestamp != nil {
-		goto deleteResponse
+		return csipayload.NewDeleteVolumeResponseBuilder().Build(), nil
 	}
 
 	if err != nil {
 		if k8serror.IsNotFound(err) {
-			goto deleteResponse
+			return csipayload.NewDeleteVolumeResponseBuilder().Build(), nil
 		}
 		return nil, errors.Wrapf(
 			err,
@@ -524,19 +524,45 @@ func (cs *controller) DeleteVolume(
 		return nil, status.Error(codes.Internal, "can not delete, volume creation is in progress")
 	}
 
-	// Delete the corresponding ZV CR
-	err = zfs.DeleteVolume(volumeID)
+	// Fetch the list of snapshot for the given volume
+	snapList, err := zfs.GetSnapshotForVolume(volumeID)
 	if err != nil {
-		return nil, errors.Wrapf(
-			err,
-			"failed to handle delete volume request for {%s}",
-			volumeID,
+		return nil, status.Errorf(
+			codes.NotFound,
+			"failed to handle delete volume request for {%s}, "+
+				"validation failed checking for snapshots. Error: %s",
+			req.VolumeId,
+			err.Error(),
 		)
+	}
+
+	// Delete the corresponding ZV CR only if there are no snapshots present for the volume
+	if len(snapList.Items) == 0 {
+		err = zfs.DeleteVolume(volumeID)
+		if err != nil {
+			return nil, errors.Wrapf(
+				err,
+				"failed to handle delete volume request for {%s}",
+				volumeID,
+			)
+		}
+	} else {
+		// add annotation to the volume to indicate that it is eligible for deletion
+		// once all the snapshots are deleted and the reclaim policy is not Retain
+		// this volume will be deleted
+		err = zfs.MarkForDeletion(volumeID)
+		if err != nil {
+			return nil, errors.Wrapf(
+				err,
+				"failed to annotate volume on deletion request for {%s}",
+				volumeID,
+			)
+		}
+
 	}
 
 	sendEventOrIgnore("", volumeID, vol.Spec.Capacity, analytics.VolumeDeprovision)
 
-deleteResponse:
 	return csipayload.NewDeleteVolumeResponseBuilder().Build(), nil
 }
 
@@ -815,8 +841,36 @@ func (cs *controller) DeleteSnapshot(
 		// should succeed when an invalid snapshot id is used
 		return &csi.DeleteSnapshotResponse{}, nil
 	}
+	volumeID := snapshotID[0]
 	unlock := cs.volumeLock.LockVolumeWithSnapshot(snapshotID[0], snapshotID[1])
 	defer unlock()
+
+	// verify if the snapshot has already been deleted
+	_, err := zfs.GetZFSSnapshot(snapshotID[1])
+
+	if err != nil {
+		if k8serror.IsNotFound(err) {
+			return csipayload.NewDeleteSnapshotResponseBuilder().Build(), nil
+		}
+		return nil, errors.Wrapf(
+			err,
+			"failed to get snapshot for {%s}",
+			snapshotID[1],
+		)
+	}
+
+	// Fetch the list of snapshot for the given volume
+	snapList, err := zfs.GetSnapshotForVolume(volumeID)
+	if err != nil {
+		return nil, status.Errorf(
+			codes.FailedPrecondition,
+			"failed to handle delete snapshot request for {%s}, "+
+				"validation failed checking for snapshot list for volume. Error: %s",
+			volumeID,
+			err.Error(),
+		)
+	}
+
 	if err := zfs.DeleteSnapshot(snapshotID[1]); err != nil {
 		return nil, status.Errorf(
 			codes.Internal,
@@ -825,7 +879,34 @@ func (cs *controller) DeleteSnapshot(
 			err.Error(),
 		)
 	}
-	return &csi.DeleteSnapshotResponse{}, nil
+
+	eligibleForDeletion, err := zfs.IsVolumeEligibleForDeletion(volumeID)
+	if err != nil {
+		return nil, status.Errorf(
+			codes.FailedPrecondition,
+			"failed to handle delete snapshot request for {%s}, "+
+				"validation failed checking for eligible for deletion. Error: %s",
+			volumeID,
+			err.Error(),
+		)
+	}
+
+	klog.Infof(" The snap list size %v and eligibleForDeletion %v", len(snapList.Items), eligibleForDeletion)
+	// Delete the corresponding ZV CR only if this is the last snapshot
+	// for the volume and the corresponding pvc is deleted
+	if len(snapList.Items) == 1 && eligibleForDeletion {
+		err = zfs.DeleteVolume(volumeID)
+		if err != nil {
+			return nil, errors.Wrapf(
+				err,
+				"failed to handle delete volume request for {%s}",
+				volumeID,
+			)
+		}
+		klog.Infof("volume %s deleted after the deletion of last snapshot %s ", volumeID, snapshotID[1])
+	}
+
+	return csipayload.NewDeleteSnapshotResponseBuilder().Build(), nil
 }
 
 // ListSnapshots lists all snapshots for the
--- a/pkg/zfs/volume.go
+++ b/pkg/zfs/volume.go
@@ -27,6 +27,8 @@ import (
 	"github.com/openebs/zfs-localpv/pkg/builder/restorebuilder"
 	"github.com/openebs/zfs-localpv/pkg/builder/snapbuilder"
 	"github.com/openebs/zfs-localpv/pkg/builder/volbuilder"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/klog/v2"
 )
@@ -213,13 +215,6 @@ func DeleteSnapshot(snapname string) (err error) {
 	return
 }
 
-// GetVolume the corresponding ZFSVolume CR
-func GetVolume(volumeID string) (*apis.ZFSVolume, error) {
-	return volbuilder.NewKubeclient().
-		WithNamespace(OpenEBSNamespace).
-		Get(volumeID, metav1.GetOptions{})
-}
-
 // DeleteVolume deletes the corresponding ZFSVol CR
 func DeleteVolume(volumeID string) (err error) {
 	err = volbuilder.NewKubeclient().WithNamespace(OpenEBSNamespace).Delete(volumeID)
@@ -249,6 +244,18 @@ func GetZFSVolume(volumeID string) (*apis.ZFSVolume, error) {
 	vol, err := volbuilder.NewKubeclient().
 		WithNamespace(OpenEBSNamespace).Get(volumeID, getOptions)
 	return vol, err
+}
+
+// UpdateZFSVolumeAnnotation updtates the ZFSVolume CR with the given annotation
+func UpdateZFSVolumeAnnotation(vol *apis.ZFSVolume) error {
+	newVol, err := volbuilder.BuildFrom(vol).
+		WithAnnotation().Build()
+
+	if err != nil {
+		return err
+	}
+	_, err = volbuilder.NewKubeclient().WithNamespace(OpenEBSNamespace).Update(newVol)
+	return err
 }
 
 // GetZFSVolumeState returns ZFSVolume OwnerNode and State for
@@ -440,4 +447,47 @@ func IsVolumeReady(vol *apis.ZFSVolume) bool {
 	}
 
 	return false
+}
+
+// GetSnapshotForVolume fetches all the snapshots for the given volume
+func GetSnapshotForVolume(volumeID string) (*apis.ZFSSnapshotList, error) {
+	listOptions := metav1.ListOptions{
+		LabelSelector: ZFSVolKey + "=" + volumeID,
+	}
+	snapList, err := snapbuilder.NewKubeclient().WithNamespace(OpenEBSNamespace).List(listOptions)
+	return snapList, err
+}
+
+// MarkForDeletion marks the volume for deletion by adding the annotation
+func MarkForDeletion(volumeName string) error {
+	zv, err := GetZFSVolume(volumeName)
+	if err != nil {
+		klog.Errorf("failed to get ZV %s: %v", volumeName, err)
+		return err
+	}
+
+	err = UpdateZFSVolumeAnnotation(zv)
+	if err != nil {
+		klog.Errorf("Failed to annotate the ZV with marked for deletion %s: %v", volumeName, err)
+		return err
+	}
+	return nil
+}
+
+// IsVolumeEligibleForDeletion checks if the volume can be deleted or not
+func IsVolumeEligibleForDeletion(volumeName string) (bool, error) {
+
+	zfsVol, err := GetZFSVolume(volumeName)
+	if err != nil {
+		return false, status.Errorf(
+			codes.Internal,
+			"failed to get ZFSVolume %s: %v",
+			volumeName,
+			err,
+		)
+	}
+	if zfsVol.Annotations[volbuilder.MarkForDeletionAnnotation] == "true" {
+		return true, nil
+	}
+	return false, nil
 }
--- a/tests/provision_test.go
+++ b/tests/provision_test.go
@@ -41,7 +41,7 @@ func exhaustiveVolumeTests(parameters map[string]string) {
 	snapshotAndCloneCreate()
 	// btrfs does not support online resize
 	if fstype != "btrfs" {
-		By("Resizing the PVC", resizeAndVerifyPVC)
+		By("Resizing the PVC", func() { resizeAndVerifyPVC(pvcNameFS) })
 	}
 	snapshotAndCloneCleanUp()
 	cleanUp()
@@ -51,8 +51,8 @@ func exhaustiveVolumeTests(parameters map[string]string) {
 func create(parameters map[string]string) {
 	By("####### Creating the storage class : " + parameters["fstype"] + " #######")
 	createFstypeStorageClass(parameters)
-	By("creating and verifying PVC bound status", createAndVerifyPVC)
-	By("Creating and deploying app pod", createDeployVerifyApp)
+	By("creating and verifying PVC bound status", func() { createAndVerifyPVC(pvcNameFS) })
+	By("Creating and deploying app pod", func() { createDeployVerifyApp(appNameFS, pvcNameFS) })
 	By("verifying ZFSVolume object", VerifyZFSVolume)
 	By("verifying storage class parameters")
 	VerifyStorageClassParams(parameters)
@@ -60,53 +60,96 @@ func create(parameters map[string]string) {
 
 // Creates the snapshot/clone resources
 func snapshotAndCloneCreate() {
-	createSnapshot(pvcName, snapName)
-	verifySnapshotCreated(snapName)
-	createClone(clonePvcName, snapName, scObj.Name)
-	By("Creating and deploying clone app pod", createDeployVerifyCloneApp)
+	createSnapshot(pvcNameFS, snapNameFS)
+	verifySnapshotCreated(snapNameFS)
+	createClone(clonePvcNameFS, snapNameFS, scObj.Name)
+	By("Creating and deploying clone app pod", func() { createDeployVerifyCloneApp(cloneAppNameFS, clonePvcNameFS) })
 }
 
 // Removes the snapshot/clone resources
 func snapshotAndCloneCleanUp() {
-	deleteAppDeployment(cloneAppName)
-	deletePVC(clonePvcName)
-	deleteSnapshot(pvcName, snapName)
+	deleteAppDeployment(cloneAppNameFS)
+	deletePVC(clonePvcNameFS)
+	deleteSnapshot(pvcNameFS, snapNameFS)
 }
 
 // Removes the resources
 func cleanUp() {
-	deleteAppDeployment(appName)
-	deletePVC(pvcName)
+	deleteAppDeployment(appNameFS)
+	deletePVC(pvcNameFS)
 	By("Deleting storage class", deleteStorageClass)
 }
 
 func blockVolCreationTest() {
 	By("Creating default storage class", createStorageClass)
-	By("creating and verifying PVC bound status", createAndVerifyBlockPVC)
+	By("creating and verifying PVC bound status", func() { createAndVerifyPVC(pvcNameBlock) })
 
-	By("Creating and deploying app pod", createDeployVerifyBlockApp)
+	By("Creating and deploying app pod", func() { createDeployVerifyApp(appNameBlock, pvcNameBlock) })
 	By("verifying ZFSVolume object", VerifyZFSVolume)
 	By("verifying ZFSVolume property change", VerifyZFSVolumePropEdit)
-	By("Deleting application deployment")
 
-	createSnapshot(pvcName, snapName)
-	verifySnapshotCreated(snapName)
-	createClone(clonePvcName, snapName, scObj.Name)
-	By("Creating and deploying clone app pod", createDeployVerifyCloneApp)
+	createSnapshot(pvcNameBlock, snapNameBlock)
+	verifySnapshotCreated(snapNameBlock)
+	createClone(clonePvcNameBlock, snapNameBlock, scObj.Name)
+	By("Creating and deploying clone app pod", func() { createDeployVerifyCloneApp(cloneAppNameBlock, clonePvcNameBlock) })
 
-	By("Deleting clone and main application deployment")
-	deleteAppDeployment(cloneAppName)
-	deleteAppDeployment(appName)
+	By("Deleting main application deployment")
+	deleteAppDeployment(appNameBlock)
 
-	By("Deleting snapshot, main pvc and clone pvc")
-	deletePVC(clonePvcName)
-	deleteSnapshot(pvcName, snapName)
-	deletePVC(pvcName)
+	zvName := getZVName(pvcNameBlock)
+	By("Deleting main pvc")
+	deletePVC(pvcNameBlock)
+
+	By("Verifying ZFSVolume object after pvc deletion when snapshot is present", VerifyZFSVolume)
+
+	By("Deleting clone application deployment")
+	deleteAppDeployment(cloneAppNameBlock)
+
+	By("Deleting snapshot and clone pvc")
+
+	deletePVC(clonePvcNameBlock)
+	By("Verifying that ZV is present after pvc deletion ", func() { IsZVPresentConsistently(zvName) })
+	deleteSnapshot(pvcNameBlock, snapNameBlock)
+	By("Verifying that ZV is deleted after snapshot deletion ", func() { IsZVDeletedEventually(zvName) })
 
 	By("Deleting storage class", deleteStorageClass)
+}
+
+func blockVolCreationWithReclaimRetainTest() {
+	By("Creating storage class retain reclaim policy", createStorageClassWithReclaimPolicy)
+	By("creating and verifying PVC bound status", func() { createAndVerifyPVC(pvcNameBlock) })
+
+	By("verifying ZFSVolume object", VerifyZFSVolume)
+
+	createSnapshot(pvcNameBlock, snapNameBlock)
+	verifySnapshotCreated(snapNameBlock)
+
+	zvName := getZVName(pvcNameBlock)
+	By("Deleting main pvc", func() { deletePVC(pvcNameBlock) })
+
+	By("Verifying ZFSVolume object after PVC deletion when snapshot is present", VerifyZFSVolume)
+
+	By("Verifying that ZV is present after PVC deletion ", func() { IsZVPresentConsistently(zvName) })
+	By("Deleting snapshot", func() { deleteSnapshot(pvcNameBlock, snapNameBlock) })
+	By("Verifying that ZV is present after snapshot deletion ", func() { IsZVPresentConsistently(zvName) })
+	deletePV(zvName)
+	By("Verifying that ZV is present after PV deletion ", func() { IsZVPresentConsistently(zvName) })
+	By("Create and Verifying PV from the retained ZV ", func() {
+		createAndVerifyPVFromRetainedZV(pvFromRetainZV, zvName)
+	})
+	By("creating and verifying PVC bound status from retained ZV", func() { createAndVerifyPVC(pvcFromRetainZV) })
+
+	zvNewName := getZVName(pvcFromRetainZV)
+	deletePVC(pvcFromRetainZV)
+	deletePV(zvNewName)
+	By("Deleting the ZV for cleanup ", func() { DeleteZV(zvName) })
+	By("Deleting storage class", deleteStorageClass)
+
 }
 
 func volumeCreationTest() {
 	By("Running volume creation test", fsVolCreationTest)
 	By("Running block volume creation test", blockVolCreationTest)
+	By("Running block volume creation test with retain reclaim policy", blockVolCreationWithReclaimRetainTest)
+
 }
--- a/tests/suite_test.go
+++ b/tests/suite_test.go
@@ -25,6 +25,7 @@ import (
 	"github.com/openebs/zfs-localpv/pkg/builder/volbuilder"
 	"github.com/openebs/zfs-localpv/tests/deploy"
 	"github.com/openebs/zfs-localpv/tests/pod"
+	"github.com/openebs/zfs-localpv/tests/pv"
 	"github.com/openebs/zfs-localpv/tests/pvc"
 	"github.com/openebs/zfs-localpv/tests/sc"
 	appsv1 "k8s.io/api/apps/v1"
@@ -41,18 +42,33 @@ const (
 )
 
 var (
-	ZFSClient      *volbuilder.Kubeclient
-	SCClient       *sc.Kubeclient
-	PVCClient      *pvc.Kubeclient
-	DeployClient   *deploy.Kubeclient
-	PodClient      *pod.KubeClient
-	scName         = "zfspv-sc"
-	ZFSProvisioner = "zfs.csi.openebs.io"
-	pvcName        = "zfspv-pvc"
-	snapName       = "zfspv-snap"
-	appName        = "busybox-zfspv"
-	clonePvcName   = "zfspv-pvc-clone"
-	cloneAppName   = "busybox-zfspv-clone"
+	ZFSClient           *volbuilder.Kubeclient
+	SCClient            *sc.Kubeclient
+	PVCClient           *pvc.Kubeclient
+	PVClient            *pv.Kubeclient
+	DeployClient        *deploy.Kubeclient
+	PodClient           *pod.KubeClient
+	scName              = "zfspv-sc"
+	ZFSProvisioner      = "zfs.csi.openebs.io"
+	RetainReclaimPolicy = corev1.PersistentVolumeReclaimPolicy(corev1.PersistentVolumeReclaimRetain)
+
+	pvcNameFS    = "zfspv-pvc-fs"
+	pvcNameBlock = "zfspv-pvc-block"
+
+	appNameFS    = "busybox-zfspv-fs"
+	appNameBlock = "busybox-zfspv-block"
+
+	snapNameFS    = "zfspv-snap-fs"
+	snapNameBlock = "zfspv-snap-block"
+
+	clonePvcNameFS    = "zfspv-pvc-clone-fs"
+	clonePvcNameBlock = "zfspv-pvc-clone-block"
+
+	cloneAppNameFS    = "busybox-zfspv-clone-fs"
+	cloneAppNameBlock = "busybox-zfspv-clone-block"
+
+	pvFromRetainZV  = "pv-from-retain-zv"
+	pvcFromRetainZV = "pvc-from-retain-zv"
 
 	scObj              *storagev1.StorageClass
 	deployObj          *appsv1.Deployment
@@ -76,6 +92,7 @@ func init() {
 	}
 	SCClient = sc.NewKubeClient(sc.WithKubeConfigPath(KubeConfigPath))
 	PVCClient = pvc.NewKubeClient(pvc.WithKubeConfigPath(KubeConfigPath))
+	PVClient = pv.NewKubeClient(pv.WithKubeConfigPath(KubeConfigPath))
 	DeployClient = deploy.NewKubeClient(deploy.WithKubeConfigPath(KubeConfigPath))
 	PodClient = pod.NewKubeClient(pod.WithKubeConfigPath(KubeConfigPath))
 	ZFSClient = volbuilder.NewKubeclient(volbuilder.WithKubeConfigPath(KubeConfigPath))
--- a/pkg/response/delete.go
+++ b/pkg/response/delete.go
@@ -39,3 +39,23 @@ func NewDeleteVolumeResponseBuilder() *DeleteVolumeResponseBuilder {
 func (b *DeleteVolumeResponseBuilder) Build() *csi.DeleteVolumeResponse {
 	return b.response
 }
+
+// DeleteSnapshotResponseBuilder helps building an
+// instance of csi DeleteSnapshotResponse
+type DeleteSnapshotResponseBuilder struct {
+	response *csi.DeleteSnapshotResponse
+}
+
+// NewDeleteSnapshotResponseBuilder returns a new
+// instance of DeleteSnapshotResponseBuilder
+func NewDeleteSnapshotResponseBuilder() *DeleteSnapshotResponseBuilder {
+	return &DeleteSnapshotResponseBuilder{
+		response: &csi.DeleteSnapshotResponse{},
+	}
+}
+
+// Build returns the constructed instance
+// of csi DeleteSnapshotResponse
+func (b *DeleteSnapshotResponseBuilder) Build() *csi.DeleteSnapshotResponse {
+	return b.response
+}
--- a/tests/pv/build.go
+++ b/tests/pv/build.go
@@ -1,0 +1,80 @@
+package pv
+
+import (
+	"github.com/openebs/lib-csi/pkg/common/errors"
+	corev1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/resource"
+)
+
+// PV is a wrapper over persistentvolume api
+type pvBuildOption func(*PV)
+
+// PV is a wrapper over persistentvolume api
+// object. It provides build, validations and other common
+// logic to be used by various feature specific callers.
+type PV struct {
+	object *corev1.PersistentVolume
+}
+
+// Builder is the builder object for PV
+type Builder struct {
+	pv   *PV
+	errs []error
+}
+
+// NewBuilder returns new instance of Builder
+func NewBuilder() *Builder {
+	return &Builder{pv: &PV{object: &corev1.PersistentVolume{}}}
+}
+
+// WithName sets the Name field of PV with provided value.
+func (b *Builder) WithName(name string) *Builder {
+	if len(name) == 0 {
+		b.errs = append(b.errs, errors.New("failed to build PV object: missing PV name"))
+		return b
+	}
+	b.pv.object.Name = name
+	return b
+}
+
+// WithStorageClass sets the StorageClass field of PV with provided arguments
+func (b *Builder) WithStorageClass(scName string) *Builder {
+	if len(scName) == 0 {
+		b.errs = append(b.errs, errors.New("failed to build PV object: missing storageclass name"))
+		return b
+	}
+	b.pv.object.Spec.StorageClassName = scName
+	return b
+}
+
+// WithCapacity sets the Capacity field in PV with provided arguments
+func (b *Builder) WithCapacity(capacity string) *Builder {
+	resCapacity, err := resource.ParseQuantity(capacity)
+	if err != nil {
+		b.errs = append(b.errs, errors.Wrapf(err, "failed to build PV object: failed to parse capacity {%s}", capacity))
+		return b
+	}
+	resourceList := corev1.ResourceList{
+		corev1.ResourceName(corev1.ResourceStorage): resCapacity,
+	}
+	b.pv.object.Spec.Capacity = resourceList
+	return b
+}
+
+// WithAccessModes sets the AccessMode field in PV with provided arguments
+func (b *Builder) WithAccessModes(accessMode []corev1.PersistentVolumeAccessMode) *Builder {
+	if len(accessMode) == 0 {
+		b.errs = append(b.errs, errors.New("failed to build PV object: missing accessmodes"))
+		return b
+	}
+	b.pv.object.Spec.AccessModes = accessMode
+	return b
+}
+
+// Build returns the PV API instance
+func (b *Builder) Build() (*corev1.PersistentVolume, error) {
+	if len(b.errs) > 0 {
+		return nil, errors.Errorf("%+v", b.errs)
+	}
+	return b.pv.object, nil
+}
--- a/tests/pv/kubernetes.go
+++ b/tests/pv/kubernetes.go
@@ -1,0 +1,178 @@
+package pv
+
+import (
+	"context"
+	"strings"
+
+	"github.com/openebs/lib-csi/pkg/common/errors"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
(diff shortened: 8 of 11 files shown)


sinhaashish:
@tiagolobocastro  PR https://github.com/openebs/zfs-localpv/pull/607 is closed as there was some issue from my side while rebasing. I have created this PR with the same code base along with checking that ZV is present in the system  as per your comment [here](https://github.com/openebs/zfs-localpv/pull/607#discussion_r1920284951).

Splitting the test to run on different cluster can be taken sometime later.

ci/ci-test.sh
```diff
@@ -107,7 +107,7 @@ runTestSuite() {
 
   echo "running ginkgo test case with coverage ${coverageFile}"
 
-  if ! ginkgo -v -coverprofile="${coverageFile}" --label-filter="${labelFilter}" -covermode=atomic; then
+  if ! ginkgo -p -v -coverprofile="${coverageFile}" --label-filter="${labelFilter}" -covermode=atomic; then
```
tiagolobocastro: Is this running it in parallel?
I'd rather run in parallel against the same cluster as can potential cause unpredictable behavior between running tests.
Maybe we can leave it entirely to another piece of work

tests/zv/build.go
```diff
@@ -0,0 +1,40 @@
+/*
+Copyright 2019-2025 The OpenEBS Authors
```
tiagolobocastro: 2019? This is a new file
Also now we don't need to and shouldn't add headers IIRC

tests/zv/build.go
tiagolobocastro, line 1: Don't we already have a client for this?

We have this kind of code in test already
```go
vol, err := ZFSClient.WithNamespace(OpenEBSNamespace).
		Get(pvcObj.Spec.VolumeName, metav1.GetOptions{})
```

tests/provision_test.go
```diff
@@ -51,57 +52,83 @@ func exhaustiveVolumeTests(parameters map[string]string) {
...
+	By("verifying ZFSVolume object", VerifyZFSVolume)
+	createSnapshot(pvcNameAplha, snapNameAlpha)
+	verifySnapshotCreated(snapNameAlpha)
+	By("Deleting main application deployment")
+	deleteAppDeployment(appNameAlpha)
+	zvName := getZVName(pvcNameAplha)
+	isZVPresent(zvName)
+	By("Deleting main pvc")
+	deletePVC(pvcNameAplha)
+	By("Verifying ZFSVolume object after pvc deletion when snapshot is present", VerifyZFSVolume)
```
tiagolobocastro, line 103: Should we use a consistently to ensure the zfs volume is not deleted within a time period?

tests/suite_test.go
```diff
@@ -46,13 +47,36 @@ var (
...
+	clonePvcName = "zfspv-pvc-clone"
+	cloneAppName = "busybox-zfspv-clone"
+
+	pvcNameFS    = "zfspv-pvc-fs"
+	pvcNameBlock = "zfspv-pvc-block"
+	pvcNameAplha = "zfspv-pvc-alpha"
+
+	appNameFS    = "busybox-zfspv-fs"
+	appNameBlock = "busybox-zfspv-block"
+	appNameAlpha = "busybox-zfspv-alpha"
```
tiagolobocastro: what's alpha?

sinhaashish: its  just a suffix to make the name distinct

tests/provision_test.go
sinhaashish: done , added `isZVPresentEventually`

tests/zv/build.go
sinhaashish: removed

ci/ci-test.sh
sinhaashish: yes for this PR we can stick to volume and snapshot functioning

tiagolobocastro: I meant, I'd rather **not** run in parallel on the same cluster as can potential cause unpredictable behavior between running tests if they step on each others toes.

tests/provision_test.go
tiagolobocastro: consistently, not eventually
https://onsi.github.io/gomega/#consistently

tests/suite_test.go
tiagolobocastro: Why does it need to be different? Shouldn't blockVolume tests using the blockVol pvc naming?

ci/ci-test.sh
sinhaashish: removed `-p`

tests/provision_test.go
sinhaashish: done

tests/suite_test.go
sinhaashish: since the test was running in parallel (on same cluster)  there were name conflicts. Thats why i used alpha.  reverted it now.

tests/provision_test.go
```diff
@@ -51,57 +52,83 @@ func exhaustiveVolumeTests(parameters map[string]string) {
...
-	deletePVC(pvcName)
+
+	deletePVC(clonePvcNameBlock)
+	deleteSnapshot(pvcNameBlock, snapNameBlock)
+	deletePVC(pvcNameBlock)
+
+	By("Deleting storage class", deleteStorageClass)
+}
+
+func blockVolDeletionTest() {
```
tiagolobocastro: Do we need this separate test, can't it be combined with the existing `blockVolCreationTest` ?

sinhaashish: Merged it into a single test

tests/provision_test.go
```diff
@@ -51,57 +51,66 @@ func exhaustiveVolumeTests(parameters map[string]string) {
...
-	deleteAppDeployment(appName)
+	By("Deleting main application deployment")
+	deleteAppDeployment(appNameBlock)
 
-	By("Deleting snapshot, main pvc and clone pvc")
-	deletePVC(clonePvcName)
-	deleteSnapshot(pvcName, snapName)
-	deletePVC(pvcName)
+	zvName := getZVName(pvcNameBlock)
+	isZVPresentConsistently(zvName)
```
tiagolobocastro: We haven't deleted the volume yet, so don't need to check it here?

tests/provision_test.go
```diff
@@ -51,57 +51,66 @@ func exhaustiveVolumeTests(parameters map[string]string) {
...
+	By("Verifying ZFSVolume object after pvc deletion when snapshot is present", VerifyZFSVolume)
+
+	By("Deleting clone application deployment")
+	deleteAppDeployment(cloneAppNameBlock)
+
+	By("Deleting snapshot and clone pvc")
+
+	deletePVC(clonePvcNameBlock)
+	By("Verifying that ZV is present after pvc deletion ", func() { isZVPresentConsistently(zvName) })
+	deleteSnapshot(pvcNameBlock, snapNameBlock)
```
tiagolobocastro, line 112: Shouldn't we check volume is deleted here?

tests/provision_test.go
sinhaashish: removed

tests/provision_test.go
sinhaashish: added `IsZVDeletedEventually` to verify that the volume is deleted after the deletion of snapshot

tiagolobocastro: hmm but you're deleting the ZV in the test? that should be done automatically?

sinhaashish: Modified it , now it deletes automatically .

pkg/driver/controller.go
```diff
@@ -502,6 +505,18 @@ func (cs *controller) DeleteVolume(
 	unlock := cs.volumeLock.LockVolume(volumeID)
 	defer unlock()
 
+	// Fetch the list of snapshot for the given volume
```
avishnu: The below code block should ideally be placed before line 542.

sinhaashish: Here I am declaring a new variable `snapList` , if i put it at line 542 , it error out stating `goto deleteResponse jumps over variable declaration`. 

`deleteResponse` is used to check if the volume is already deleted.

avishnu: Seems to be a Go limitation. Not sure if it's addressed with this issue: https://github.com/golang/go/issues/27165. There are ways around this, eg. declare the variable on the top.

The current logic seems incorrect.

sinhaashish: Declared the variable at top and initialised it later, when used.

tiagolobocastro: Why do we need to declare it at the top?
Can't we just do snapList, err := ..... ?

sinhaashish: no, there is a `goto deleteResponse` which checks if volume is present . If we do `:=`  for snapList then its errors saying `goto deleteResponse jumps over variable declaration.`

tiagolobocastro: Let's not use `goto`, it's generally seen as a bad practice nowadays.

sinhaashish: removed `goto`

pkg/driver/controller.go
```diff
@@ -1128,3 +1198,44 @@ func LabelIndexFunc(label string) cache.IndexFunc {
...
+		return false, status.Errorf(
+			codes.Internal,
+			"failed to build k8s clientset: %v",
+			err,
+		)
+	}
+	isPVCPresent := true
+
+	pv, err := clientset.CoreV1().PersistentVolumes().Get(context.TODO(), volumeName, metav1.GetOptions{})
+	// If the PV is not found, or the pv does not have the calim referennce
```
tiagolobocastro:
```suggestion
	// If the PV is not found, or the PV does not have the claim reference
```

pkg/driver/controller.go
```diff
@@ -825,7 +881,21 @@ func (cs *controller) DeleteSnapshot(
 			err.Error(),
 		)
 	}
-	return &csi.DeleteSnapshotResponse{}, nil
+
+	// Delete the corresponding ZV CR only if this is the last snapshot
+	// for the volume and the corresponding pvc is deleted
+	if len(snapList.Items) == 1 && !isCorrespondingPvcPresent {
+		err = zfs.DeleteVolume(volumeID)
```
tiagolobocastro, line 898: hmm I think this might inadvertently delete a volume with the retain binding mode where a user is allowed to delete the PVC and PV but the store asset should remain.
I think on the `(cs *controller) DeleteVolume` we need to mark it there as deleted, which would allow us here to proceed with the deletion? Or something along those lines?

sinhaashish: The flow doesnt come to `(cs *controller) DeleteVolume` in case the storage class has `reclaimPolicy: Retain`. So we cant do any thing there . 

On the other hand i check for pv for `persistentVolumeReclaimPolicy` . If its value is `Retain` then i don't delete the zv .  
added the function `isVolumeEligibleForDeletion`

tiagolobocastro: > The flow doesnt come to (cs *controller) DeleteVolume in case the storage class has reclaimPolicy: Retain. 

Yes exactly, that's the idea.
When the `reclaimPolicy` is `Retain` we don't receive the `DeleteVolume` and therefore wouldn't mark the `ZV` for deletion. So when the last snapshot is deleted, we don't delete the the `ZV`.

Otherwise, if we do receive the `DeleteVolume` then when the last snapshot is deleted, we do delete the `ZV`.

Would this workflow make sense @Abhinandan-Purkait ? What's the best way to mark a deletion on the `ZV` without actually adding the `deletionTimestamp`? (that would trigger a storm of reconciles)

Abhinandan-Purkait: Yes this workflow sounds good to me. We can use annotations to mark things. I will try to summarize the workflow:

1. If We received a DeleteVolume call, and volume has snapshots, add `marked for deletetion` annotation on ZV and succeed the DeleteVolume call with no op.
2. If We received a DeleteVolume call, and volume doesn't have any snapshots, succeed the DeleteVolume call with and also remove the assets.
3. If we never received a DeleteVolume call, the annotation will never be present so, we just deal with deletion of snapshots in DeleteSnapshot and ignore ZV.

It still might be a good idea to fetch the PV if present in `isVolumeEligibleForDeletion` and double check the reclaim policy just in case :-;

tiagolobocastro: ~~Looks good, one thing missing: when we receive a `CreateVolume` call, and the `ZV` exists, we have to remove this annonation. This completes the circle.
We should have a test for this.~~

tiagolobocastro: Actually, I'm not sure if we get another `CreateVolume` when the `PV` is re-created. Maybe we don't!

But maybe this is not an issue, because if we had received the `DeleteCall` then I don't think the user would be able to create the `PV` and have it linked with the `ZV` or would he?

pkg/driver/controller.go
```diff
@@ -1128,3 +1198,51 @@ func LabelIndexFunc(label string) cache.IndexFunc {
...
+	// then the ZV is not eligible for deletion
+	if pv.Spec.PersistentVolumeReclaimPolicy == "Retain" {
+		return false, nil
+	}
+
+	// Check if the PVC exists
+	_, err = clientset.CoreV1().PersistentVolumeClaims(pv.Spec.ClaimRef.Namespace).Get(context.TODO(), pv.Spec.ClaimRef.Name, metav1.GetOptions{})
+	// if error then pvc does not exist
+	// and thus the ZV is eligible for deletion
+	if err != nil {
```
dsharma-dc: Shall we match a specific error code before assuming pvc doesn't exist?

pkg/driver/controller.go
```diff
@@ -1128,3 +1198,51 @@ func LabelIndexFunc(label string) cache.IndexFunc {
...
+		)
+	}
+
+	pv, err := clientset.CoreV1().PersistentVolumes().Get(context.TODO(), volumeName, metav1.GetOptions{})
+	// If the PV is not found, then assuming the PVC is not present
+	// and thus the ZV is eligible for deletion
+	if err != nil {
+		return true, nil
+	}
+	// If the PV exist and does not have the calim reference
```
dsharma-dc:
```suggestion
	// If the PV exist and does not have the claim reference
```

pkg/driver/controller.go
sinhaashish: there is a CreateVolume call when the pv is recreated . 


```
ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get sc
NAME                       PROVISIONER          RECLAIMPOLICY   VOLUMEBINDINGMODE      ALLOWVOLUMEEXPANSION   AGE
hcloud-volumes (default)   csi.hetzner.cloud    Delete          WaitForFirstConsumer   true                   19h
openebs-zfspv              zfs.csi.openebs.io   Delete          Immediate              false                  18h
openebs-zfspv-retain       zfs.csi.openebs.io   Retain          Immediate              false                  18h

ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k apply -f pvc-retain.yaml
persistentvolumeclaim/csi-zfspv-retain created

ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get pvc
NAME               STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS           VOLUMEATTRIBUTES
apiVersion: v1
CLASS   AGE
csi-zfspv-retain   Bound    pvc-f56a88cc-3333-4689-ac81-905ad3a5ceda   4Gi        RWO            openebs-zfspv-retain   <unset>                 63s

ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get pv
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM                      STORAGECLASS           VOLUMEATTRIBUTESCLASS   REASON   AGE
pvc-f56a88cc-3333-4689-ac81-905ad3a5ceda   4Gi        RWO            Retain           Bound    default/csi-zfspv-retain   openebs-zfspv-retain   <unset>                          65s

ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get zv -n openebs
apiVersion: v1
NAME                                       ZPOOL        NODEID          SIZE         STATUS   FILESYSTEM   AGE
pvc-f56a88cc-3333-4689-ac81-905ad3a5ceda   zfspv-pool   node-0-309370   4294967296   Ready    zfs          73s

ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k delete pvc csi-zfspv-retain
persistentvolumeclaim "csi-zfspv-retain" deleted

ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k delete pv pvc-f56a88cc-3333-4689-ac81-905ad3a5ceda
persistentvolume "pvc-f56a88cc-3333-4689-ac81-905ad3a5ceda" deleted

ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get zv -n openebs
NAME                                       ZPOOL        NODEID          SIZE         STATUS   FILESYSTEM   AGE
pvc-f56a88cc-3333-4689-ac81-905ad3a5ceda   zfspv-pool   node-0-309370   4294967296   Ready    zfs          4m31s


ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ vi generate-pv.yaml
ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k apply -f generate-pv.yaml
persistentvolume/my-reclaimed-pv created

ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get pv
NAME              CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM                      STORAGECLASS           VOLUMEATTRIBUTESCLASS   REASON   AGE
my-reclaimed-pv   5Gi        RWO            Retain           Bound    default/my-reclaimed-pvc   openebs-zfspv-retain   <unset>                          10m

ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ vi generate-pvc.yaml
ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k apply -f generate-pvc.yaml
persistentvolumeclaim/my-reclaimed-pvc created

ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get pvc
NAME               STATUS   VOLUME            CAPACITY   ACCESS MODES   STORAGECLASS           VOLUMEATTRIBUTESCLASS   AGE
my-reclaimed-pvc   Bound    my-reclaimed-pv   5Gi        RWO            openebs-zfspv-retain   <unset>                 9s

```

The logs shows the call for create volume 
```
ashish@ubuntu-32gb-nbg1-2:~$ k logs openebs-zfs-localpv-controller-5899cc4798-wx8fk -n openebs -c openebs-zfs-plugin


I0130 05:52:45.925003       1 controller.go:294] zfs: trying volume creation zfspv-pool/pvc-f56a88cc-3333-4689-ac81-905ad3a5ceda on node [node-0-309370 node-1-309370 node-2-309370]
I0130 05:52:45.941036       1 volume.go:145] zfs: waiting for volume zfspv-pool/pvc-f56a88cc-3333-4689-ac81-905ad3a5ceda to be created on nodeid node-0-309370
I0130 05:52:46.944836       1 controller.go:474] created the volume zfspv-pool/pvc-f56a88cc-3333-4689-ac81-905ad3a5ceda on node node-0-309370
I0130 05:52:46.944914       1 grpc.go:81] GRPC response: {"volume":{"accessible_topology":[{"segments":{"openebs.io/nodeid":"node-0-309370"}}],"capacity_bytes":4294967296,"volume_context":{"openebs.io/cas-type":"localpv-zfs","openebs.io/poolname":"zfspv-pool"},"volume_id":"pvc-f56a88cc-3333-4689-ac81-905ad3a5ceda"}}
```

pkg/driver/controller.go
sinhaashish: now we dont check for PVC

pkg/builder/volbuilder/build.go
```diff
@@ -154,6 +154,15 @@ func (b *Builder) WithVolBlockSize(bs string) *Builder {
 	return b
 }
 
+// WithAnnotation sets the annotation of ZFSVolume
+func (b *Builder) WithAnnotation() *Builder {
+	if b.volume.Object.Annotations == nil {
+		b.volume.Object.Annotations = make(map[string]string)
+	}
+	b.volume.Object.Annotations["openebs.zv.isEligibleForDeletion"] = "true"
```
Abhinandan-Purkait: `openebs.io/marked-for-deletion`

pkg/driver/controller.go
```diff
@@ -524,19 +524,45 @@ func (cs *controller) DeleteVolume(
...
+		err = zfs.DeleteVolume(volumeID)
+		if err != nil {
+			return nil, errors.Wrapf(
+				err,
+				"failed to handle delete volume request for {%s}",
+				volumeID,
+			)
+		}
+	} else {
+		// add annotation to the volume to indicate that it is elegible for deletion
```
Abhinandan-Purkait: elegible --> eligible

pkg/driver/controller.go
```diff
@@ -524,19 +524,45 @@ func (cs *controller) DeleteVolume(
...
+		// once all the snapshots are deleted and the reclaim policy is not Retain
+		// this volume will be deleted
+		err = zfs.MarkForDeletion(volumeID)
+		if err != nil {
+			return nil, errors.Wrapf(
+				err,
+				"failed to handle delete volume request for {%s}",
```
Abhinandan-Purkait: failed to annotate volume on deletion request

pkg/driver/controller.go
```diff
@@ -815,8 +841,50 @@ func (cs *controller) DeleteSnapshot(
...
+		return nil, status.Errorf(
+			codes.NotFound,
+			"failed to handle delete snapshot request for {%s}, "+
+				"validation failed checking for snapshot list for volume. Error: %s",
+			volumeID,
+			err.Error(),
+		)
+	}
+
+	eligibeForDeltion, err := zfs.IsVolumeEligibleForDeletion(volumeID)
```
Abhinandan-Purkait: eligibeForDeltion --> eligibleForDeletion

pkg/driver/controller.go
```diff
@@ -815,8 +841,50 @@ func (cs *controller) DeleteSnapshot(
...
+
+	// Fetch the list of snapshot for the given volume
+	snapList, err := zfs.GetSnapshotForVolume(volumeID)
+	if err != nil {
+		return nil, status.Errorf(
+			codes.NotFound,
+			"failed to handle delete snapshot request for {%s}, "+
+				"validation failed checking for snapshot list for volume. Error: %s",
+			volumeID,
+			err.Error(),
```
Abhinandan-Purkait: This bit looks incorrect, if the we failed to get Snapshot for volume we seem to return not found for all error cases?

pkg/driver/controller.go
```diff
@@ -815,8 +841,50 @@ func (cs *controller) DeleteSnapshot(
...
+
+	eligibeForDeltion, err := zfs.IsVolumeEligibleForDeletion(volumeID)
+	if err != nil {
+		return nil, status.Errorf(
+			codes.NotFound,
+			"failed to handle delete snapshot request for {%s}, "+
+				"validation failed checking for eligine for deletion. Error: %s",
+			volumeID,
+			err.Error(),
+		)
```
Abhinandan-Purkait: Again this bit is not correct, if the volume is not eligible for deletion method returns error we always say not found?

pkg/driver/controller.go
```diff
@@ -825,7 +893,21 @@ func (cs *controller) DeleteSnapshot(
 			err.Error(),
 		)
 	}
-	return &csi.DeleteSnapshotResponse{}, nil
+
```
Abhinandan-Purkait, line 893: Why are we deleting the snapshot after all the checks? Shouldn't we first delete the snapshot and then check for the length of the snaplist? If snaplist is 0 then delete the volume? If there is any error in figuring out the snaplist after volume deletion, return an error so that it retries again and cleans up the volume.

pkg/driver/controller.go
```diff
@@ -825,7 +893,21 @@ func (cs *controller) DeleteSnapshot(
...
+		err = zfs.DeleteVolume(volumeID)
+		if err != nil {
+			return nil, errors.Wrapf(
+				err,
+				"failed to handle delete volume request for {%s}",
+				volumeID,
+			)
+		}
+	}
+
```
Abhinandan-Purkait: We also need an info that volume has been cleanup on last snapshot deletion. Please add that.

pkg/zfs/volume.go
```diff
@@ -251,6 +247,18 @@ func GetZFSVolume(volumeID string) (*apis.ZFSVolume, error) {
 	return vol, err
 }
 
+// UpdateZFSVolumeAnnotation updtates the ZFSVolume CR with the given annotation
+func UpdateZFSVolumeAnnotation(vol *apis.ZFSVolume) error {
+	newVol, err := volbuilder.BuildFrom(vol).
+		WithAnnotation().Build()
+
+	if err != nil {
+		return err
```
Abhinandan-Purkait, line 255: Wrap the error with failed to annotate error? So that there is a bit of context.

pkg/zfs/volume.go
```diff
@@ -251,6 +247,18 @@ func GetZFSVolume(volumeID string) (*apis.ZFSVolume, error) {
...
+	}
+	_, err = volbuilder.NewKubeclient().WithNamespace(OpenEBSNamespace).Update(newVol)
+	return err
```
Abhinandan-Purkait, line 258: return nil?

pkg/zfs/volume.go
```diff
@@ -441,3 +449,79 @@ func IsVolumeReady(vol *apis.ZFSVolume) bool {
...
+func MarkForDeletion(volumeName string) error {
+	zv, err := GetZFSVolume(volumeName)
+	if err != nil {
+		klog.Errorf("failed to get ZV %s: %v", volumeName, err)
+		return err
+	}
+
+	err = UpdateZFSVolumeAnnotation(zv)
+	if err != nil {
+		klog.Errorf("failed to update ZV %s: %v", volumeName, err)
```
Abhinandan-Purkait: Failed to update --> Failed to annotate the ZV with marked for deletion

pkg/builder/volbuilder/build.go
sinhaashish: done

pkg/zfs/volume.go
```diff
@@ -441,3 +449,79 @@ func IsVolumeReady(vol *apis.ZFSVolume) bool {
...
+	clientset, err := kubernetes.NewForConfig(cfg)
+	if err != nil {
+		return false, status.Errorf(
+			codes.Internal,
+			"failed to build k8s clientset: %v",
+			err,
+		)
+	}
+
+	pv, err := clientset.CoreV1().PersistentVolumes().Get(context.TODO(), volumeName, metav1.GetOptions{})
```
Abhinandan-Purkait: IIRC, you can avoid all this by just using the k8sapi generic resource getter. I think we are meant to use those utilities rather than getting creating the clientset from config manually everytime.

For example:- 

```
	pv, err := k8sapi.GetResource(schema.GroupVersionResource{
		Group: "",
		Version: "v1",
		Resource: "persistentvolumes",
	}, "").Get(context.TODO(), volumeName, metav1.GetOptions{})
```
Just this line should do it.

pkg/zfs/volume.go
```diff
@@ -441,3 +449,79 @@ func IsVolumeReady(vol *apis.ZFSVolume) bool {
...
+		// If the PV exist and does not have the claim reference
+		// then assuming the PVC is not present and thus the ZV is eligible for deletion
+		if pv.Spec.ClaimRef == nil {
+			return true, nil
+		}
+
+		// If the PV exist and has the reclaim policy as retain
+		// then the ZV is not eligible for deletion
+		if pv.Spec.PersistentVolumeReclaimPolicy == "Retain" {
+			return false, nil
```
Abhinandan-Purkait: Probably cleanup the annotation if it exists as well? cc @tiagolobocastro ?

pkg/zfs/volume.go
```diff
@@ -441,3 +449,79 @@ func IsVolumeReady(vol *apis.ZFSVolume) bool {
...
+	zfsVol, err := GetZFSVolume(volumeName)
+	if err != nil {
+		return false, status.Errorf(
+			codes.Internal,
+			"failed to get ZFSVolume %s: %v",
+			volumeName,
+			err,
+		)
+	}
+	if zfsVol.Annotations["openebs.io/marked-for-deletion"] == "true" {
```
Abhinandan-Purkait: Please define the `openebs.io/marked-for-deletion` from the constants above.

pkg/driver/controller.go
Abhinandan-Purkait: Okay in that case can you check, this scenario.

1. Create Retain PVC.
2. Create Snapshot.
3. Delete PVC.
4. Check for the PV to exist.
5. Delete the PV.
6. The PV should be gone and the ZV should stay around as there is the a snapshot, or probably because we never received a delete call.
7. Recreate the PV and PVC. 
8. Now delete the snapshot.
9. Does the ZV stay and have no deletion timestamp and no marked for deletion timestamp?

Abhinandan-Purkait: Also one more scenario.

1. Create Non-Retain PVC.
2. Create Snapshot.
3. Delete PVC.
4. Check for the PV to exist.
5. Delete the PV.
6. The PV should be gone and the ZV should stay around as there is the a snapshot, with marked for delete annotation.
7. Recreate the PV and PVC. 
8. Now delete the snapshot.
9. Does the ZV go away + the underlying ZFS zvol/dataset?

Abhinandan-Purkait: If the ZV goes away in the scenario 2, then we should remove marked for deletion annotation CreateVolume with existing ZV.

pkg/driver/controller.go
Abhinandan-Purkait: We are delaying the snapshot cleanup which might have just worked in the first place, by doing checks for volume cleanup. This will make the snapshot stay until the volume cleanup checks pass, which means it will keep consuming space. IMO we should let the snapshot be deleted if it can be, then retry on the volume cleanup if it errors out. This way atleast the user will have some space freed up. Any thoughts @tiagolobocastro @avishnu ?

pkg/zfs/volume.go
sinhaashish: will return nil if there is no error.

pkg/zfs/volume.go
sinhaashish: Its already handled in `MarkForDeletion` . so  IMHO no need to duplicate the info

pkg/zfs/volume.go
sinhaashish: Since we are creating a client for all , so i created. Following the same way so make code look uniform

pkg/driver/controller.go
sinhaashish: changed it with `FailedPrecondition`

pkg/driver/controller.go
sinhaashish: same here , changed it with `FailedPrecondition`

pkg/driver/controller.go
sinhaashish: > Why are we deleting the snapshot after all the checks? Shouldn't we first delete the snapshot and then check for the length of the snaplist? If snaplist is 0 then delete the volume? If there is any error in figuring out the snaplist after volume deletion, return an error so that it retries again and cleans up the volume.

Just to make a point here is that the validation are checked before the operation . 
Moreover upon deletion of last snapshot , the snapList size is still 1. it doesn't get reflected immediately . I tried by adding sleep 10 sec  then snapList becomes 0 . One solution to this is wait till ve verify that the last snapshot is deleted and snapList becomes 0 and then we proceed further . 
This actually increases the response time for the operation .

pkg/zfs/volume.go
sinhaashish: This return an unstructured object which we need to figure out from `Object map[string]interface{}` . Better to use the resource specific get call.

sinhaashish: anyway removed it

pkg/driver/controller.go
sinhaashish: > Okay in that case can you check, this scenario.
> 
> 1. Create Retain PVC.
> 2. Create Snapshot.
> 3. Delete PVC.
> 4. Check for the PV to exist.
> 5. Delete the PV.
> 6. The PV should be gone and the ZV should stay around as there is the a snapshot, or probably because we never received a delete call.
> 7. Recreate the PV and PVC.
> 8. Now delete the snapshot.
> 9. Does the ZV stay and have no deletion timestamp and no marked for deletion timestamp?


```
ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k apply -f pvc-retain.yaml
persistentvolumeclaim/csi-zfspv-retain created
ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get pvc
NAME               STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS           VOLUMEATTRIBUTESCLASS   AGE
csi-zfspv-retain   Bound    pvc-cdf18873-c77e-461b-9f3e-249c69e93e17   4Gi        RWO            openebs-zfspv-retain   <unset>                 5s
ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k apply -f snapshot-retain.yaml
volumesnapshot.snapshot.storage.k8s.io/zfspv-snap-retain created

PVC and snapshot created till now 

ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get pvc
No resources found in default namespace.
ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get vs
NAME                READYTOUSE   SOURCEPVC          SOURCESNAPSHOTCONTENT   RESTORESIZE   SNAPSHOTCLASS     SNAPSHOTCONTENT                                    CREATIONTIME   AGE
zfspv-snap-retain   true         csi-zfspv-retain                           4Gi           zfspv-snapclass   snapcontent-bba00218-dc9f-407b-8222-1800d0c62f7a   5m15s          5m15s
ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get pv
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS     CLAIM                      STORAGECLASS           VOLUMEATTRIBUTESCLASS   REASON   AGE
pvc-cdf18873-c77e-461b-9f3e-249c69e93e17   4Gi        RWO            Retain           Released   default/csi-zfspv-retain   openebs-zfspv-retain   <unset>                          7m6s
ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get zv -n openebs
NAME                                       ZPOOL        NODEID          SIZE         STATUS   FILESYSTEM   AGE
pvc-cdf18873-c77e-461b-9f3e-249c69e93e17   zfspv-pool   node-0-309791   4294967296   Ready    zfs          7m28s
ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k delete pv pvc-cdf18873-c77e-461b-9f3e-249c69e93e17
persistentvolume "pvc-cdf18873-c77e-461b-9f3e-249c69e93e17" deleted
ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get pv
No resources found
ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get zv -n openebs
NAME                                       ZPOOL        NODEID          SIZE         STATUS   FILESYSTEM   AGE
pvc-cdf18873-c77e-461b-9f3e-249c69e93e17   zfspv-pool   node-0-309791   4294967296   Ready    zfs          7m52s



 Till now the PV is deleted and zv remains 



ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k apply -f generate-pv.yaml
persistentvolume/my-reclaimed-pv created
ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k apply -f generate-pvc.yaml
persistentvolumeclaim/my-reclaimed-pvc created
ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get pvc
NAME               STATUS   VOLUME            CAPACITY   ACCESS MODES   STORAGECLASS           VOLUMEATTRIBUTESCLASS   AGE
my-reclaimed-pvc   Bound    my-reclaimed-pv   5Gi        RWO            openebs-zfspv-retain   <unset>                 5s
ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get zv -n openebs
NAME                                       ZPOOL        NODEID          SIZE         STATUS   FILESYSTEM   AGE
pvc-cdf18873-c77e-461b-9f3e-249c69e93e17   zfspv-pool   node-0-309791   4294967296   Ready    zfs          12m
ashish@ubuntu-32gb-nbg1-2:~/sample-data/retain$ k get zv -n openebs -oyaml
apiVersion: v1
items:
- apiVersion: zfs.openebs.io/v1
  kind: ZFSVolume
  metadata:
    creationTimestamp: "2025-01-30T10:37:48Z"
    finalizers:
    - zfs.openebs.io/finalizer
    generation: 2
    labels:
      kubernetes.io/nodename: node-0-309791
    name: pvc-cdf18873-c77e-461b-9f3e-249c69e93e17
    namespace: openebs
    resourceVersion: "5699"
    uid: 2b97b9b2-11a9-44fd-9f21-a1387549f7e6
  spec:
    capacity: "4294967296"
    compression: "off"
    dedup: "off"
    fsType: zfs
    ownerNodeID: node-0-309791
    poolName: zfspv-pool
    quotaType: quota
    recordsize: 128k
    volumeType: DATASET
  status:
    state: Ready
kind: List
metadata:
  resourceVersion: ""
```

So this verifies that the ZV does not have `deletion timestamp` nor `is marked for deletion`.

sinhaashish: > Also one more scenario.
> 
> 1. Create Non-Retain PVC.
> 2. Create Snapshot.
> 3. Delete PVC.
> 4. Check for the PV to exist.
> 5. Delete the PV.
> 6. The PV should be gone and the ZV should stay around as there is the a snapshot, with marked for delete annotation.
> 7. Recreate the PV and PVC.
> 8. Now delete the snapshot.
> 9. Does the ZV go away + the underlying ZFS zvol/dataset?

```
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k apply -f pvc.yaml
persistentvolumeclaim/csi-zfspv created
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k apply -f snapshot.yaml
volumesnapshot.snapshot.storage.k8s.io/zfspv-snap created
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k get pvc
NAME        STATUS   VOLUME                                     CAPACITY   ACCESS MODES   STORAGECLASS    VOLUMEATTRIBUTESCLASS   AGE
csi-zfspv   Bound    pvc-d7a49bb4-98a3-436c-b92c-9cf4e78b2ef8   4Gi        RWO            openebs-zfspv   <unset>                 18s
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k get pv
NAME                                       CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM               STORAGECLASS    VOLUMEATTRIBUTESCLASS   REASON   AGE
pvc-d7a49bb4-98a3-436c-b92c-9cf4e78b2ef8   4Gi        RWO            Delete           Bound    default/csi-zfspv   openebs-zfspv   <unset>                          22s
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k get zv -n openebs
NAME                                       ZPOOL        NODEID          SIZE         STATUS   FILESYSTEM   AGE
pvc-d7a49bb4-98a3-436c-b92c-9cf4e78b2ef8   zfspv-pool   node-0-309791   4294967296   Ready    zfs          29s
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k delete pvc csi-zfspv
persistentvolumeclaim "csi-zfspv" deleted
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k get pv
No resources found
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k get zv -n openebs
NAME                                       ZPOOL        NODEID          SIZE         STATUS   FILESYSTEM   AGE
pvc-d7a49bb4-98a3-436c-b92c-9cf4e78b2ef8   zfspv-pool   node-0-309791   4294967296   Ready    zfs          55s
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k get zv -n openebs -oyaml
apiVersion: v1
items:
- apiVersion: zfs.openebs.io/v1
  kind: ZFSVolume
  metadata:
    annotations:
      openebs.io/mark-for-deletion: "true"
    creationTimestamp: "2025-01-30T11:02:43Z"
    finalizers:
    - zfs.openebs.io/finalizer
    generation: 2
    labels:
      kubernetes.io/nodename: node-0-309791
    name: pvc-d7a49bb4-98a3-436c-b92c-9cf4e78b2ef8
    namespace: openebs
    resourceVersion: "8276"
    uid: eae87e92-19da-4b90-a1ac-9db03142540d
  spec:
    capacity: "4294967296"
    compression: "off"
    dedup: "off"
    fsType: zfs
    ownerNodeID: node-0-309791
    poolName: zfspv-pool
    quotaType: quota
    recordsize: 128k
    volumeType: DATASET
  status:
    state: Ready
kind: List
metadata:
  resourceVersion: ""


Create the PV and pv using the existing ZV

ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k apply -f generate-pv.yaml
persistentvolume/my-reclaimed-pv created
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k get pv
NAME              CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS      CLAIM   STORAGECLASS    VOLUMEATTRIBUTESCLASS   REASON   AGE
my-reclaimed-pv   5Gi        RWO            Retain           Available           openebs-zfspv   <unset>                          4s

ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k apply -f pvc.yaml
persistentvolumeclaim/csi-zfspv created
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k get pvc
NAME        STATUS   VOLUME            CAPACITY   ACCESS MODES   STORAGECLASS    VOLUMEATTRIBUTESCLASS   AGE
csi-zfspv   Bound    my-reclaimed-pv   5Gi        RWO            openebs-zfspv   <unset>                 5s

ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k get zv -n openebs
NAME                                       ZPOOL        NODEID          SIZE         STATUS   FILESYSTEM   AGE
pvc-d7a49bb4-98a3-436c-b92c-9cf4e78b2ef8   zfspv-pool   node-0-309791   4294967296   Ready    zfs          8m43s
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k get vs
NAME         READYTOUSE   SOURCEPVC   SOURCESNAPSHOTCONTENT   RESTORESIZE   SNAPSHOTCLASS     SNAPSHOTCONTENT                                    CREATIONTIME   AGE
zfspv-snap   true         csi-zfspv                           4Gi           zfspv-snapclass   snapcontent-4f3e67cd-5d2c-4953-95f3-5551eaefebef   8m51s          8m51s

root@node-0-309791:~# zfs list -t all
NAME                                                                                                USED  AVAIL     REFER  MOUNTPOINT
zfspv-pool                                                                                          165K  9.20G       24K  /zfspv-pool
zfspv-pool/pvc-d7a49bb4-98a3-436c-b92c-9cf4e78b2ef8                                                  24K  4.00G       24K  legacy
zfspv-pool/pvc-d7a49bb4-98a3-436c-b92c-9cf4e78b2ef8@snapshot-4f3e67cd-5d2c-4953-95f3-5551eaefebef     0B      -       24K  -


Now delete the snapshot 

ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k get vs
NAME         READYTOUSE   SOURCEPVC   SOURCESNAPSHOTCONTENT   RESTORESIZE   SNAPSHOTCLASS     SNAPSHOTCONTENT                                    CREATIONTIME   AGE
zfspv-snap   true         csi-zfspv                           4Gi           zfspv-snapclass   snapcontent-4f3e67cd-5d2c-4953-95f3-5551eaefebef   8m51s          8m51s
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k delete vs zfspv-snap
volumesnapshot.snapshot.storage.k8s.io "zfspv-snap" deleted
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k get vs
No resources found in default namespace.

Snapshot deleted at zfs level 
root@node-0-309791:~# zfs list -t all
NAME                                                  USED  AVAIL     REFER  MOUNTPOINT
zfspv-pool                                            171K  9.20G       24K  /zfspv-pool
zfspv-pool/pvc-d7a49bb4-98a3-436c-b92c-9cf4e78b2ef8    24K  4.00G       24K  legacy

The ZV is present in the system along with pv and pvc and underneath zfs volume .

ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k get zv -n openebs
NAME                                       ZPOOL        NODEID          SIZE         STATUS   FILESYSTEM   AGE
pvc-d7a49bb4-98a3-436c-b92c-9cf4e78b2ef8   zfspv-pool   node-0-309791   4294967296   Ready    zfs          13m
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k get pvc
NAME        STATUS   VOLUME            CAPACITY   ACCESS MODES   STORAGECLASS    VOLUMEATTRIBUTESCLASS   AGE
csi-zfspv   Bound    my-reclaimed-pv   5Gi        RWO            openebs-zfspv   <unset>                 6m29s
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k get pv
NAME              CAPACITY   ACCESS MODES   RECLAIM POLICY   STATUS   CLAIM               STORAGECLASS    VOLUMEATTRIBUTESCLASS   REASON   AGE
my-reclaimed-pv   5Gi        RWO            Retain           Bound    default/csi-zfspv   openebs-zfspv   <unset>                          8m14s
ashish@ubuntu-32gb-nbg1-2:~/sample-data$ k get zv -n openebs -oyaml
apiVersion: v1
items:
- apiVersion: zfs.openebs.io/v1
  kind: ZFSVolume
  metadata:
    annotations:
      openebs.io/mark-for-deletion: "true"
    creationTimestamp: "2025-01-30T11:02:43Z"
    finalizers:
    - zfs.openebs.io/finalizer
    generation: 2
    labels:
      kubernetes.io/nodename: node-0-309791
    name: pvc-d7a49bb4-98a3-436c-b92c-9cf4e78b2ef8
    namespace: openebs
    resourceVersion: "8276"
    uid: eae87e92-19da-4b90-a1ac-9db03142540d
  spec:
    capacity: "4294967296"
    compression: "off"
    dedup: "off"
    fsType: zfs
    ownerNodeID: node-0-309791
    poolName: zfspv-pool
    quotaType: quota
    recordsize: 128k
    volumeType: DATASET
  status:
    state: Ready
kind: List
metadata:
  resourceVersion: ""


root@node-0-309791:~# zfs list -t all
NAME                                                  USED  AVAIL     REFER  MOUNTPOINT
zfspv-pool                                            171K  9.20G       24K  /zfspv-pool
zfspv-pool/pvc-d7a49bb4-98a3-436c-b92c-9cf4e78b2ef8    24K  4.00G       24K  legacy
root@node-0-309791:~#

File: pkg/zfs/volume.go (hunk 25 above)
Comment (Abhinandan-Purkait) [reply pkg/zfs/volume.go]: Let's remove this as we discussed?

File: pkg/driver/controller.go (hunk 10 above)
Comment (Abhinandan-Purkait) [reply pkg/driver/controller.go:898]: Right, so the zv remains and it still says marked for deletion. I think it would be good if you clean it up. When we get a create call and ZV exists and ZV has `openebs.io/mark-for-deletion: "true"` annotation, clear the annotation.

Comment (Abhinandan-Purkait) [reply pkg/driver/controller.go:898]: I think that was @tiagolobocastro's initial suggestion.

**Comment by Abhinandan-Purkait** [comment]
Looks good to me mostly. Thanks for taking this up. Will approve once the test passes and the other comments are addressed.

File: pkg/driver/controller.go (hunk 10 above)
Comment (tiagolobocastro) [reply pkg/driver/controller.go:898]: I kinda lost the thread here, but if we've verified the PV creation does re-trigger another CreateVolume call then yes we should remove the `openebs.io/mark-for-deletion: "true"`

File: pkg/driver/controller.go [hunk 27]
```diff
@@ -815,8 +841,39 @@ func (cs *controller) DeleteSnapshot(
...
+	// verify if the snapshot has already been deleted
+	snap, err := zfs.GetZFSSnapshot(snapshotID[1])
+	if snap != nil && snap.DeletionTimestamp != nil {
+		return csipayload.NewDeleteSnapshotResponseBuilder().Build(), nil
+	}
+
+	if err != nil {
+		if k8serror.IsNotFound(err) {
+			return csipayload.NewDeleteSnapshotResponseBuilder().Build(), nil
+		}
```
tiagolobocastro: Isn't there a race condition here, if the pod crashes after issuing `DeleteSnapshot`. Then when it restarts, if it finds the timestamp is set or snapshot is already deleted, it doesn't delete the volume?

tests/provision_test.go
```diff
@@ -51,62 +51,103 @@ func exhaustiveVolumeTests(parameters map[string]string) {
...
+
+	deletePVC(clonePvcNameBlock)
+	By("Verifying that ZV is present after pvc deletion ", func() { IsZVPresentConsistently(zvName) })
+	deleteSnapshot(pvcNameBlock, snapNameBlock)
+	By("Verifying that ZV is deleted after snapshot deletion ", func() { IsZVDeletedEventually(zvName) })
+
+	By("Deleting storage class", deleteStorageClass)
+}
+
+func blockVolCreationWithReclaimRetainTest() {
```
tiagolobocastro, line 118: If this is just for checking the retain, then we don't need any applications deployed?

pkg/driver/controller.go
```diff
@@ -524,19 +524,45 @@ func (cs *controller) DeleteVolume(
...
+			codes.NotFound,
+			"failed to handle delete volume request for {%s}, "+
+				"validation failed checking for snapshots. Error: %s",
+			req.VolumeId,
+			err.Error(),
 		)
 	}
 
+	// Delete the corresponding ZV CR only if there are no snapshots present for the volume
+	if len(snapList.Items) == 0 {
```
tiagolobocastro, line 540: Can we rely on the list of CRs btw, isn't this behind a cache?

tiagolobocastro: I guess it's not so bad if we get it wrong and mark-for-deletion is set. Might just mean we need a garbage collector in case we get the wrong list count?

pkg/driver/controller.go
sinhaashish: Sorry for my wrong statement earlier. The `CreateVolume ` is not called if we manually create the pv from zv.
Earlier the log which i had pasted was earlier one, i.e. the first time when the pvc was created. 

So finally we dont get any `CreateVolume` call when we create the PV  manually from ZV . This means that the ZV will be having the annotation `openebs.io/mark-for-deletion: "true"`

pkg/driver/controller.go
sinhaashish: When snapshot is there but snapList size return 0  because of cache; and set the marker for deletion.
There is no issue as zv is present with the marker set. Upon deletion of snapshot we will get  snapList size as 1 and the marker set ,so the volume will be deleted .

pkg/driver/controller.go
Abhinandan-Purkait: Hmm then what happens if we recreate the pv and the zv has an annotation and then we delete the last snapshot?

tiagolobocastro: hmm I think it's fine actually.
In case of retain there's no `DeleteVolume` to begin with, so the `ZV` does not have the annotation, so it won't get deleted when the last snapshot is deleted?

pkg/zfs/volume.go
sinhaashish: done

tests/provision_test.go
sinhaashish: added the code to create a pv and pv and check if the pvc is bound . Have not deployed the app

pkg/driver/controller.go
sinhaashish: removed the condition

pkg/driver/controller.go
sinhaashish: yes, its zv is left as is

pkg/builder/volbuilder/build.go
```diff
@@ -21,6 +21,9 @@ import (
 	apis "github.com/openebs/zfs-localpv/pkg/apis/openebs.io/zfs/v1"
 )
 
+// MarkForDeletionAnnotation is the annotation key
+const MarkForDeletionAnnotation string = "openebs.io/marked-for-deletion"
```
Abhinandan-Purkait, line 25: Shouldn't we use the one from, zfs/volume.go?

Abhinandan-Purkait (approved):

pkg/builder/volbuilder/build.go
sinhaashish: done

tests/provision_test.go
tiagolobocastro: So what's L122 and L129 doing?

pkg/driver/controller.go
```diff
@@ -503,14 +503,14 @@ func (cs *controller) DeleteVolume(
...
+	vol, err := zfs.GetZFSVolume(volumeID)
 	if vol != nil && vol.DeletionTimestamp != nil {
-		goto deleteResponse
+		return csipayload.NewDeleteVolumeResponseBuilder().Build(), nil
 	}
 
 	if err != nil {
 		if k8serror.IsNotFound(err) {
-			goto deleteResponse
+			return csipayload.NewDeleteVolumeResponseBuilder().Build(), nil
```
tiagolobocastro, line 513: I still see this, I think we need to do `zfs.MarkForDeletion(volumeID)` before succeeding in this case?

tests/provision_test.go
sinhaashish: I mis understood your comment . I thought you need the application to be deployed from manully created PVC which comes from retained ZV, so i added creating a manual pv and pvc from retained ZV. 

Any ways i have remove the app deployment code in this test as this was redundant, but the snapshot test exist as we want to verify that ZV is present even after snapshot is deleted.

pkg/driver/controller.go
sinhaashish: Why do we need it . Since we dont find the volume itslef, how can we mark the volume for deletion ?

tiagolobocastro: You mean if we can't find the snapshot, then we can't locate the parent volume either?

1. Add a new finalizer to the Snapshot, and remove after we either confirm parent volume deletion or confirm it doesn't need deletion
2. Make use of existing finalizer: The zfs node could issue ZV delete (if marked for deletion), before removing the snapshot finalizer

CC @Abhinandan-Purkait

We could do this as enhancement in addition to this PR and merge this which has the bulk of it?

sinhaashish: Here `IsNotFound` is for finding the zfs volume not snapshot . The line which you quoted in the comment errors out if teh zfs volume is not present .

tiagolobocastro: Doh this was on the wrong function, meant for equivalent thing on the `DeleteSnapshot` and not `DeleteVolume`

tiagolobocastro (approved):

PR #642, opened 2025-03-09
fix: backwards compatibility for quotatype on restore

Abhinandan-Purkait:
- Fix QuotaType being empty when coming from older version snap restore.


tiagolobocastro (approved):
Are we missing a test for this?

Abhinandan-Purkait:
> Are we missing a test for this?

We have tests, this is a special case when the upgrade happens, we have upgrade tests in plan for this.

PR #647, opened 2025-03-14
fix(grpc): stop server on SIGTERM and SIGINT

### LINKED ISSUE #646: does not cleanly shut down on SIGTERM
Running zfs-localpv natively on a node without daemonset requires common service handling (systemd and alternatives). When stopping the service (`SIGTERM`) the process keeps running. Thus there must be a goroutine still running. A second stop (`SIGTERM`) stops the process finally. If it matters it was running successfully in agent mode. Running the binary manually yields the same issue, first `Ctrl-C` (`SIGINT`) does not exit the process but yields the output below, needs a second `Ctrl-C` (`SIGINT`) to finally exit.

Last few lines in the output:
```
I0314 03:28:09.527507   10320 volume.go:198] Shutting down ZV workers
I0314 03:28:09.527609   10320 restore.go:179] Shutting down Restore workers
I0314 03:28:09.527531   10320 zfsnode.go:214] Shutting down Node controller
I0314 03:28:09.527638   10320 snapshot.go:187] Shutting down Snap workers
I0314 03:28:09.527644   10320 backup.go:184] Shutting down Bkp workers
```

process cleanly stops on SIGTERM

**Environment:**
- LocalPV-ZFS version: v2.7.1
- Kubernetes version (use `kubectl version`): doesn't matter
- Kubernetes installer & version: none
- Cloud provider or hardware configuration: doesn't matter
- OS (e.g. from `/etc/os-release`): Ubuntu jammy

dionysius:
**Why is this PR required? What issue does it fix?**: Stopping the binary doesn't exit the process because the grpc server goroutine is still running and doesn't get cancelled in any form.

**What this PR does?**: Listens (additionally, see below) to system signals SIGTERM and SIGINT to handle grpc server stopping.

**Does this PR require any upgrade changes?**: no

**If the changes in this PR are manually verified, list down the scenarios covered:**: Stopping the binary in agent mode will now stop the process (since all goroutines have naturally stopped).

This is only a quick solution but small and concise. Better approach would be a refactoring and passing around of the stop-channel from the top. Even better would be to introduce a clean context instead for the whole lifetime of the application process. Even the used function which register those signals has been updated to return a context many years ago: https://github.com/kubernetes-sigs/controller-runtime/blob/main/pkg/manager/signals/signal.go. See also: https://stackoverflow.com/a/74895157 and https://gist.github.com/embano1/e0bf49d24f1cdd07cffad93097c04f0a

**Checklist:**
- [x] Fixes #646
- [ ] Commit has unit tests

### DIFF
--- a/pkg/driver/agent.go
+++ b/pkg/driver/agent.go
@@ -53,6 +53,8 @@ func NewNode(d *CSIDriver) csi.NodeServer {
 	var ControllerMutex = sync.RWMutex{}
 
 	// set up signals so we handle the first shutdown signal gracefully
+	// TODO: (tech-debt) Setup signal handler more above, several files want to use stopCh and this function is only allowed to be used once (see #647)
+	// Affected files: pkg/driver/agent.go pkg/driver/controller.go pkg/driver/grpc.go
 	stopCh := signals.SetupSignalHandler()
 
 	// start the zfsnode resource watcher
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -103,6 +103,8 @@ func (cs *controller) init() error {
 		0, informers.WithNamespace(zfs.OpenEBSNamespace))
 
 	// set up signals so we handle the first shutdown signal gracefully
+	// TODO: (tech-debt) Setup signal handler more above, several files want to use stopCh and this function is only allowed to be used once (see #647)
+	// Affected files: pkg/driver/agent.go pkg/driver/controller.go pkg/driver/grpc.go
 	stopCh := signals.SetupSignalHandler()
 
 	cs.k8sNodeInformer = kubeInformerFactory.Core().V1().Nodes().Informer()
--- a/pkg/driver/grpc.go
+++ b/pkg/driver/grpc.go
@@ -21,8 +21,10 @@ import (
 	"fmt"
 	"net"
 	"os"
+	"os/signal"
 	"strings"
 	"sync"
+	"syscall"
 
 	"golang.org/x/net/context"
 	"google.golang.org/grpc"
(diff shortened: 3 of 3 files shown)


Abhinandan-Purkait (approved):
LGTM

Abhinandan-Purkait:
cc @tiagolobocastro @niladrih

niladrih (commented):
You could use https://github.com/kubernetes-sigs/controller-runtime/blob/v0.2.0/pkg/manager/signals/signal.go#L29, this is what we use for the operator controllers:
- https://github.com/openebs/zfs-localpv/blob/develop/pkg/driver/agent.go#L56
- https://github.com/openebs/zfs-localpv/blob/develop/pkg/driver/controller.go#L106

dionysius:
@niladrih I've considered it, but it will crash since https://github.com/kubernetes-sigs/controller-runtime/blob/v0.2.0/pkg/manager/signals/signal.go#L24. This function is only allowed to be called once.

Thats why the notes that the function should be called somewhere on the "top" of the application and the `stopCh` then passed down everywhere. Currently the agent mode and the controller mode do use this function and the grpc server is next to it. No easy way to pass around `stopCh` without starting a refactoring.

This PR is designed to be small and concise. Any better approach is a bigger refactor.

niladrih (approved):
Makes sense. Could you add a `// TODO:` comment explaining the tech debt involved in this approach? Thanks.

dionysius:
Hmm... I was about to prepare such commit. I just saw one more option, define:
```
var (
  stopCh := signals.SetupSignalHandler()
)
```
on package level for `pkg/driver`. All affected files reside in there

dionysius:
Yeah, this variant doesn't look to bad either: https://github.com/openebs/zfs-localpv/compare/develop...dionysius:zfs-localpv:grpc_stop_2?expand=1.

But needs some comments and a quick check here since I want to be sure that two routines can wait for the channel simultanously

dionysius:
No, it doesn't behave the same. While the issue might be fixed (the process stops), it doesn't report the stopped routines (there are a few missing):
```
I0314 07:32:14.850218   23156 grpc.go:142] Shutting down gRPC server gracefully
I0314 07:32:14.850366   23156 backup.go:184] Shutting down Bkp workers
I0314 07:32:14.850422   23156 zfsnode.go:214] Shutting down Node controller
```

Yeah this PR is the quick way to go, anything else would be best using a context and or "proxy"copy the `stopCh` to the various places.

I've updated the places with TODO infos accordingly.

tiagolobocastro (approved):

Abhinandan-Purkait:
@dionysius Are we good to merge?

PR #653, opened 2025-04-29
feat(helm chart): refactor labels and add logging labels

Abhinandan-Purkait:
- Adds label selectors for logging.
- Removes redundant labels.

### DIFF
--- a/deploy/helm/charts/templates/zfs-controller.yaml
+++ b/deploy/helm/charts/templates/zfs-controller.yaml
@@ -19,9 +19,12 @@ spec:
       annotations: {{ toYaml . | nindent 8 }}
       {{- end }}
       labels:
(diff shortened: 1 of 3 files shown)


niladrih (approved):

deploy/helm/charts/templates/_helpers.tpl
tiagolobocastro, line 1: role is being remove here, but still defined in the values.yaml
also it's used in documents etc for querying components with role=openebs-zfs
perhaps we should not remove it?

Abhinandan-Purkait: Hmm, then we will run into limits with Loki, as it can go upto 15 labels at max.

Abhinandan-Purkait: Reverted this.

tiagolobocastro (approved):

PR #655, opened 2025-05-14
feat(csi): update csi spec to v1_11 and sidecars

Abhinandan-Purkait:
- Updates CSI Spec to v1.11
- Updates the sidecars to latest

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -65,6 +65,8 @@ type controller struct {
 	zfsNodeInformer cache.SharedIndexInformer
 
 	volumeLock *volumeLock
+
+	csi.UnimplementedControllerServer
 }
 
 // NewController returns a new instance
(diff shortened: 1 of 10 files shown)


pkg/driver/controller.go
```diff
@@ -65,6 +65,8 @@ type controller struct {
 	zfsNodeInformer cache.SharedIndexInformer
 
 	volumeLock *volumeLock
+
+	csi.UnimplementedControllerServer
```
niladrih, line 69: This type should be member of the CSIDriver type?

Abhinandan-Purkait: No, it's different for different services shouldn't be a part of the common type.

niladrih (approved):

niladrih (approved):

dsharma-dc (approved):

PR #656, opened 2025-05-16
fix(node): handle properly volume resize when asking for more than the available space

### LINKED ISSUE #649: REFQUOTA is updated even when volume expansion fails due to insufficient space
When attempting to expand a PVC using ZFS LocalPV, I've observed that the REFQUOTA value gets updated to the new desired size even when the actual volume expansion fails due to insufficient space in the zpool.

The REFQUOTA value should only be updated after a successful volume expansion. If the expansion fails due to insufficient space, the REFQUOTA should remain at its original value or be reverted back.

**The output of the following commands will help us better understand what's going on**:
```
zpool list
NAME         SIZE  ALLOC   FREE  CKPOINT  EXPANDSZ   FRAG    CAP  DEDUP    HEALTH  ALTROOT
zfspv-pool  99.5G   192M  99.3G        -         -     4%     0%  1.00x    ONLINE  -
```
```
 zfs list -o name,quota,reservation,refquota,refreservation,used,available
NAME                                                 QUOTA  RESERV  REFQUOTA  REFRESERV   USED  AVAIL
zfspv-pool                                            none    none      none         3G  64.3G  35.1G
zfspv-pool/pvc-4245caa5-cab9-40b1-aac8-431569fb195a   none    none      200G         5G     5G  37.1G
```

**Environment:**
- LocalPV-ZFS version 2.7.1

plaffitt:
## Pull Request template

**Why is this PR required? What issue does it fix?**: closes #649

**What this PR does?**: It reverts `(ref)quota` to the previous value of `(ref)reservation` (the value that we read from zfs before setting those 2 properties together). 

**Does this PR require any upgrade changes?**: Not sure about what this question mean, but I don't think so.

- Expand a PVC to an acceptable size and check that it worked as before
- Expand a PVC to an unreasonable size and check that it reverts to the original size while returning a grpc `ResourceExhausted` code

_N/A_

**Checklist:**
- [x] Fixes #649
- [ ] Commit has unit tests

### DIFF
--- a/pkg/driver/agent.go
+++ b/pkg/driver/agent.go
@@ -380,8 +380,15 @@ func (ns *node) NodeExpandVolume(
 	}
 
 	if err = zfs.ResizeZFSVolume(vol, req.GetVolumePath(), resizefs); err != nil {
+		code := codes.Internal
+
+		// error defined in zfs source code: https://github.com/openzfs/zfs/blob/zfs-2.3.99/lib/libzfs/libzfs_util.c#L551
+		if strings.Contains(err.Error(), "size is greater than available space") {
+			code = codes.ResourceExhausted
+		}
+
 		return nil, status.Errorf(
-			codes.Internal,
+			code,
 			"failed to handle NodeExpandVolume Request for %s, {%s}",
 			req.VolumeId,
 			err.Error(),
--- a/pkg/zfs/volume_test.go
+++ b/pkg/zfs/volume_test.go
@@ -62,3 +62,26 @@ func TestReservationProperty(t *testing.T) {
 		})
 	}
 }
+
+func TestQuotaProperty(t *testing.T) {
+	tests := []struct {
+		name      string
+		quotaType string
+		expected  string
+	}{
+		{"Empty quotaType defaults to quota", "", "quota"},
+		{"Valid quotaType quota", "quota", "quota"},
+		{"Valid quotaType refquota", "refquota", "refquota"},
+		{"Invalid quotaType defaults to quota", "invalid", "quota"},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			result := quotaProperty(tt.quotaType)
+			if result != tt.expected {
+				t.Errorf("For quotaType %q, expected %q but got %q",
+					tt.quotaType, tt.expected, result)
+			}
+		})
+	}
+}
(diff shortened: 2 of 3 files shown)


pkg/driver/agent.go
```diff
@@ -379,8 +379,15 @@ func (ns *node) NodeExpandVolume(
 	}
 
 	if err = zfs.ResizeZFSVolume(vol, req.GetVolumePath(), resizefs); err != nil {
+		code := codes.Internal
+
+		// error defined in zfs source code: https://github.com/openzfs/zfs/blob/zfs-2.3.99/lib/libzfs/libzfs_util.c#L551
+		if strings.Contains(err.Error(), "size is greater than available space") {
+			code = codes.ResourceExhausted
```
tiagolobocastro, line 387: I wonder why the actual resize isn't confirmed in the controller resize, CC @Abhinandan-Purkait

Abhinandan-Purkait: Isn't it because the NodeExpand happens later and the job of controller is just to update the CR and the node plugin does the job based on the CR change?

tiagolobocastro: Yeah that's what I mean, so the offline resize doesn't really resize? Just sets the request in the CR

Abhinandan-Purkait (commented):
Thanks! Changes look reasonable to me. Can you please add a CI test for the same if possible?

> Does this PR require any upgrade changes?: 
> Not sure about what this question mean, but I don't think so.

It basically means do we need to handle anything for upgrading from an older version to this.
Anyway the changes don't seem to directly impact upgrades but I will double check the scenarios you have mentioned, on a volume created on a previous version and the the resize being performed after upgrade.

pkg/driver/agent.go
Abhinandan-Purkait: Yeah

Abhinandan-Purkait:
@plaffitt I see you have created a new issue on the same, and your suggestion on adding a check on the ControllerExpand. That sounds fair to me, can you add the same in the PR. I tried the same and you can refer the code from below if it looks good to you. Thanks. 

https://github.com/openebs/zfs-localpv/compare/add_check_controller?expand=1 cc @tiagolobocastro

pkg/driver/agent.go
tiagolobocastro: I'd suggest we should also change that, otherwise the offline resize is not really resizing, it's just setting the request, which may fail only when a pod is scheduled to run.

plaffitt: I've already opened an issue about it #657

plaffitt:
Hello, is there anything missing to review/merge this PR?

Abhinandan-Purkait (approved):

tiagolobocastro (approved):

Abhinandan-Purkait:
@plaffitt Are we good to merge now?

PR #671, opened 2025-06-25
ci: add staging ci for release prep

Abhinandan-Purkait:
- Adds a workflow dispatch to populate the release artifacts to the ghcr
- Modifies the release workflow to populate artifacts from ghcr dev registry to prod registries
- Modifies the build and push to produce artifacts on the ghcr

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -6,111 +6,8 @@ on:
       - "created"
 
 jobs:
-  lint:
+  release-images-and-chart:
     runs-on: ubuntu-22.04
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-        with:
-          fetch-depth: 0
-
-      - uses: cachix/install-nix-action@v22
-      - uses: rrbutani/use-nix-shell-action@v1.1.0
-        with:
-          file: shell.nix
-
-      - name: Check if the chart is publishable
-        run: |
-          TAG=${{ github.event.release.tag_name }}
-          ./scripts/update-chart-version.sh --tag $TAG --publish-release
-
-      - name: Run chart-testing lint
-        run: |
-          ct lint --config ct.yaml
-
-  unit-tests:
-    runs-on: ubuntu-22.04
-    needs: ["lint"]
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-
-      - name: Set up Go 1.20
-        uses: actions/setup-go@v4
-        with:
-          go-version: 1.20.14
-          cache: false
-
-      - name: Unit test
-        run: make test
-
-      - name: Upload Coverage Report
-        uses: codecov/codecov-action@v4
-        with:
-          token: ${{ secrets.CODECOV_TOKEN }}
-          file: ./coverage.txt
-          name: coverage-$(date +%s)
-          flags: unittests
-
-  bdd-tests:
-    runs-on: ubuntu-22.04
-    needs: ["unit-tests"]
-    strategy:
-      fail-fast: true
-      matrix:
-        kubernetes: [v1.27.3]
-    steps:
-      - name: Checkout
-        uses: actions/checkout@v4
-
-      - name: Set up Go 1.20
-        uses: actions/setup-go@v4
-        with:
-          go-version: 1.20.14
-          cache: false
-
-      - name: Build images locally
-        run: make zfs-driver-image
-
-      - name: Install zfs utils
-        run: sudo apt-get install zfsutils-linux -y
-
-      - name: Setup Minikube-Kubernetes
-        uses: medyagh/setup-minikube@latest
-        with:
-          cache: false
-          minikube-version: 1.31.1
-          driver: none
-          kubernetes-version: ${{ matrix.kubernetes }}
-          cni: calico
-          start-args: "--install-addons=false"
-
-      - name: Setting environment variables
-        run: |
-          echo "KUBECONFIG=$HOME/.kube/config" >> $GITHUB_ENV
-
-      - name: bootstrap
-        run: make bootstrap
-
-      - name: Running tests
-        env:
-          OPENEBS_NAMESPACE: "openebs"
-        run: |
-          make ci
-          make helm-install
-          make sanity
-
-      - name: Upload CI Test Coverage Report
-        uses: codecov/codecov-action@v4
-        with:
-          token: ${{ secrets.CODECOV_TOKEN }}
-          file: ./tests/bdd_coverage.txt
-          name: coverage-bdd_coverage-$(date +%s)
-          flags: bddtests
-
-  csi-driver:
-    runs-on: ubuntu-22.04
-    needs: ["bdd-tests"]
     steps:
       - name: Checkout
         uses: actions/checkout@v4
@@ -122,55 +19,37 @@ jobs:
           echo "NIX_PATH=$NIX_PATH" >> $GITHUB_ENV
           nix-shell --pure --run "echo" ./shell.nix
 
-      - name: Check if the chart is publishable
-        run: |
-          TAG=${{ github.event.release.tag_name }}
-          nix-shell --pure --run "./scripts/update-chart-version.sh --tag $TAG --publish-release" ./shell.nix
-
       - name: Set Image Org
         run: |
-          [ -z "${{ secrets.IMAGE_ORG }}" ] && IMAGE_ORG=openebs || IMAGE_ORG=${{ secrets.IMAGE_ORG}}
+          [ -z "${{ secrets.IMAGE_ORG }}" ] && IMAGE_ORG=openebs || IMAGE_ORG=${{ secrets.IMAGE_ORG }}
           echo "IMAGE_ORG=${IMAGE_ORG}" >> $GITHUB_ENV
 
-      - name: Set IMAGE_TAG and BRANCH
+      - name: Log in to GHCR for Helm
+        run: nix-shell --pure --run "echo ${{ secrets.GITHUB_TOKEN }} | helm registry login ghcr.io -u ${{ github.actor }} --password-stdin" ./shell.nix
+
+      - name: Pull OCI chart from GHCR
+        env:
+          IMAGE_ORG: ${{ env.IMAGE_ORG }}
+          RAW_TAG: ${{ github.ref_name }}
         run: |
-          BRANCH=${{ github.ref_name }}
-          echo "BRANCH=$BRANCH" >> $GITHUB_ENV
-          echo "IMAGE_TAG=$(awk -F': ' '/^version:/ {print $$2}' deploy/helm/charts/Chart.yaml)" >> $GITHUB_ENV
+          TAG="${RAW_TAG#v}"
+          echo "Using chart version: $TAG"
 
-      - name: Set Build Date
-        id: date
+          rm -rf deploy/helm/charts
+          mkdir -p charts-staged charts-final deploy/helm/charts
+
+          CHART_REF=oci://ghcr.io/${IMAGE_ORG}/helm/zfs-localpv
+          helm pull "$CHART_REF" \
+            --version "$TAG" \
+            --destination charts-staged
+
+          tar -xzf charts-staged/zfs-localpv-"$TAG".tgz -C charts-final
+
+          mv charts-final/zfs-localpv/* deploy/helm/charts
+
+      - name: Set Chart Version
         run: |
-          echo "DATE=$(date -u +'%Y-%m-%dT%H:%M:%S%Z')" >> $GITHUB_OUTPUT
-
-      - name: Docker meta
-        id: docker_meta
-        uses: docker/metadata-action@v4
-        with:
-          # add each registry to which the image needs to be pushed here
-          images: |
-            ${{ env.IMAGE_ORG }}/zfs-driver
-            quay.io/${{ env.IMAGE_ORG }}/zfs-driver
-            ghcr.io/${{ env.IMAGE_ORG }}/zfs-driver
-          tags: |
-            type=raw,value=latest,enable=false
-            type=raw,value=${{ env.IMAGE_TAG }}
-
-      - name: Print Tag info
-        run: |
-          echo "BRANCH: ${{ env.BRANCH }}"
-          echo "${{ steps.docker_meta.outputs.tags }}"
-
-      - name: Set up QEMU
-        uses: docker/setup-qemu-action@v3
-        with:
-          platforms: all
-
-      - name: Set up Docker Buildx
-        id: buildx
-        uses: docker/setup-buildx-action@v3
-        with:
-          version: v0.13.1
+          echo "VERSION=$(awk -F': ' '/^version:/ {print $2}' deploy/helm/charts/Chart.yaml)" >> $GITHUB_ENV
 
       - name: Login to Docker Hub
         uses: docker/login-action@v3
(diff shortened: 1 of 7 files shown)


scripts/yq_utils.sh
```diff
@@ -0,0 +1,15 @@
+# yq-go eats up blank lines
+# this function gets around that using diff with --ignore-blank-lines
+yq_ibl()
+{
+  set +e
+  diff_out=$(diff -B <(yq '.' "$2") <(yq "$1" "$2"))
+  error=$?
```
tiagolobocastro: this is not right, error may not be caught

```suggestion
  diff_out=$(diff -B <(yq '.' "$2") <(yq "$1" "$2")) || error=$?
```

.github/workflows/build_and_push.yml
```diff
@@ -154,11 +154,8 @@ jobs:
         id: docker_meta
         uses: docker/metadata-action@v4
         with:
-          # add each registry to which the image needs to be pushed here
           images: |
-            ${{ env.IMAGE_ORG }}/zfs-driver
-            quay.io/${{ env.IMAGE_ORG }}/zfs-driver
-            ghcr.io/${{ env.IMAGE_ORG }}/zfs-driver
+            ghcr.io/${{ env.IMAGE_ORG }}/dev/zfs-driver
```
tiagolobocastro: can't we keep pushing dev and prerel as usual?

.github/workflows/build_and_push.yml
```diff
@@ -219,8 +203,35 @@ jobs:
...
+          nix-shell --pure --run "echo" ./shell.nix
+
+      - name: Set Image Org
+        run: |
+          [ -z "${{ secrets.IMAGE_ORG }}" ] && IMAGE_ORG=openebs || IMAGE_ORG=${{ secrets.IMAGE_ORG}}
+          echo "IMAGE_ORG=${IMAGE_ORG}" >> $GITHUB_ENV
+
+      - name: Update the registry and the repository
+        run: |
+          nix-shell --pure --run "./scripts/update-reg-repo.sh --registry ghcr.io --repository ${{ env.IMAGE_ORG }}/dev/zfs-driver" ./shell.nix
```
tiagolobocastro: why is all this needed for dev and prerel?

Abhinandan-Purkait: Otherwise the images will point to dockerhub right?

.github/workflows/build_and_push.yml
Abhinandan-Purkait: This is the dev only right?

Abhinandan-Purkait: Ohh do you mean gh-pages and docker hub? I though since we are doing it anyway for staging, why not for this as well?

.github/workflows/build_and_push.yml
Abhinandan-Purkait: I have made the dev and prerel go to ghcr as well now. Everything goes to ghcr dev and only the final tagged ones would go to openebs dockerhub and gh-pages.

tiagolobocastro: This would mean we'd have to change the umbrella prerel/dev to pull from ghcr as well?

.github/workflows/build_and_push.yml
tiagolobocastro: Yeah we could leave the dev/prerel as is. I think it may reduce a lot of the changes here?

Abhinandan-Purkait: Reverted

.github/workflows/build_and_push.yml
Abhinandan-Purkait: Reverted

.github/workflows/release.yml
```diff
@@ -122,55 +19,37 @@ jobs:
...
+          helm pull "$CHART_REF" \
+            --version "$TAG" \
+            --destination charts-staged
 
-      - name: Set up Docker Buildx
-        id: buildx
-        uses: docker/setup-buildx-action@v3
-        with:
-          version: v0.13.1
+          tar -xzf charts-staged/zfs-localpv-"$TAG".tgz -C charts-final
```
tiagolobocastro, line 46: you can use --untar=false ?

.github/workflows/release.yml
```diff
@@ -122,55 +19,37 @@ jobs:
...
-            ghcr.io/${{ env.IMAGE_ORG }}/zfs-driver
-          tags: |
-            type=raw,value=latest,enable=false
-            type=raw,value=${{ env.IMAGE_TAG }}
-
-      - name: Print Tag info
-        run: |
-          echo "BRANCH: ${{ env.BRANCH }}"
-          echo "${{ steps.docker_meta.outputs.tags }}"
+          rm -rf deploy/helm/charts
```
tiagolobocastro, line 38: why is this being deleted?

.github/workflows/release.yml
```diff
@@ -122,55 +19,37 @@ jobs:
...
-        uses: docker/setup-buildx-action@v3
-        with:
-          version: v0.13.1
+          tar -xzf charts-staged/zfs-localpv-"$TAG".tgz -C charts-final
+
+          mv charts-final/zfs-localpv/* deploy/helm/charts
+
+      - name: Set Chart Version
+        run: |
+          echo "VERSION=$(awk -F': ' '/^version:/ {print $2}' deploy/helm/charts/Chart.yaml)" >> $GITHUB_ENV
```
tiagolobocastro, line 52: Isn't version the TAG?

.github/workflows/release.yml
```diff
@@ -122,55 +19,37 @@ jobs:
...
-          [ -z "${{ secrets.IMAGE_ORG }}" ] && IMAGE_ORG=openebs || IMAGE_ORG=${{ secrets.IMAGE_ORG}}
+          [ -z "${{ secrets.IMAGE_ORG }}" ] && IMAGE_ORG=openebs || IMAGE_ORG=${{ secrets.IMAGE_ORG }}
           echo "IMAGE_ORG=${IMAGE_ORG}" >> $GITHUB_ENV
 
-      - name: Set IMAGE_TAG and BRANCH
-        run: |
-          BRANCH=${{ github.ref_name }}
-          echo "BRANCH=$BRANCH" >> $GITHUB_ENV
-          echo "IMAGE_TAG=$(awk -F': ' '/^version:/ {print $$2}' deploy/helm/charts/Chart.yaml)" >> $GITHUB_ENV
+      - name: Log in to GHCR for Helm
```
tiagolobocastro, line 27: Do we need to login? Ah because the pushed package is private?

.github/workflows/release.yml
```diff
@@ -192,40 +71,32 @@ jobs:
...
+          QY: quay.io/${{ env.IMAGE_ORG }}/zfs-driver:${{ env.VERSION }}
+        run: |
+            nix-shell --pure --run "crane copy --platform all ${{ env.SRC }} ${{ env.QY }}"
 
-      - name: Check if the chart is publishable
+      - name: Update the registry and the repository
         run: |
-          TAG=${{ github.event.release.tag_name }}
-          nix-shell --pure --run "./scripts/update-chart-version.sh --tag $TAG --publish-release" ./shell.nix
+          nix-shell --pure --run "./scripts/update-reg-repo.sh --registry docker.io/ --repository ${{ env.IMAGE_ORG }}/zfs-driver" ./shell.nix
```
tiagolobocastro: ah we still need to modify the chart! that's unfortunate

scripts/update-reg-repo.sh
```diff
@@ -0,0 +1,73 @@
...
+
+set -euo pipefail
+
+SCRIPT_DIR="$(dirname "$(realpath "${BASH_SOURCE[0]:-"$0"}")")"
+ROOT_DIR="$SCRIPT_DIR/.."
+CHART_DIR="$ROOT_DIR/deploy/helm/charts"
+VALUES_YAML="$CHART_DIR/values.yaml"
+
+NEW_REGISTRY="ghcr.io"
+NEW_REPOSITORY="openebs/dev"
```
tiagolobocastro: this is the namespace not the repository?

scripts/update-reg-repo.sh
```diff
@@ -0,0 +1,73 @@
...
+help() {
+  cat <<EOF
+Usage: $(basename "$0") [OPTIONS]
+
+Options:
+  --registry                                The registry to be updated to.
+  --repository                              The repository to be updated to.
+
+Examples:
+  $(basename "$0") --registry ghcr.io --repository openebs/dev
```
tiagolobocastro: this here seems to be the namespace

scripts/update-reg-repo.sh
```diff
@@ -0,0 +1,73 @@
...
+echo_stderr() {
+  echo -e "${1}" >&2
+}
+
+die()
+{
+  local _return="${2:-1}"
+  echo_stderr "$1"
+  exit "${_return}"
+}
```
tiagolobocastro: these could be a source as well, and maybe rename them to match what we have on openebs

.github/workflows/staging.yml
```diff
@@ -0,0 +1,205 @@
...
+          echo "IMAGE_ORG=${IMAGE_ORG}" >> $GITHUB_ENV
+
+      - name: Check if chart is publishable
+        run: nix-shell --pure --run "./scripts/update-chart-version.sh --tag ${TAG} --publish-release" ./shell.nix
+
+      - name: Set IMAGE_TAG and BRANCH
+        run: |
+          BRANCH=${{ github.ref_name }}
+          echo "BRANCH=$BRANCH" >> $GITHUB_ENV
+          echo "IMAGE_TAG=$(awk -F': ' '/^version:/ {print $2}' deploy/helm/charts/Chart.yaml)" >> $GITHUB_ENV
```
tiagolobocastro, line 123: Is this not the TAG that we check above on L117?

tiagolobocastro: Ah because TAG has a v but then we strip the v out?

.github/workflows/release.yml
tiagolobocastro: then we don't need the tar?

tiagolobocastro: how about packing the chart on its final form and adding a template function which replaces the registry and namespace if a helm env var is set? ex: `--set zfs.staging=true` ?

.github/workflows/release.yml
```diff
@@ -6,111 +6,8 @@ on:
       - "created"
 
 jobs:
-  lint:
+  copy-images-and-chart:
```
tiagolobocastro: nit: maybe
```suggestion
  release-images-and-chart:
```

tiagolobocastro (approved):

.github/workflows/release.yml
Abhinandan-Purkait: It's a safety measure to ensure the action doesn't confuse with other charts in the workspace.

.github/workflows/release.yml
Abhinandan-Purkait: Yes, it fails to pull without login.

.github/workflows/release.yml
Abhinandan-Purkait: What is the use of that? Also doesn't seem like an option in the tar binary.

.github/workflows/release.yml
Abhinandan-Purkait: No the v needs to be stripped so, instead always get from the chart keep it consistent and we don't miss it.

.github/workflows/release.yml
Abhinandan-Purkait: We can, but any issues doing this way?

Abhinandan-Purkait: Either way we need to set the values. Also we would need to expose an extra value?

.github/workflows/staging.yml
Abhinandan-Purkait: Yeah, locals use just the version without v everywhere.

scripts/update-reg-repo.sh
Abhinandan-Purkait: Yeah namespace, but in the values it's repository that's why kept it same so that there is no confusion.

scripts/update-reg-repo.sh
Abhinandan-Purkait: Yes it is the namespace but if you see the values.yaml it calls it repository here.

scripts/update-reg-repo.sh
Abhinandan-Purkait: Openebs seems to have these same names it seems? Can you point me to the one you are mentioning?

.github/workflows/release.yml
tiagolobocastro: I meant on the helm pull command
But also why do we need the package at all here if we're modifying the values.yaml later?

.github/workflows/release.yml
tiagolobocastro: No issues, was just to avoid modifying the package so even the sha of the tar would be the same

scripts/update-reg-repo.sh
tiagolobocastro: Yeah but here it's not the repository it's the namespace, the repository would be openebs/dev/zfs-driver

scripts/update-reg-repo.sh
tiagolobocastro: no, on the values it is the repository it's `openebs/zfs-driver` which is namespace + image

scripts/update-reg-repo.sh
tiagolobocastro: https://github.com/openebs/openebs/blob/aaed909deb9ccebb69e7d36edd01ed07c52b54e2/scripts/helm/images.sh#L13
But I see even there it's not used everywhere, it's a mix

.github/workflows/release.yml
Abhinandan-Purkait: Yes, otherwise how do we ensure it's the same chart that we tested on staging?

.github/workflows/release.yml
Abhinandan-Purkait: The action doesn't seem to support the bundled tar.

scripts/update-reg-repo.sh
Abhinandan-Purkait: It is namespace but the value that we set on the values.yaml is termed as repository and not namespace.

scripts/update-reg-repo.sh
Abhinandan-Purkait: Done.

scripts/update-reg-repo.sh
Abhinandan-Purkait: Done.

scripts/update-reg-repo.sh
tiagolobocastro: What I meant is, it must be consistent. You've renamed it but it's still not consistent IMHO.
If the parameter name is repository, then we should pass the entire thing "openebs/dev/zfs-driver" rather than "openebs/dev" which is a namespace
If the parameter name is namespace, then we should only pass "openebs/dev", and let the script add the "zfs-driver", rather than passing "openebs/dev/zfs-driver" which is not a namespace

Abhinandan-Purkait: I have removed the defaults and made the args compulsory,  so I will revert it back to repository and expect repository to be passed.

Abhinandan-Purkait: Done.

tiagolobocastro (approved):

PR #678, opened 2025-10-28
chore(go runtime): update go runtime and all dependents to latest

Abhinandan-Purkait:
### Changes in the PR

- Updates go runtime to 1.24.7
- Updates all the dependencies to the latest version
- Updates all the workflows to point to 1.24.7
- Updates the CRDs and generated code as per newer code generation.
- Updates Dockerfiles to point to go 1.24.7


buildscripts/generate-manifests.sh
```diff
@@ -1,44 +1,79 @@
...
+
+if [[ -z "${HELM_BIN}" ]]; then
+  echo "ERROR: helm not found. Please install Helm (https://helm.sh)." >&2
+  exit 1
+fi
+
+# Ensure output directories exist
+mkdir -p -- "${DEPLOY_YAML_DIR}" "${CRD_CHART_TEMPLATE_DIR}"
+
+# If a stray 'yamls' directory with a trailing U+200B exists, consolidate and remove it
```
tiagolobocastro, line 40: why would this exist?

buildscripts/generate-manifests.sh
```diff
@@ -1,44 +1,79 @@
...
+if [[ -d "${DEPLOY_YAML_DIR}${ZWSP}" ]]; then
+  shopt -s dotglob nullglob
```
tiagolobocastro, line 42: do we need to revert this and end of the if?

Makefile
```diff
@@ -10,9 +10,9 @@ VETARGS?=-asmdecl -atomic -bool -buildtags -copylocks -methods \
 EXTERNAL_TOOLS=\
 	golang.org/x/tools/cmd/cover@latest \
 	golang.org/x/lint/golint@latest \
-	github.com/axw/gocov/gocov@v1.1 \
+	github.com/axw/gocov/gocov@latest \
```
tiagolobocastro: should all these be versioned to avoid future breakage?

tiagolobocastro (approved):

buildscripts/generate-manifests.sh
Abhinandan-Purkait: Not sure, but whenever the code generation tool runs it trys to put everything in a directory it creates temporary even if the directory is provided it just takes that name and creates a new directory with <given-path><invisible trailing chars>. That's why this is needed to make sure we don't get confused.

buildscripts/generate-manifests.sh
Abhinandan-Purkait: Indeed. Done!

Makefile
Abhinandan-Purkait: Indeed. Let me fix that.

PR #692, opened 2026-01-05
ci: reuse actions across workflows

tiagolobocastro:

Duplication of gha code across multiple workflows and different jobs within the same workflow

**What this PR does?**:

Use install and minikube composite actions in-place. Create separate workflows for lint and test, allowing reuse across different workflows.

No

To consider also, do the tests need to be sequential?

**Checklist:**
- [ ] Fixes #<issue number>
- [ ] Commit has unit tests

### DIFF
--- a/.github/actions/setup-minikube/action.yml
+++ b/.github/actions/setup-minikube/action.yml
@@ -1,0 +1,34 @@
+name: Setup Minikube K8s Cluster
+
+description: Setup Minikube K8s Cluster
+
+inputs:
+  kubernetes-version:
+    description: Version of K8s to install
+    required: true
+  minikube-version:
+    description: Version of minikube to install
+    required: false
+    default: 1.31.1
+
+runs:
+  using: composite
+  steps:
+    - name: Install zfs utils
+      shell: bash
+      run: sudo apt-get install zfsutils-linux -y
+
+    - name: Setup Minikube-Kubernetes
+      uses: medyagh/setup-minikube@latest
+      with:
+        cache: false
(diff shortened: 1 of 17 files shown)


tiagolobocastro:
@Abhinandan-Purkait the custom node id tests take half the time of the run, would it make sense to run that in parallel with the matrix strategy?

Abhinandan-Purkait:
> @Abhinandan-Purkait the custom node id tests take half the time of the run, would it make sense to run that in parallel with the matrix strategy?

That's a good idea tbh. It takes a while to run those two one after another.

.github/actions/setup-minikube/action.yml
```diff
@@ -0,0 +1,34 @@
...
+description: Setup Minikube K8s Cluster
+
+inputs:
+  kubernetes-version:
+    description: Version of K8s to install
+    required: true
+  minikube-version:
+    description: Version of minikube to install
+    required: false
+    default: 1.31.1
```
Abhinandan-Purkait: Since you are here let's bump it up to some non EOL version. Probably 1.33.x

Abhinandan-Purkait (approved):
Looks neat. Thanks! Btw I saw the code coverage dropped to 12% we are probably not pushing everything now?

tiagolobocastro:
> Looks neat. Thanks! Btw I saw the code coverage dropped to 12% we are probably not pushing everything now?

The problem is that uni-tests were not being uploaded correctly, see <https://github.com/openebs/zfs-localpv/pull/692/commits/6e4e95b2ede37df0efc7d5637fb8ccaaeb05e038>

PR #700, opened 2026-03-10
feat: add helm global values

krishnaGajabi:
## Why is this PR required? What issue does it fix?

Currently, the Helm charts do not provide global variables for configuring common parameters such as the image registry, image pull secrets, image pull policy, and analytics settings.

When deploying the OpenEBS Helm charts, users often need to override these values across multiple subcharts and components. This makes it difficult to locate and configure all the required fields consistently. Providing global variables simplifies configuration and improves the user experience.

## What does this PR do?

This PR introduces the following global configuration variables:
```yaml
global:
  imageRegistry: ""
  # set the global image registry
  imagePullSecrets: []
  # - secret
  # or 
  # -name: secret
  imagePullPolicy: ""
  # set global image pull policy
  analytics:
    enabled: true
  # disable analytics
```

These variables allow users to configure common settings in a single place, which will then be applied across the chart components.

## Does this PR require any upgrade changes?

No.
This change is backward compatible and does not require any upgrade steps.

## Verification

The changes were verified using the following scenarios:

Rendered the templates using helm template to confirm correct value substitution.

Deployed the chart on a Kubernetes cluster with 3 worker nodes (kubeadm-based setup).

Verified that the global values are rendered and applied correctly across the components.

### DIFF
--- a/deploy/helm/charts/README.md
+++ b/deploy/helm/charts/README.md
@@ -67,6 +67,11 @@ The following table lists the configurable parameters of the OpenEBS ZFS Localpv
 
 | Key                                                                 | Type   | Default Value               | Description                                                                                           |
 |----------------------------------------------------------------------|--------|-----------------------------|-------------------------------------------------------------------------------------------------------|
+| `global.imageRegistry`                                                | string | `""`                        | Global override for container image registry                                                           |
+| `global.imagePullPolicy`                                              | string | `""`                        | Global override for  image pull policy.                                                            |
+| `global.imagePullSecrets`                                             | list   | `[]`                        | Global image pull secrets (merged with local imagePullSecrets and not overridden).                                                                       |
+| `global.analytics.enabled`                                            | bool   | `null`                      | Global override for analytics.                                                                         |
+| `global.kubeletDir`                                                   | string |  `""`                       | Global override for kubelet directory                                                                  |
 | `analytics.enabled`                                                  | bool   | `true`                      | Enables or disables analytics reporting for the chart.                                                 |
 | `analytics.installerType`                                             | string | `"zfs-localpv-helm"`        | Specifies the installer type for analytics.                                                           |
 | `backupGC.enabled`                                                    | bool   | `false`                     | Enables or disables garbage collection for backups.                                                   |
--- a/deploy/helm/charts/templates/_helpers.tpl
+++ b/deploy/helm/charts/templates/_helpers.tpl
@@ -158,5 +158,54 @@ Enable zfsController containers leader election if replicas > 1
 Ensure that the path to kubelet ends with a slash
 */}}
 {{- define "zfslocalpv.zfsNode.kubeletDir" -}}
-{{- printf "%s/" (.Values.zfsNode.kubeletDir | trimSuffix "/") -}}
+{{- printf "%s/" (default .Values.zfsNode.kubeletDir .Values.global.kubeletDir | trimSuffix "/") -}}
 {{- end }}
+
+{{/*
+Creates the image URL ie registry/repository:tag
+*/}}
+{{- define "zfslocalpv.common.image" -}}
+{{- $registryName := default .imageRoot.registry ((.global).imageRegistry) | trimSuffix "/" -}}
+{{- $repositoryName := .imageRoot.repository -}}
+{{- $termination := .imageRoot.tag | toString -}}
+{{- if $registryName }}
+    {{- printf "%s/%s:%s" $registryName $repositoryName $termination -}}
+{{- else -}}
+    {{- printf "%s:%s"  $repositoryName $termination -}}
+{{- end -}}
+{{- end -}}
+
+{{/*
+Concatenates imagepullsecrets and handles different formats (example - secret or - name: secret)
+*/}}
+{{- define "zfslocalpv.common.pullSecrets" -}}
+{{- $names := list -}}
+{{- with .Values.global.imagePullSecrets -}}
+  {{- range . -}}
+    {{- if kindIs "map" . }}
+      {{- if and (hasKey . "name") (not (empty .name)) -}}
+        {{ $names = append $names .name }}
+      {{- end -}}
+    {{- else if not (empty .) -}}
+      {{ $names = append $names . -}}
+    {{- end -}}
+{{- end -}}
+{{- end -}}
+{{- with .Values.imagePullSecrets -}}
+  {{- range . }}
+    {{- if kindIs "map" . -}}
+      {{- if and (hasKey . "name") (not (empty .name)) -}}
+        {{- $names = append $names .name }}
+      {{- end -}}
+    {{- else if not (empty .) -}}
+      {{- $names = append $names . -}}
+    {{- end -}}
+  {{- end -}}
+{{- end -}}
+{{- $names = uniq $names -}}
+{{- if $names -}}
+{{- range $names }}
+- name: {{ . }}
+{{- end -}}
+{{- end -}}
+{{- end -}}
--- a/deploy/helm/charts/templates/zfs-controller.yaml
+++ b/deploy/helm/charts/templates/zfs-controller.yaml
@@ -44,7 +44,7 @@ spec:
 {{- end }}
       containers:
         - name: {{ .Values.zfsController.resizer.name }}
-          image: "{{ .Values.zfsController.resizer.image.registry }}{{ .Values.zfsController.resizer.image.repository }}:{{ .Values.zfsController.resizer.image.tag }}"
+          image: {{ include "zfslocalpv.common.image" (dict "imageRoot" .Values.zfsController.resizer.image "global" .Values.global) }}
           args:
             - "--v=5"
             - "--csi-address=$(ADDRESS)"
@@ -55,15 +55,15 @@ spec:
           env:
             - name: ADDRESS
               value: /var/lib/csi/sockets/pluginproxy/csi.sock
-          imagePullPolicy: {{ .Values.zfsController.resizer.image.pullPolicy }}
+          imagePullPolicy: {{ default .Values.zfsController.resizer.image.pullPolicy .Values.global.imagePullPolicy }}
           volumeMounts:
             - name: socket-dir
               mountPath: /var/lib/csi/sockets/pluginproxy/
           resources:
             {{- toYaml .Values.zfsController.resources | nindent 12 }}
         - name: {{ .Values.zfsController.snapshotter.name }}
-          image: "{{ .Values.zfsController.snapshotter.image.registry }}{{ .Values.zfsController.snapshotter.image.repository }}:{{ .Values.zfsController.snapshotter.image.tag }}"
-          imagePullPolicy: {{ .Values.zfsController.snapshotter.image.pullPolicy }}
+          image: {{ include "zfslocalpv.common.image" (dict "imageRoot" .Values.zfsController.snapshotter.image "global" .Values.global) }}
+          imagePullPolicy: {{ default .Values.zfsController.snapshotter.image.pullPolicy .Values.global.imagePullPolicy }}
           args:
             - "--csi-address=$(ADDRESS)"
             {{- include "zfslocalpv.zfsController.leaderElection" . | indent 12 }}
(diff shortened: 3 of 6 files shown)


deploy/helm/charts/templates/zfs-controller.yaml
```diff
@@ -44,7 +44,7 @@ spec:
 {{- end }}
       containers:
         - name: {{ .Values.zfsController.resizer.name }}
-          image: "{{ .Values.zfsController.resizer.image.registry }}{{ .Values.zfsController.resizer.image.repository }}:{{ .Values.zfsController.resizer.image.tag }}"
+          image: {{ include "zfs-controller-resizer.image" .}}
```
tiagolobocastro: put space after `.`

```suggestion
          image: {{ include "zfs-controller-resizer.image" . }}
```

deploy/helm/charts/values.yaml
```diff
@@ -1,3 +1,16 @@
...
+  # set the global image registry
+  imagePullSecrets: []
+  # - secret
+  # or
+  # -name: secret
+  imagePullPolicy: ""
+  # set global image pull policy
+  analytics:
+    enabled: true
+  # disable analytics
```
tiagolobocastro: usually, comments are above

```suggestion
  # set the global image registry
  imageRegistry:
  # set global image pull secrets
  # - secret
  # or
  # -name: secret
  imagePullSecrets: []
  # set global image pull policy
  imagePullPolicy:
  analytics:
    # -- Enable OpenEBS analytics which help track engine traction and usage.
    enabled:
```

tiagolobocastro: Also I think either global.analytics needs to be empty by default ?

tiagolobocastro (commented):
Please check the analytics as I think it may be breaking

Also something to consider, on rawfile looks like we use global `imageRegistry` for only rawfile: <https://github.com/openebs/rawfile-localpv/blob/59e252ec40bdd93f19d89ba5ced166116093c2fc/deploy/helm/rawfile-localpv/values.yaml#L24>
I think this way here is better, since on air-gapped you'd probably move all images anyway, but I guess this means we may want to modify rawfile as well?

@pasha-gurkov are you perchance using `global.*Registry` ? Modifying this may be another breaking change :/

deploy/helm/charts/values.yaml
krishnaGajabi: > usually, comments are above

Done

krishnaGajabi:
> Please check the analytics as I think it may be breaking
> 
> Also something to consider, on rawfile looks like we use global `imageRegistry` for only rawfile: https://github.com/openebs/rawfile-localpv/blob/59e252ec40bdd93f19d89ba5ced166116093c2fc/deploy/helm/rawfile-localpv/values.yaml#L24 I think this way here is better, since on air-gapped you'd probably move all images anyway, but I guess this means we may want to modify rawfile as well?
> 
> @pasha-gurkov are you perchance using `global.*Registry` ? Modifying this may be another breaking change :/

Yes, It is better to have a single global variable for all imageregistry. global variable overides all local variables even if local variable is explicity set as per this change.(ie the variable is only usefull for airgapped environments)

Also i checked the analytics setting on rawfile, it only uses global variable to set them ?
since for zfs we were earlier using local variables to set the value. if we decide to go with rawfile way it would disturbing the old settings.(this change supports both local and global values)

deploy/helm/charts/templates/_helpers.tpl
```diff
@@ -160,3 +160,63 @@ Ensure that the path to kubelet ends with a slash
...
+{{ include "common.images.image" (dict "imageRoot" .Values.zfsController.provisioner.image "global" .Values.global) }}
+{{- end -}}
+
+{{- define "zfsPlugin.image" -}}
+{{ include "common.images.image" (dict "imageRoot" .Values.zfsPlugin.image "global" .Values.global) }}
+{{- end -}}
+
+{{- define "zfsNode.driverRegistrar.image" -}}
+{{ include "common.images.image" (dict "imageRoot" .Values.zfsNode.driverRegistrar.image "global" .Values.global) }}
+{{- end -}}
```
niladrih: This should be one single template helper, and it should be supplied with an argument for the image value key.

niladrih (commented):
The names for the helpers should start with a `zfslocalpv.` like it is with the other helpers. This is done so that on the openebs chart, when the helpers from all of the subcharts are accumulated on to the same context, they don't have any naming conflicts.

deploy/helm/charts/templates/_helpers.tpl
krishnaGajabi: Changes Made:
Renamed common helper function
removed individual helper functions for image url
instead of seperate helper function for each value, pasted the logic directly in templates

tiagolobocastro:
> > Please check the analytics as I think it may be breaking
> > Also something to consider, on rawfile looks like we use global `imageRegistry` for only rawfile: https://github.com/openebs/rawfile-localpv/blob/59e252ec40bdd93f19d89ba5ced166116093c2fc/deploy/helm/rawfile-localpv/values.yaml#L24 I think this way here is better, since on air-gapped you'd probably move all images anyway, but I guess this means we may want to modify rawfile as well?
> > @pasha-gurkov are you perchance using `global.*Registry` ? Modifying this may be another breaking change :/
> 
> Yes, It is better to have a single global variable for all imageregistry. global variable overides all local variables even if local variable is explicity set as per this change.(ie the variable is only usefull for airgapped environments)
> 
> Also i checked the analytics setting on rawfile, it only uses global variable to set them ? since for zfs we were earlier using local variables to set the value. if we decide to go with rawfile way it would disturbing the old settings.(this change supports both local and global values)

My preference would be to modify `rawfile`, moving the rawfile's own image registry to `image.registry` and have `global.imageRegistry` modify all containers and not just rawfile own images.

Any thoughts @pasha-gurkov ? We'd have to move to `v0.14` as it'd be a breaking change

deploy/helm/charts/values.yaml
```diff
@@ -1,3 +1,16 @@
+global:
+  # set the global image registry
+  imageRegistry: ""
+  # - secret
+  # or
+  # -name: secret
```
niladrih: `name: secret` is left on the template function for backwards compatibility, let's not advertise this way of using the value in the comment.

deploy/helm/charts/README.md
```diff
@@ -67,6 +67,10 @@ The following table lists the configurable parameters of the OpenEBS ZFS Localpv
 
 | Key                                                                 | Type   | Default Value               | Description                                                                                           |
 |----------------------------------------------------------------------|--------|-----------------------------|-------------------------------------------------------------------------------------------------------|
+| `global.imageRegistry`                                                  | string   | `""`                      | set global image registry.
+| `global.imagePullPolicy`                                                  | string   | `""`                      | set global image pull policy.
+| `global.imagePullSecrets`                                                  | list   | `[]`                      | set global image pull secrets.
+| `global.analytics.enabled`                                                  | bool   | `true`                      | Enables or disables analytics reporting for the chart.
```
niladrih: Let's please align these to the same level as the existing table cells.
```suggestion
| `global.imageRegistry`                                                              | string   | `""`                      | set global image registry.
| `global.imagePullPolicy`                                                            | string   | `""`                      | set global image pull policy.
| `global.imagePullSecrets`                                                          | list   | `[]`                      | set global image pull secrets.
| `global.analytics.enabled`                                                        | bool   | `true`                      | Enables or disables analytics reporting for the chart.
```

deploy/helm/charts/templates/zfs-controller.yaml
```diff
@@ -135,7 +135,11 @@ spec:
             - name: OPENEBS_IO_INSTALLER_TYPE
               value: "{{ if (not (hasKey .Values.analytics "installerType")) }}zfs-localpv-helm{{ else }}{{ .Values.analytics.installerType }}{{ end }}"
             - name: OPENEBS_IO_ENABLE_ANALYTICS
+              {{- if not (eq (dig "analytics" "enabled" nil .Values.global) nil) }}
+              value: "{{ .Values.global.analytics.enabled }}"
+              {{- else }}
               value: "{{ .Values.analytics.enabled }}"
+              {{- end }}
             {{- if .Values.analytics.gaId }}
```
niladrih: This is not the correct sequence. If local values are configured, then we must prefer the local one. If they are `nil`, we go with the global option.

deploy/helm/charts/templates/_helpers.tpl
```diff
@@ -160,3 +160,44 @@ Ensure that the path to kubelet ends with a slash
 {{- define "zfslocalpv.zfsNode.kubeletDir" -}}
 {{- printf "%s/" (.Values.zfsNode.kubeletDir | trimSuffix "/") -}}
 {{- end }}
+
+{{/*
+Creates the image URL ie registry/repository:tag
+*/}}
+{{- define "zfslocalpv.common.image" -}}
+{{- $registryName := default .imageRoot.registry ((.global).imageRegistry) | trimSuffix "/" -}}
```
niladrih, line 168: This is not the right order for how we plan to use the global values. The global has to be the fallback and the local has to be preferred. If the local is `nil` or empty, then we prefer the global. You'd also have to remove the values from the existing image related yaml keys from the values.yaml and only populate the global fields.

tiagolobocastro: hmm doesn't seem to be the case with bitnami global, at least:

```
## @section Global parameters
## Global Docker image parameters
## Please, note that this will override the image parameters, including dependencies, configured to use the global value
## Current available global Docker image parameters: imageRegistry, imagePullSecrets and storageClass
##
```

If you see the image.registry is set to `docker.io` and gets overridden if you define the global

deploy/helm/charts/README.md
krishnaGajabi: Done

deploy/helm/charts/templates/_helpers.tpl
krishnaGajabi: Kept the changes as it is, Please let me know if any modification is required.
local is default and global overrides, global values are empty and local values have default values

deploy/helm/charts/templates/zfs-controller.yaml
```diff
@@ -44,7 +44,7 @@ spec:
 {{- end }}
       containers:
         - name: {{ .Values.zfsController.resizer.name }}
-          image: "{{ .Values.zfsController.resizer.image.registry }}{{ .Values.zfsController.resizer.image.repository }}:{{ .Values.zfsController.resizer.image.tag }}"
+          image: {{ include "zfslocalpv.common.image" (dict "imageRoot" .Values.zfsController.resizer.image "global" .Values.global) }}
```
niladrih, line 47: This maybe can be improved further? The global is already known, we need not add it as an input in every template call. `"global" .Values.global` can be hardcoded inside the template function.

niladrih (commented):
Then names zfslocalpv.common don't make sense in our context, we don't have a common context. Maybe zfslocalpv.image is enough?

deploy/helm/charts/templates/zfs-controller.yaml
niladrih: The global values are overrides in most cases. The ordering is correct as is.

deploy/helm/charts/templates/zfs-controller.yaml
krishnaGajabi: yes but the dict should have values from current context. so it will be tricky

krishnaGajabi:
> Then names zfslocalpv.common don't make sense in our context, we don't have a common context. Maybe zfslocalpv.image is enough?

we have helper functions with names like 'zfslocalpv.zfsController' 'zfslocalpv.zfsNode', since this helper function is common for both node and controller kept the name as 'zfslocalpv.common'.
here is a function where we used same namin convention > https://github.com/openebs/zfs-localpv/blob/develop/deploy/helm/charts/templates/_helpers.tpl#L60
I can change it if its really necessary?

deploy/helm/charts/values.yaml
```diff
@@ -1,8 +1,19 @@
+global:
+  # Global override for container image registry.
+  imageRegistry: ""
+  # - secret
+  imagePullSecrets: []
+  # Global overide for image pull policy
+  imagePullPolicy: ""
+  analytics:
+    # Global overide for analytics
```
tiagolobocastro:
```suggestion
    # Global override for analytics
```

deploy/helm/charts/values.yaml
```diff
@@ -1,8 +1,19 @@
+global:
+  # Global override for container image registry.
+  imageRegistry: ""
+  # - secret
+  imagePullSecrets: []
+  # Global overide for image pull policy
```
tiagolobocastro:
```suggestion
  # Global override for image pull policy
```

deploy/helm/charts/values.yaml
```diff
@@ -1,8 +1,19 @@
...
+  analytics:
+    # Global overide for analytics
+    enabled:
+
 # Default values for openebs-zfslocalpv.
 # This is a YAML-formatted file.
 # Declare variables to be passed into your templates.
 imagePullSecrets:
-# - name: "image-pull-secret"
+# - image-pull-secret
```
tiagolobocastro: nit: just to match the example given on global

```suggestion
# - secret
```

deploy/helm/charts/values.yaml
```diff
@@ -1,8 +1,19 @@
+global:
```
tiagolobocastro, line 5: let's add the `kubeletDir` as well?

tiagolobocastro: Anything else that might be useful on global? CC @niladrih

tiagolobocastro (commented):
@krishnaGajabi could you please add an issue on openebs to track all the required changes? thank you

krishnaGajabi:
> @krishnaGajabi could you please add an issue on openebs to track all the required changes? thank you

[Issue: ](https://github.com/openebs/openebs/issues/4176). >> done

deploy/helm/charts/values.yaml
krishnaGajabi: Done

deploy/helm/charts/values.yaml
krishnaGajabi: Done

deploy/helm/charts/values.yaml
krishnaGajabi: > let's add the `kubeletDir` as well?

Added

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

tiagolobocastro:
@mergify queue

Abhinandan-Purkait:
@mergifyio queue

PR #702, opened 2026-03-14
feat: support cloning to different paths on same pool

### LINKED ISSUE #701: Support cloning from one dataset/root on a pool to a different dataset/root on the same pool
Currently [there's a strict equality check done](https://github.com/openebs/zfs-localpv/blob/3192ca432d3b0bb6abbbdd30273040743b6d911c/pkg/driver/controller.go#L416) but I think as far as ZFS is concerned only the root pool has to be the same.

The check should compare the pool, i.e. whatever is before the slash.

michaelbeaumont:
## Pull Request template

Fixes https://github.com/openebs/zfs-localpv/issues/701

**What this PR does?**:

Makes the source/destination check more flexible.

No.

To be fair I haven't tested this but as far as I can tell this is supported by `zfs clone`.

**Checklist:**
- [x] Fixes #701
- [x] Commit has unit tests

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -339,36 +339,36 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 }
 
 // CreateVolClone creates the clone from a volume
-func CreateVolClone(ctx context.Context, req *csi.CreateVolumeRequest, srcVol string) (string, error) {
+func CreateVolClone(ctx context.Context, req *csi.CreateVolumeRequest, srcVolName string) (string, error) {
 	volName := strings.ToLower(req.GetName())
 	parameters := req.GetParameters()
 	// lower case keys, cf CreateZFSVolume()
-	pool := helpers.GetInsensitiveParameter(&parameters, "poolname")
+	destPoolName := helpers.GetInsensitiveParameter(&parameters, "poolname")
 	size := getRoundedCapacity(req.GetCapacityRange().RequiredBytes)
 	volsize := strconv.FormatInt(int64(size), 10)
 
-	vol, err := zfs.GetZFSVolume(srcVol)
+	srcVol, err := zfs.GetZFSVolume(srcVolName)
 	if err != nil {
 		return "", status.Error(codes.NotFound, err.Error())
 	}
 
-	if !isSamePool(vol.Spec.PoolName, pool) {
+	if !isSamePool(srcVol.Spec.PoolName, destPoolName) {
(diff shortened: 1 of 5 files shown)


pkg/driver/controller.go
Abhinandan-Purkait: Where is the new path getting updated? I can see that we just passed the old spec into the new volume. So that would just create on the src volume's pool right? The different destination would just be ignored.

I see you have not tested this, I think this needs to be tested, this might not work as expected.

The check would pass (same root pool data), but then line 379 (volObj.Spec = vol.Spec) copies the source's PoolName (data/sub1) to the new volume — ignoring the requested data/sub2         
  entirely. The clone ends up in the source's dataset, not the one the user asked for.

Abhinandan-Purkait (changes requested):

pkg/driver/controller.go
michaelbeaumont: Yeah it copies the whole spec in. Changed that. I'll write an e2e test for this

pkg/driver/controller.go
```diff
@@ -579,6 +581,12 @@ func (cs *controller) DeleteVolume(
 	return csipayload.NewDeleteVolumeResponseBuilder().Build(), nil
 }
 
+func isSamePool(firstLocation, secondLocation string) bool {
+	firstPool, _, _ := strings.Cut(firstLocation, "/")
+	secondPool, _, _ := strings.Cut(secondLocation, "/")
+	return firstPool == secondPool
+}
```
Abhinandan-Purkait, line 588: Nit: 2 empty strings yield true? Might not happen ever but still could be good to check that.

Abhinandan-Purkait (commented):
LGTM. Thanks @michaelbeaumont.  Let me know one the tests are added, we can approve and take this in.

tiagolobocastro:
@michaelbeaumont the tests are still failing, could you please take a look?

michaelbeaumont:
> @michaelbeaumont the tests are still failing, could you please take a look?

Sorry, I'm having trouble running the tests locally so realistically I don't think I'll have time for this in the near future.

tiagolobocastro:
> > @michaelbeaumont the tests are still failing, could you please take a look?
> 
> Sorry, I'm having trouble running the tests locally so realistically I don't think I'll have time for this in the near future.

You can use the vm with zfs: https://github.com/openebs/zfs-localpv/blob/develop/docs/developer-setup.md#nixos-shell
Sure, let us know if you get some time and need help, thank you

Issue #540, opened 2024-05-31
Refactor the zfs localpv product logging and error propagation


Abhinandan-Purkait:
### Issues

1. The logging lacks verbosity levels.
2. There error propagation is not very well, it's difficult to track an error.

This issue was uncovered on #536.

Abhinandan-Purkait:
Has been released in v2.10.0

PR #710, opened 2026-05-11
refactor(zfs): replace error classification with verbatim stderr propagation

Abhinandan-Purkait:
- ZFS operation failures now surface the exact CLI error via Warning events on the CR — visible with `kubectl describe zfsvolume/zfssnapshot`
  - All `zfs`/`zpool` exec failures are wrapped with operation and dataset context so logs are never vague
  - Exec verbosity controllable via `--v=4` (commands) and `--v=5` (output on failure)

### DIFF
--- a/pkg/zfs/zfs_util.go
+++ b/pkg/zfs/zfs_util.go
@@ -62,6 +62,18 @@ const (
 	VolTypeZVol    = "ZVOL"
 )
 
+// runCmd executes cmd, logs the invocation at V(4) and full output at V(5),
+// and returns combined stdout+stderr. Use this instead of cmd.CombinedOutput()
+// throughout this package so verbosity can be tuned with --v.
+func runCmd(cmd *exec.Cmd, dataset string) ([]byte, error) {
+	klog.V(4).Infof("zfs: executing %v on %q", cmd.Args, dataset)
+	out, err := cmd.CombinedOutput()
+	if err != nil {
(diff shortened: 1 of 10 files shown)


Abhinandan-Purkait:
Resolves #540

pkg/zfs/zfs_util.go
```diff
@@ -62,6 +62,18 @@ const (
 	VolTypeZVol    = "ZVOL"
 )
 
+// runCmd executes cmd, logs the invocation at V(4) and full output at V(5),
+// and returns combined stdout+stderr. Use this instead of cmd.CombinedOutput()
+// throughout this package so verbosity can be tuned with --v.
+func runCmd(cmd *exec.Cmd) ([]byte, error) {
+	klog.V(4).Infof("zfs: executing %v", cmd.Args)
```
tiagolobocastro: Can we tie this to a specific volume?

tiagolobocastro (approved):

niladrih (commented):
Events need to be truncated to the max permissible length for a kubernetes event. Here's a naive approach for this. Curious if there is a better way to not miss out on important keywords.
https://github.com/openebs/mayastor-control-plane/blob/develop/k8s/operators/src/pool/context.rs#L395

niladrih (approved):
LGTM

Abhinandan-Purkait:
@mergifyio queue

PR #718, opened 2026-05-14
ci: fix the csi driver and chart releaser

Abhinandan-Purkait:
- The dockerfile points to a wrong location.
- The chart releaser version was wrong.


tiagolobocastro (approved):

Abhinandan-Purkait:
@mergifyio queue

PR #719, opened 2026-05-14
ci: point to the correct version of chart releaser

Abhinandan-Purkait:
- Fix helm chart releaser version on release workflow


tiagolobocastro (approved):

Abhinandan-Purkait:
@merigifyio queue

PR #723, opened 2026-06-03
feat: support atime and logbias StorageClass parameters

firecow:
## Why

`atime` and `logbias` can be set in StorageClass `parameters` but are silently dropped — volumes always get the ZFS defaults (`atime=on`, `logbias=latency`). `logbias=throughput` in particular matters for write-heavy workloads on pools without a SLOG.

## What

Wire both through the same path as `compression`/`dedup`/`recordsize`: API field, volbuilder, controller param parsing, and the create/clone/set/recv arg builders.

- `atime` is filesystem-only → gated to `DATASET` (skipped for zvols).
- `logbias` applies to both datasets and zvols.
- Both editable post-creation (`PropertyChanged` + set args).
- CRDs regenerated, docs updated, unit tests added.

Possible follow-up (not in this PR): the per-property arg blocks could be collapsed into a helper.

### DIFF
--- a/deploy/helm/charts/charts/crds/templates/zfsrestore.yaml
+++ b/deploy/helm/charts/charts/crds/templates/zfsrestore.yaml
@@ -81,6 +81,18 @@ spec:
               In case of Cloned volumes, the parameters are assigned the same values
               as the source volume.
             properties:
+              atime:
+                description: |-
+                  ATime controls whether the access time for files is updated when they are read.
+                  Turning this property off ("off") avoids producing write traffic when reading files,
+                  which can result in significant performance gains for read-heavy workloads.
+                  This is a filesystem (dataset) only property and is ignored for zvols.
+                  ATime property can be edited after the volume has been created.
+                  Default Value: on.
+                enum:
+                - "on"
+                - "off"
+                type: string
               capacity:
                 description: Capacity of the volume
                 minLength: 1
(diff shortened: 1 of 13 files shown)


Abhinandan-Purkait:
Unrelated to your changes. I think the storageclass way of setting properties is not a good idea overall. We should rather do it via annotations on the PVC and then the controller just sets it up on the ZFSVolume resource. That way the properties are limited to that specific volume, rather than having one storageclass for a fixed set of properties. 

Either way the changes on this PR looks good to me.

@tiagolobocastro @niladrih Thoughts + reviews please.

firecow:
> Unrelated to your changes. I think the storageclass way of setting properties is not a good idea overall. We should rather do it via annotations on the PVC and then the controller just sets it up on the ZFSVolume resource. That way the properties are limited to that specific volume, rather than having one storageclass for a fixed set of properties.
> 
> Either way the changes on this PR looks good to me.
> 
> @tiagolobocastro @niladrih Thoughts + reviews please.

I just fixed the two copilot findings.

This PR is intentionally scoped to match the existing SC properties (recordsize/compression/dedup) so it stays consistent with the current model.

Abhinandan-Purkait:
@firecow Thanks! We will review this again.

tiagolobocastro (approved):

Abhinandan-Purkait (approved):

Abhinandan-Purkait:
@mergifyio squash

Abhinandan-Purkait:
@mergifyio queue

PR #732, opened 2026-07-03
test(zfs-localpv): Add ci test for shared volume

krishnaGajabi:
Adds ci test for shared volume

### DIFF
--- a/tests/utils.go
+++ b/tests/utils.go
@@ -686,6 +686,29 @@ func verifyAppPodRunning(appname string) {
 	gomega.Expect(status).To(gomega.Equal(true), "while checking status of pod {%s}", appPod.Items[0].Name)
 }
 
+func verifyAppPodNotRunning(appname string) {
+	var err error
+	gomega.Eventually(func() bool {
+		appPod, err = PodClient.WithNamespace(OpenEBSNamespace).
+			List(metav1.ListOptions{LabelSelector: "role=test,appName=" + appname})
+		gomega.Expect(err).ShouldNot(gomega.HaveOccurred())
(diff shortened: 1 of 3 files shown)


tests/utils.go
```diff
@@ -809,6 +809,28 @@ func getStoragClassParams() []map[string]string {
...
+			"fstype": "xfs",
+			"shared": "yes",
+		},
+		{
+			"fstype": "btrfs",
+			"shared": "yes",
+		},
+	}
+
+}
```
Abhinandan-Purkait, line 859: Nice, let's also add a failure scenario, where, the mount should fail without shared param. The failure should be expected

tiagolobocastro: yeah that should IIRC
also do you know why "shared" even exists, was it to kind of simulate `ReadWriteOncePod` ?

Abhinandan-Purkait: Yeah, I think this was done before RWOP came. But now it just makes sense to use RWO and RWOP.

tiagolobocastro: Maybe we can deprecate this shared from the storageClass parameters
And if not present we take it from the access mode? This should make it a non-breaking change?

krishnaGajabi: Added negative test case scenario(`(shared=no)`)

Abhinandan-Purkait (approved):

tiagolobocastro (approved):

tests/utils.go
Abhinandan-Purkait: Yeah that's a good idea.

Abhinandan-Purkait:
@mergifyio queue

PR #735, opened 2026-07-14
oep(localpv): add design proposal for poolpattern-based volume provisioning

### LINKED ISSUE #403: Multiple zpools per node
I need to dynamically provision zfs-backed volumes without redundancy (just single zfs dataset per physical disk). Though it could be simply accomplished with zfs-provisioner on a different nodes, I cannot have more than one volume on a single node, because of StorageClass needs a poolname parameter. So that StorageClass is limited to only one zpool per node.

Trying to set up multiple zpools on a every single node to provision zfs persistent volume on every single drive without redundancy. 

Support multiple zpools per StorageClass. Allow a list of poolnames in a StorageClass

**Environment:**
- ZFS-LocalPV version: latest
- Kubernetes version: v1.19.2
- Hardware configuration: 6 baremetal machines with 4 HDD Drives each.
- OS: Centos 7.8.2003

krishnaGajabi:
Why is this PR required? What issue does it fix?:

Fixes #403. Today a LocalPV-ZFS StorageClass must name one exact ZFS pool via poolname, which limits provisioning to a single pool per node. Issue #403 describes the real-world pain: a cluster of baremetal nodes each with several independent drives (one zpool per drive, no redundancy) cannot be served by a single StorageClass — every drive needs its own StorageClass pinned to that exact pool name.

This proposal introduces a poolpattern parameter — a regular expression that lets one StorageClass span a family of pools (e.g. zfspv-pool.* or tank[0-9]+) and lets the driver choose among the matching pools on each node. It is the ZFS analog of lvm-localpv's vgpattern (a ZFS pool ↔ an LVM volume group).

What this PR does?:

Adds the design document design/poolpattern.md (OEP) describing the feature. It covers:

The poolpattern StorageClass API and its precedence with poolname (poolname wins if both set; error if neither).
Pool resolution & scheduling: reuse the existing node scheduler's weight metric (VolumeWeighted/CapacityWeighted) to pick among matching pools, with free capacity used only as a fit filter.
The fit filter being thick-only (skipped for thin volumes, where overcommit is intended).
Clone/snapshot path behavior (source pool must match the regex in pattern mode).
Validation, implementation plan, and test plan.
This is a design-only PR — no production code changes are included. Implementation will follow in subsequent PR(s).

Does this PR require any upgrade changes?:

No. The proposal introduces no CRD changes — StorageClass parameters are free-form and the resolved concrete pool is already persisted per-volume in ZFSVolume.Spec.PoolName. (This is a documentation-only PR regardless.)

If the changes in this PR are manually verified, list down the scenarios covered::

N/A — documentation only. No runtime behavior changes to verify.

Any additional information for your reviewer?:

This is a design proposal (OEP) intended to align on the approach before implementation, and it addresses the multi-pool-per-node request in #403.  Feedback on the scheduling/fit-filter policy and the thin-vs-thick handling is especially welcome, as those are the main departures from a naive "largest free capacity" approach.

### DIFF
--- a/design/poolpattern.md
+++ b/design/poolpattern.md
@@ -1,0 +1,179 @@
+---
+title: Pool Pattern based Volume Provisioning for LocalPV-ZFS
+authors:
+  - "@krishna-gajabi"
+creation-date: 2026-07-13
+last-updated: 2026-07-13
+status: proposed
+---
+
+# Pool Pattern (`poolpattern`) based Volume Provisioning
+
+## Table of Contents
+
+* [Summary](#summary)
+* [Motivation](#motivation)
+* [Current Behaviour](#current-behaviour)
+* [Proposal](#proposal)
+  * [StorageClass API](#storageclass-api)
+  * [Pool Resolution and Scheduling](#pool-resolution-and-scheduling)
+  * [Clone / Snapshot Paths](#clone--snapshot-paths)
+  * [Validation](#validation)
+* [Implementation Plan](#implementation-plan)
+* [Test Plan](#test-plan)
+* [Docs](#docs)
+* [Resolved Decisions](#resolved-decisions)
+
+## Summary
+
+This proposal adds a `poolpattern` StorageClass parameter to LocalPV-ZFS. It is a
+regular expression that selects the ZFS pool to provision into at CreateVolume
+time, as an alternative to the existing fixed `poolname` parameter. This mirrors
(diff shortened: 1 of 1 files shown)


Abhinandan-Purkait (commented):
Looks good to me at a glance. I would say make sure the existing behavior doesn't change the pool pattern related calculations and considerations should stictly not impact the pool name way. Also please transfer this to the openebs/openebs.

krishnaGajabi:
https://github.com/openebs/openebs/issues/4268
Moved it to OpenEBS repo

PR #738, opened 2026-07-27
test(bdd): add bdd spec for poolpattern based volume provisioning

krishnaGajabi:

The `poolpattern` StorageClass parameter (design: OEP 4268, `designs/local-pv/zfs/poolpattern.md`
in openebs/openebs) lets one StorageClass select a ZFS pool by regex instead of a fixed
`poolname`. Before the implementation lands, this PR pins down what the feature must actually do
from a user's point of view, so the behaviour is reviewed as a contract rather than inferred from
the implementation afterwards.

It also captures behaviour the design changes on the **existing** `poolname` path (`CapacityWeighted`
ordering moving to the pool's real `Used`, and a free-space fit filter that now fails fast), so those
changes are visible to reviewers who only care about today's behaviour.

Fixes #<issue number>

**What this PR does?**:

Adds tests/bdd/poolname_poolpattern.feature, the BDD specification for the poolpattern StorageClass parameter proposed in OEP 4268 (openebs/openebs#4269).

provisioning against a matching pattern across fstypes, and a fixed poolname carrying a dataset path
exactly one of poolname / poolpattern must be set — both-set and neither-set are rejected with InvalidArgument
the fit filter for space-reserving volumes and its capacity-aware fail-fast (ResourceExhausted), including that thin zvols and quota-only datasets bypass it, and that a pending volume binds once space is freed
snapshot, clone, restore and resize under a pattern StorageClass, including that a clone inherits its source pool rather than being resolved again
the three scheduling algorithms: VolumeWeighted, CapacityWeighted and the new SpaceWeighted

No — this PR adds a spec file only.

The feature it specifies does change existing behaviour on upgrade (`CapacityWeighted` ordering shifts
from driver-provisioned sum to the pool's real `Used`, and a full or absent pool now returns
`ResourceExhausted`/`FailedPrecondition` instead of failing at `zfs create` and retrying). That needs
a release note when the implementation lands; it is recorded under Resolved Decisions in the OEP.

None — nothing in this PR executes. Two prerequisites are needed before these scenarios can be
implemented, and neither is included here:

1. `ci/ci-test.sh` currently creates a single 100G pool named `zfspv-pool`. The spec needs two pools
   of different and small sizes (a few GiB), so that "largest free space" is deterministic and "does
   not fit" is reachable with a modest request. `zfspv-pool-a` should be sized comfortably larger than
   `zfspv-pool-b`, since volumes accumulate in it as the suite runs.
2. A `HasPVCEventEventually(pvcName, reason, substr)` helper in `tests/utils.go`. The suite has no
   events client today, and two scenarios assert a `ProvisioningFailed` event reporting a specific
   gRPC code.

Part of the `poolpattern` design (OEP 4268). The implementation PR follows and will carry the ginkgo
tests for these scenarios, the unit tests listed above, and the CI pool changes.

**Checklist:**
- [ ] Fixes #<issue number>
- [x] PR Title follows the convention of `<type>(<scope>): <subject>`
- [ ] Has the change log section been updated? — n/a, no user-facing change in this PR
- [ ] Commit has unit tests — n/a, spec only; unit tests land with the implementation PR
- [ ] Commit has integration tests — n/a, this PR *is* the integration test spec; the ginkgo
      implementation lands with the feature PR
      not applicable to a spec file; the `poolname` behaviour change is tracked in the OEP
- [ ] (Optional) If documentation changes are required, which issue on https://github.com/openebs/website
      is used to track them: none for this PR; `docs/storageclasses.md` and `docs/scheduler.md` are
      updated in the implementation PR

### DIFF
--- a/tests/bdd/poolname_poolpattern.feature
+++ b/tests/bdd/poolname_poolpattern.feature
@@ -1,0 +1,209 @@
+Feature: Validate pool pattern (poolpattern) based volume provisioning
+         Validate thick fit filter and capacity-aware fail-fast
+         Validate snapshot, restore and clone under a poolpattern storage class
+         Validate the VolumeWeighted, CapacityWeighted and SpaceWeighted scheduling algorithms
+
+  Background:
+    Given a single zfs cluster is configured
+    And the node advertises its pools through the ZFSNode resource
+    And pools zfspv-pool-a and zfspv-pool-b are created on the node
+    And zfspv-pool-a has more free space than zfspv-pool-b
+    And every storage class is created with volumeBindingMode as Immediate unless stated otherwise
+
+  #########################################################################################################
+  ################## Provisioning with poolpattern ########################################################
+  #########################################################################################################
+
+  Scenario : Volume provision using a matching poolpattern
+    Given a storage class is created with poolpattern as "zfspv-pool.*"
+    When pvc is created referencing this same storage class
+    Then pvc should be in bound state
+    And zfsvolume must be created and should be in Ready state to use by any application
+    And the zfsvolume PoolName must be zfspv-pool-a, the matching pool with the largest free space
+
+  Scenario: Fixed poolname with a pool dataset path
+    Given a storage class is created with poolname as zfspv-pool-a/k8s/localpv
+    When pvc is created referencing this same storage class
+    Then pvc should be in bound state
+    And the zfsvolume PoolName must be exactly zfspv-pool-a/k8s/localpv
+    And the volume must be created under the dataset zfspv-pool-a/k8s/localpv on the node
+
+  #########################################################################################################
+  ################## poolname and poolpattern combinations ################################################
+  #########################################################################################################
+
+  Scenario Outline: A storage class must set exactly one of poolname and poolpattern
+    Given a storage class is created with poolname as <poolname> and poolpattern as <poolpattern>
+    When pvc is created referencing this same storage class
+    Then the pvc must remain in pending state
+    And the zfsvolume must not be created
+    And a ProvisioningFailed event on the pvc must report InvalidArgument
+    Examples:
+      | poolname      | poolpattern    |
+      | zfspv-pool-a  | "zfspv-pool.*" |
+      | not specified | not specified  |
+
+  #########################################################################################################
+  ################## Thick fit filter and capacity-aware fail-fast ########################################
+  #########################################################################################################
+
+  Scenario Outline: Reserving volume must fit in a single matching pool
+    Given a storage class is created with poolpattern as "zfspv-pool.*" and fsType as <fsType> and thinprovision as <thinprovision>
+    And every matching pool on the node has less free space than the requested volume size
+    When pvc is created referencing this same storage class
+    Then the pvc must remain in pending state
+    And the zfsvolume must not be created
+    And a ProvisioningFailed event on the pvc must report ResourceExhausted
+    Examples:
+      | fsType | thinprovision |
+      |  ext4  |               |
+      |  zfs   |    no         |
+
+  Scenario Outline: Non reserving volume bypasses the fit filter
+    Given a storage class is created with poolpattern as "zfspv-pool.*" and fsType as <fsType> and thinprovision as <thinprovision> and quotatype as <quotatype>
+    And every matching pool on the node has less free space than the requested volume size
+    When pvc is created referencing this same storage class
+    Then pvc should be in bound state
+    And zfsvolume must be created and should be in Ready state to use by any application
+    Examples:
+      | fsType | thinprovision | quotatype |
+      |  ext4  |    yes        |           |
+      |  zfs   |    yes        |           |
+      |  zfs   |               |           |
+      |  zfs   |               | refquota  |
+
+  Scenario: A reserving volume is placed on the matching pool that has enough free space
+    Given a storage class is created with poolpattern as "zfspv-pool.*" and thinprovision as no
+    And one matching pool on the node has enough free space and the other does not
+    When pvc is created referencing this same storage class
+    Then pvc should be in bound state
+    And the zfsvolume PoolName must be the matching pool that has enough free space
+
+  Scenario: A pending reserving volume binds once a matching pool has enough free space
+    Given a storage class is created with poolpattern as "zfspv-pool.*" and thinprovision as no
+    And every matching pool on the node has less free space than the requested volume size
+    When pvc is created referencing this same storage class
+    Then the pvc must remain in pending state
+    And the zfsvolume must not be created
+    When free space is made available on a matching pool
+    Then pvc should be in bound state
+    And the zfsvolume must be created in the pool that now has enough free space
+
+  #########################################################################################################
+  ################## Snapshot, restore, clone and resize under a poolpattern storage class ########################
+  #########################################################################################################
+
+  Scenario : Create a snapshot of a pattern provisioned volume
+    Given a storage class is created with poolpattern as "zfspv-pool.*"
+    And pvc is created referencing this storage class and a deployment using the same pvc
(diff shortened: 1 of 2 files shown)


tests/bdd/poolname_poolpattern.feature
```diff
@@ -0,0 +1,199 @@
...
+      |  ext4  |
+      |  xfs   |
+      |  btrfs |
+
+  Scenario: Fixed poolname with a pool dataset path
+    Given a storage class is created with poolname as zfspv-pool-a/k8s/localpv
+    When pvc is created referencing this same storage class
+    Then pvc should be in bound state
+    And the zfsvolume PoolName must be exactly zfspv-pool-a/k8s/localpv
+    And the volume must be created under the dataset zfspv-pool-a/k8s/localpv on the node
```
Abhinandan-Purkait, line 29: What has this to do with pool pattern?

Abhinandan-Purkait (approved):

tests/bdd/poolname_poolpattern.feature
krishnaGajabi: poolname path allows the dataset path to be specified as a pool. Since it only looks at zfsvolume CR . This is going to be allowed after this feature also. It is not allowed for poolpattern since it only looks at the root since zfsnode cr only adevertises pool root. So no happy path changes are done with this feature.

Abhinandan-Purkait: But this is already an exisitng feature

krishnaGajabi: Agreed, it's not new — a pool/dataset poolname has been supported and documented for a while (and was fixed once for GetCapacity in #393). This is regression coverage, not a feature claim: the PR rewrites the code path under it.

The pool parameter used to be compared verbatim (zv.Spec.PoolName == pool). Now it's parsed to its pool root and matched as a regex against ZFSNode.Pools, which only advertises pool-level names — so zfspv-pool-a/k8s/localpv matches as ^zfspv-pool-a$, while the full path still has to reach Spec.PoolName for the create. Keeping poolname off resolvePool is what preserves that; a later refactor routing it through "for consistency" would silently place volumes in zfspv-pool-a instead. Unit tests cover poolRoot and the pattern compilation, but nothing asserts end to end that the dataset path survives — and the suite only ever uses the flat POOLNAME = "zfspv-pool".

The limitation to not use dataset with poolpattern is documented in OEP as well.

Happy to relabel it as a regression scenario, or split it out if you'd rather keep this PR to the new parameter.

tests/bdd/poolname_poolpattern.feature
tiagolobocastro, line 1: nit: rename file to `poolname-poolpattern.feature`
it's already in bdd

tests/bdd/poolname_poolpattern.feature
```diff
@@ -0,0 +1,199 @@
...
+  #########################################################################################################
+  ################## Provisioning with poolpattern ########################################################
+  #########################################################################################################
+
+  Scenario Outline: Volume provision using a matching poolpattern
+    Given a storage class is created with poolpattern as "zfspv-pool.*" and fsType as <fsType>
+    When pvc is created referencing this same storage class
+    Then pvc should be in bound state
+    And zfsvolume must be created and should be in Ready state to use by any application
+    And the zfsvolume PoolName must be zfspv-pool-a, the matching pool with the largest free space
```
tiagolobocastro, line 22: Doesn't this depend on the scheduling algorithm?

krishnaGajabi: Node selection supports two algorithms, both of which order nodes only:
 - CapacityWeighted (default when scheduler is unset) — weights each node by the used capacity of its pools matching the pattern, and prefers the least-weighted node. Note this is least-used, not most-free: the two differ when pools vary in total size.
 - VolumeWeighted — weights each node by the number of existing ZFSVolumes whose pool matches the pattern, and prefers the node with the fewest.

Once a node is selected, the pool is chosen independently of the algorithm: the matching pool on that node with the most free space wins. The scheduler parameter never enters pool selection.
Since this suite runs on a single-node cluster, there is only one candidate node and the choice of algorithm has no observable effect — so there is no scenario for it here. VolumeWeighted node ordering is covered by the multi-node e2e scenario below ("VolumeWeighted scheduling orders nodes by volume count"), and the pool-level selection is covered by TestPoolForNode in pkg/driver/schd_helper_test.go.

krishnaGajabi: I have added bdds for scheduling algorithm including the newly added space weighted algorithm to e2e tests for multi clusters

tests/bdd/poolname_poolpattern.feature
```diff
@@ -0,0 +1,238 @@
...
+    Given a multi node cluster where each node carries a differently named pool matching the pattern
+    And a storage class is created with poolpattern as "zfspv-pool.*" and fsType as <fsType>
+    When several pvcs are created referencing this same storage class
+    Then each pvc must be bound and provisioned on whichever node carries a matching pool
+    Examples:
+      | fsType |
+      |  zfs   |
+      |  ext4  |
+      |  xfs   |
+      |  btrfs |
```
tiagolobocastro: Why are we testing all filesystems here, surely this has no relation to what we are checking here, the poolname pattern?

tests/bdd/poolname_poolpattern.feature
```diff
@@ -0,0 +1,238 @@
...
+  # being weighted last would be preferred over a pool with room
+  Scenario: SpaceWeighted orders a node whose matching pool is full last, not first
+    Given a multi node cluster with a matching pool on each node
+    And the matching pool on one node is full and the matching pool on another node has free space
+    And a storage class is created with poolpattern as "zfspv-pool.*" and scheduler as SpaceWeighted and thinprovision as yes
+    When pvc is created referencing this same storage class
+    Then the zfsvolume must be provisioned on the node whose matching pool has free space
+    And the node whose matching pool is full must not be preferred over it
+
+  Scenario: SpaceWeighted applies to a fixed poolname as well
```
tiagolobocastro: shouldn't poolname (no pattern) be tested elsewhere?

tiagolobocastro (approved):

tests/bdd/poolname_poolpattern.feature
krishnaGajabi: moved it to common zfs.feature bdd file

tests/bdd/poolname_poolpattern.feature
krishnaGajabi: Removed all fileystem, and just kept the scenario

tests/bdd/poolname_poolpattern.feature
```diff
@@ -0,0 +1,225 @@
...
+
+  Scenario Outline: Volume provision using a matching poolpattern
+    Given a storage class is created with poolpattern as "zfspv-pool.*" and fsType as <fsType>
+    When pvc is created referencing this same storage class
+    Then pvc should be in bound state
+    And zfsvolume must be created and should be in Ready state to use by any application
+    And the zfsvolume PoolName must be zfspv-pool-a, the matching pool with the largest free space
+    Examples:
+      | fsType |
+      |  zfs   |
```
tiagolobocastro: also filesystem still here

tiagolobocastro (commented):
Still a few fsType outlines here, are those also meaningful here on poolpattern?

tests/bdd/poolname_poolpattern.feature
krishnaGajabi: i have removed removed fstype outline for most of the bdds, kept it for clone and snapshot bdd (only for zfs and ext4 volume). to check it once for dataset and zvol path.

PR #500, opened 2024-02-07
feat(deploy/helm): make snapshotcontroller optional

danielkza:
**Why is this PR required? What issue does it fix?**:

The CSI Snapshot Controler is already deployed by some Kubernetes distributions
such as RKE2. Make it possible to omit from the zfs-localpv chart for those cases.

**What this PR does?**:

Add `zfsController.snapshotController.enabled` Helm value, defaulting to true.
When set to false, the snapshot controller container is not created.

**Does this PR require any upgrade changes?**:

No.

**If the changes in this PR are manually verified, list down the scenarios covered:**:

**Checklist:**
- [ ] Fixes #<issue number>
- [x] PR Title follows the convention of  `<type>(<scope>): <subject>`
- [ ] Has the change log section been updated?
- [ ] Commit has unit tests
- [ ] Commit has integration tests
- [ ] (Optional) Are upgrade changes included in this PR? If not, mention the issue/PR to track:
- [ ] (Optional) If documentation changes are required, which issue on https://github.com/openebs/openebs-docs is used to track them:

PR #742, opened 2026-08-10
feat(helm): allow disabling the snapshot-controller

### LINKED ISSUE #736: add a flag to disable the snapshot-controller in the helm chart
[The chart always renders a snapshot-controller container alongside zfs-localpv](https://github.com/openebs/zfs-localpv/blob/b367b43211e9f01648a351a011bc88423d0b180b/deploy/helm/charts/templates/zfs-controller.yaml#L81-L91). I saw https://github.com/openebs/zfs-localpv/pull/500 which would solve this but I believe the reviewer accidentally confused the snapshot-controller with the csi-snapshotter there. My understanding is the snapshot-controller manages the volumeSnapshot CRDs in the cluster. it makes sense to only run one snapshot-controller container in-cluster (or leader-election, etc.) otherwise there are separate deployments with (possibly) separate versions trying to manage the same resources. Reading the review on the PR, it sounds like the reviewer thought the change made the *csi-snapshotter* opt-out.

A `zfsController.snapshotController.enabled` helm value, as the linked PR above implemented

I have not run into any issues running this snapshot-controller alongside another I deployed, I just recently noticed and, assuming my understanding is correct, it seems like it *could* cause issues.

**Environment:**
N/A

aclerici38:
There should only be 1 snapshot-controller per-cluster as the controller manages the volumeSnapshot CRDs.

**What this PR does?**:
Allows disabling the snapshot-controller container in the zfs-localpv pod. This should be disabled if there is already a snapshot-controller in-cluster e.g. provided by the k8s distribution or just managed separately.

No

```
anthony@anthony-laptop ~/zfs-localpv (fix-snapshot-controller)> helm template openebs deploy/helm/charts -s templates/zfs-controller.yaml --set zfsController.snapshotController.enabled=false | grep snapshot-controller
anthony@anthony-laptop ~/zfs-localpv (fix-snapshot-controller) [0|1]> helm template openebs deploy/helm/charts -s templates/zfs-controller.yaml | grep snapshot-controller
        - name: snapshot-controller
          image: "registry.k8s.io/sig-storage/snapshot-controller:v8.2.0"
```
Reimplements #500

**Checklist:**
- [x] Fixes #736
- [ ] Commit has unit tests


Abhinandan-Purkait (approved):

tiagolobocastro (approved):

krishnaGajabi:
Should this be implemented for other engines as well??
cc @Abhinandan-Purkait / @tiagolobocastro

tiagolobocastro:
> Should this be implemented for other engines as well?? cc @Abhinandan-Purkait / @tiagolobocastro

Indeed it should

krishnaGajabi:
> > Should this be implemented for other engines as well?? cc @Abhinandan-Purkait / @tiagolobocastro
> 
> Indeed it should

@aclerici38 , would you like to contribute similar change to other relevant openebs engines?

aclerici38:
Sure!

aclerici38:
Looking through the charts, IMO the snapshot-controller deployment should be controlled by the top-level openebs/openebs chart. I can create an issue if you'd like to talk further about it, but I think it would be best to have something like this in the umbrella chart
```
snapshotController:
  enabled: true
  args: []
  ...
```
which would spawn a separate deployment of the snapshot-controller instead of using the sidecars on each openebs controller. I will still need to go and add a flag for the sidecar on each of these subCharts, but my thought is the snapshot-controller can remain enabled by default in each subChart just disabled automatically in the umbrella chart to instead use the new deployment. That is a much cleaner pattern IMO, especially if managing multiple openebs controllers through the umbrella chart. Otherwise the admin would need to only enable one sidecar or the charts would have to unconditionally use leader-election with the snapshot-controller. 
Any thoughts on that?

edit: I made a draft here https://github.com/openebs/openebs/pull/4288

Abhinandan-Purkait:
Sure. Thanks! @aclerici38 Merging this.

PR #743, opened 2026-08-13
feat(controller): provision into a pool matching poolpattern

krishnaGajabi:
## Summary

Wires the pool-pattern-aware scheduler helpers merged in #739 into the
fresh-create path of `CreateZFSVolume`, so the `poolpattern` StorageClass parameter (OEP-4268)
actually selects a pool. Clone/snapshot-restore, `GetCapacity` and StorageClass
validation are untouched and follow in later tasks.

## What this PR does

**Reads `poolpattern`** and compiles it, together with `poolname`, into the single
`*regexp.Regexp` the helpers match pool roots against.

**Filters the scheduler's output for volumes that reserve space.** lib-csi's
scheduler only *orders* nodes — it takes its candidates from the cluster topology
and moves a node missing from the weight map to the front of the list rather than
dropping it — so nodes that cannot hold the reservation are removed from its
output via `filterKeep`, preserving the weighted order of the survivors. When the
intersection goes empty, the request fails immediately instead of attempting a
create that cannot succeed:

- a matching pool exists but nothing fits → `ResourceExhausted`, which
  external-provisioner retries with backoff and, under capacity tracking,
  reschedules onto a node with room
- the pattern matches no pool on any node → `FailedPrecondition`, since a
  StorageClass naming a pool that exists nowhere will not start working on a retry

Both replace today's generic `codes.Internal, "scheduler failed, node list is
empty"`.

**Resolves the concrete pool per candidate node** in the provisioning loop, so
`ZFSVolume.Spec.PoolName` always names a pool present on the node the volume was
assigned to. A fixed `poolname` keeps being used as-is, since it can carry a
dataset path the `ZFSNode` CR does not advertise.

## Behaviour change — needs a release note

The fit filter applies to the existing fixed-`poolname` path as well as to
`poolpattern`: a full or absent pool now fails fast instead of retrying a create
that cannot succeed. OEP-4268 accepts this as an improvement but records that it
must be called out in the release notes.

## Left for later tasks

- clone / snapshot-restore: the pool guard currently rejects every clone under a
  `poolpattern` StorageClass, since `poolname` is empty there — fixed in the
  next PR, together with `GetCapacity`
- `GetCapacity` under `poolpattern`
- `validateVolumeCreateReq` rejection of both-set / neither-set / invalid regex
- docs and samples

## Testing

`TestFilterKeep` covers that survivors keep the scheduler's weighted order, that
an unsuitable node is dropped rather than promoted to the front, the nothing-fits
case, and a suitable node outside the topology staying out. `TestPoolDesc` covers
the error-message rendering for both parameter forms.

`go build ./pkg/... ./cmd/...`, `go vet ./pkg/...` and `go test ./pkg/...` are
clean, and `gofmt` reports nothing.

Worth running `make ci profile=custom-node-id` (or `profile=all`): the regular
profile cannot exercise the node-name keying the intersection depends on, since
node name and node id are equal there.

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -248,6 +248,7 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 	kf := parameters["keyformat"]
 	kl := parameters["keylocation"]
 	pool := parameters["poolname"]
+	poolpattern := parameters["poolpattern"]
 	tp := parameters["thinprovision"]
 	schld := parameters["scheduler"]
 	fstype := parameters["fstype"]
@@ -280,7 +281,7 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 
 	// the pool parameters are matched as a single regular expression, an exact
 	// name being an anchored, quoted one
-	pattern, err := compilePoolPattern(pool, "")
+	pattern, err := compilePoolPattern(pool, poolpattern)
 	if err != nil {
 		return "", status.Error(codes.InvalidArgument, err.Error())
 	}
@@ -304,6 +305,51 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 		return "", status.Error(codes.Internal, "scheduler failed, node list is empty for creating the PV")
 	}
 
+	reserves := reservesSpace(vtype, tp)
+
+	// resolvePool only picks among the pools which can hold the reservation
+	var minFree int64
+	if reserves {
+		minFree = size
+	}
+
+	// only a poolpattern or a reservation needs the pools the agents reported,
+	// so a thin volume on a fixed poolname never waits on that report
+	var suitable map[string]bool
+	if poolpattern != "" || reserves {
+		var matched bool
+		var serr error
+
+		suitable, matched, serr = getSuitableNodes(pattern, size)
+		if serr != nil {
+			return "", status.Errorf(codes.Internal, "get suitable nodes failed : %s", serr.Error())
+		}
+
+		// retrying will not fix a pattern which matches no pool anywhere. Only a
+		// poolpattern is held to this, since a fixed poolname is used as it is
+		// and "zfs create" stays its arbiter even when an agent is behind.
+		if poolpattern != "" && !matched {
+			return "", status.Errorf(codes.FailedPrecondition,
+				"no pool matching %s is present on any node, volume %s",
+				poolDesc(pool, poolpattern), volName)
+		}
+	}
+
+	// the scheduler orders the nodes but never drops one, so the nodes which
+	// cannot hold the reservation are filtered out of its output here. The fit
+	// is best effort: the capacity is the pool root's, ignoring a dataset quota
+	// under a poolname, and the ZFSNode capacity is a periodic snapshot, so a
+	// create which slips through still fails on the node and CSI retries.
+	if reserves {
+		if prfList = filterKeep(prfList, suitable); len(prfList) == 0 {
+			// the pools are there but full: external-provisioner retries with a
+			// backoff and reschedules once a node has room
+			return "", status.Errorf(codes.ResourceExhausted,
+				"no pool matching %s has %d bytes free, volume %s",
+				poolDesc(pool, poolpattern), size, volName)
+		}
+	}
+
 	volObj, err := volbuilder.NewBuilder().
 		WithName(volName).
 		WithPVCName(pvcName).
@@ -312,7 +358,6 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 		WithCapacity(capacity).
 		WithRecordSize(rs).
 		WithVolBlockSize(bs).
-		WithPoolName(pool).
 		WithDedup(dedup).
 		WithEncryption(encr).
 		WithKeyFormat(kf).
@@ -332,17 +377,44 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 		return "", status.Error(codes.Internal, err.Error())
 	}
 
-	klog.Infof("zfs: trying volume creation %s/%s on node %s", pool, volName, prfList)
+	klog.Infof("zfs: trying volume creation %s in %s on nodes %v", volName, poolDesc(pool, poolpattern), prfList)
 
 	// try volume creation sequentially on all nodes
 	for _, node := range prfList {
 		var nodeid string
 		nodeid, err = zfs.GetNodeID(node)
 		if err != nil {
+			klog.Warningf("zfs: volume %s skipping node %s, no node id : %s", volName, node, err.Error())
 			continue
 		}
 
-		vol, _ := volbuilder.BuildFrom(volObj).WithOwnerNodeID(nodeid).WithVolumeStatus(zfs.ZFSStatusPending).Build()
+		// a poolname is used as it is, since it can carry a dataset path the
+		// ZFSNode CR does not advertise. A poolpattern resolves against the
+		// pools of the node picked, so the agent finds the pool on it.
+		volPool := pool
+		if poolpattern != "" {
+			if volPool, err = resolvePool(schld, nodeid, pattern, minFree); err != nil {
+				klog.Warningf("zfs: volume %s skipping node %s, pool lookup failed : %s", volName, node, err.Error())
+				continue
+			}
+			if volPool == "" {
+				// only reachable when the volume was not fit filtered
+				err = fmt.Errorf("no pool matching poolpattern %q on node %s", poolpattern, node)
+				klog.Warningf("zfs: volume %s skipping node %s : %s", volName, node, err.Error())
+				continue
+			}
+		}
+
+		var vol *zfsapi.ZFSVolume
+		vol, err = volbuilder.BuildFrom(volObj).
+			WithOwnerNodeID(nodeid).
+			WithPoolName(volPool).
+			WithVolumeStatus(zfs.ZFSStatusPending).Build()
+		if err != nil {
+			continue
+		}
+
+		klog.Infof("zfs: creating volume %s/%s on node %s", volPool, volName, node)
 
 		timeout := false
 
--- a/pkg/driver/schd_helper.go
+++ b/pkg/driver/schd_helper.go
@@ -85,6 +85,15 @@ func compilePoolPattern(poolname, poolpattern string) (*regexp.Regexp, error) {
 	return nil, errors.New("either poolname or poolpattern must be set")
 }
 
+// Names the pool selector as the storageclass declares it, for the errors and
+// the logs, where the compiled pattern would read back as "^zpool$".
+func poolDesc(poolname, poolpattern string) string {
+	if poolpattern != "" {
+		return fmt.Sprintf("poolpattern %q", poolpattern)
+	}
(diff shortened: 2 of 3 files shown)


pkg/driver/controller.go
```diff
@@ -300,12 +301,39 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
...
 	}
 
+	suitable, matched, serr := getSuitableNodes(pattern, size)
+	if serr != nil {
+		return "", status.Errorf(codes.Internal, "get suitable nodes failed : %s", serr.Error())
+	}
+
+	// a storageclass naming a pool which is present on no node at all will not
+	// start working on a retry, whatever the volume type
+	if !matched {
```
Abhinandan-Purkait: This makes the fixed `poolname` path depend on ZFSNode CR content too, even for volumes that don't reserve space. If the node agent is restarting or hasn't reported pools yet, a perfectly valid SC will fail with FailedPrecondition and a message saying retries won't help - but they will, once the agent catches up. Maybe gate this on `poolpattern != ""` only, or use ResourceExhausted for the transient case?

pkg/driver/controller.go
```diff
@@ -300,12 +301,39 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 		return "", status.Error(codes.Internal, "scheduler failed, node list is empty for creating the PV")
 	}
 
+	suitable, matched, serr := getSuitableNodes(pattern, size)
```
Abhinandan-Purkait: nit: `getNodeMap` just listed the ZFSNodes and this lists them again, so the two can see different snapshots. Might be cleaner to list once and pass it into both.

Also wondering about the sanity test hack above - the `parameters["node"]` path now goes through this check too. Does `make sanity` still pass?

pkg/driver/controller.go
```diff
@@ -300,12 +301,39 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
...
+		return "", status.Errorf(codes.FailedPrecondition,
+			"no pool matching %s is present on any node, volume %s",
+			poolDesc(pool, poolpattern), volName)
+	}
+
+	// a volume which gets a ZFS reservation has to fit in the free capacity of
+	// a single pool, everything else is created regardless of it. The scheduler
+	// only orders the nodes and never drops one, so the nodes which can not
+	// hold the reservation are filtered out of its output here.
+	if reservesSpace(vtype, tp) {
```
Abhinandan-Purkait: `poolname` can point at a dataset with its own quota, but the fit check only looks at the pool root's free space. Fine as best effort, just worth a comment so nobody trips on it later.

pkg/driver/controller.go
```diff
@@ -334,7 +362,33 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
...
 
-		vol, _ := volbuilder.BuildFrom(volObj).WithOwnerNodeID(nodeid).WithVolumeStatus(zfs.ZFSStatusPending).Build()
+		// a poolname is used as it is, since it can carry a dataset path which
+		// the ZFSNode CR does not advertise. A poolpattern is resolved against
+		// the pools of the node which was picked, so that the node agent always
+		// finds the pool the volume was assigned to.
+		volPool := pool
+		if poolpattern != "" {
+			if volPool, err = resolvePool(nodeid, pattern); err != nil {
+				continue
```
Abhinandan-Purkait, line 398: nit: when `resolvePool` fails we silently move to the next node and only the last err makes it into the final error. A warning log per skipped node would help debug multi-node failures.

Abhinandan-Purkait (commented):
Wiring looks good and matches the plan from #739. A few things I'd like to discuss before this goes in.

pkg/driver/controller.go
krishnaGajabi: addressed, see new commit

pkg/driver/controller.go
krishnaGajabi: Addressed, see new commit

pkg/driver/controller.go
krishnaGajabi: done

pkg/driver/controller.go
krishnaGajabi: **On the double list** — you're right that the two can see different snapshots.
`a3f5901d` narrows it: `getSuitableNodes` is now only called when there is a
`poolpattern` to resolve or a reservation to fit, so a thin volume on a fixed
`poolname` makes one list instead of two. A pattern or a thick volume still
lists twice.

I'll take the "list once, pass it in" refactor as a follow-up rather than fold
it in here. It touches `getNodeMap`, `getCapacityWeightedMap`,
`getSpaceWeightedMap` and `getVolumeWeightedMap` (which additionally needs the
ZFSVolume list) — all merged in #739, with their signatures written into
OEP-4268 — so it deserves its own review rather than riding along in this one.
The skew is also bounded by the same 60s ZFSNode poll each snapshot already
carries: one list is used only for ordering, the other only for filtering, and
`zfs create` stays the final arbiter, so it yields a slightly stale placement,
never a wrong one.

**On sanity** — `ci/sanity.sh` passes `poolname: "zfspv-pool"` with
`thinprovision: "yes"`, and after `a3f5901d` that combination skips the block
entirely: no extra list and no dependency on the ZFSNode report. `make sanity`
runs in CI on the `regular` profile and is green on this branch, as is the
`custom-node-id` BDD profile.

pkg/driver/controller.go
```diff
@@ -300,12 +301,60 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 		return "", status.Error(codes.Internal, "scheduler failed, node list is empty for creating the PV")
 	}
 
+	reserves := reservesSpace(vtype, tp)
+
+	// the pools the agents reported are needed to resolve a poolpattern and to
+	// fit a volume which reserves space. A thin volume on a fixed poolname needs
```
tiagolobocastro: I'm confused with this comment, a thin volume doesn't pick the best pool?

Also the comment makes it sound like it's and, but here I see an or?

```suggestion
	// fit a volume which reserves space. A volume on a fixed poolname needs
```

abhilashshetty04 (approved):

pkg/driver/controller.go
krishnaGajabi: A thin volume selects the best pool in all scenario. This code only handles a specific scenario >> When a thin volume is created using a fixed pool name path, pool pattern matching is not required. In addition, it does not need to evaluate the reserve-space fit filter, as that logic is skipped in the code.

The OR condition handles these two cases. If poolPattern is not empty, the code performs pool matching and also validates suitable nodes for volume reservation. For the fixed pool name path, getSuitableNode() is executed only when reserve space needs to be allocated. This ensures that the fit filter check is performed only when it is relevant, avoiding unnecessary validation for scenarios where reserve space allocation is not required.

pkg/driver/controller.go
```diff
@@ -324,17 +373,46 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
...
+			continue
+		}
+
+		// a poolname is used as it is, since it can carry a dataset path which
+		// the ZFSNode CR does not advertise. A poolpattern is resolved against
+		// the pools of the node which was picked, so that the node agent always
+		// finds the pool the volume was assigned to.
+		volPool := pool
+		if poolpattern != "" {
+			if volPool, err = resolvePool(nodeid, pattern); err != nil {
```
abhilashshetty04: https://github.com/openebs/zfs-localpv/pull/739/changes#diff-5df8dda72c6252053011798747eb39268924c39d7fa13c21884bdada3fde0bfeR274

It returns pool with max free space. We are going with `SpaceWeighted` algorithm here irrespective of user's choice ie (could be `VolumeWeighted` or `CapacityWeighted` also). Shouldn't we run the algorithm here again for these two algorithms too so that  we select the pool based on user's choice.

User provided algorithm should be considered for selecting node as well as picking the pool on the node.

pkg/driver/controller.go
tiagolobocastro: > When a thin volume is created using a fixed pool name path, pool pattern matching is not required

But why does a thin volume not need to select a pool with the most free space?

krishnaGajabi: It does select the pool with most free space for thin vols.
getsuitablenodes >> this function is purely run for reserving volumes or poolpattern created volumes  only (free space > volume size requested).This is the fail fast mechanism for thick volumes. This returns list of suitable nodes and nodes with matching poolpattern. if there is no macthing poolpattern on nodes it throws the error on line 327.

We build a prefered list of nodes from in [prflist](https://github.com/krishnaGajabi/zfs-localpv/blob/a3f5901d100c87acd4102baf839fc3d5ee9da721/pkg/driver/controller.go#L297) schedulable nodes and suitable nodes interesection for reserving volumes. if no nodes in prflist then a resource exhausted error is thrown.

schedulers help you select node based on scheduling algorithms(Not pools).

once a node is selected , resolvepool helps you select pool with most free space.

https://github.com/krishnaGajabi/zfs-localpv/blob/a3f5901d100c87acd4102baf839fc3d5ee9da721/pkg/driver/controller.go#L297

pkg/driver/controller.go
krishnaGajabi: once a node is selected based on scheduling algorithm, resolvepool helps you select pool with most free space on that node, irresepctive of scheduling algorithm (same thing happens in lvm i guess). Should we account for scheduling in pool selection logic also??

@Abhinandan-Purkait / @tiagolobocastro , wdyt?

abhilashshetty04: As discussed, localpv-lvm picks the vg in the node-agent. Controller only determines the node. I feel the intention was to ensure we select the vg with the most updated view of resource utilization. In node-agent, it currently does not know the user chosen algorithm. It just choses the vg with least free space that can accommodate the lv size. We should honor user algorithm there as well. 

Here, we know the user algorithm as pool is determined in controller. We should run algorithm again.

pkg/driver/schd_helper.go
```diff
@@ -280,8 +320,93 @@ func resolvePool(nodeid string, pattern *regexp.Regexp) (string, error) {
...
+	return poolForNode(zfsNode.Pools, pattern, minFree, pmap), nil
+}
+
+// Returns the pool mapping for the given scheduling algorithm, the
+// per pool counterpart of getNodeMap over the pools of one node.
+func getPoolMap(schd, nodeid string, pools []apis.Pool) (map[string]int64, error) {
+	if schd == VolumeWeighted {
+		return getPoolVolumeMap(nodeid)
+	}
+
```
abhilashshetty04, line 328: are we defaulting on `CapacityWeighted` for `SpaceWeighted` also?

krishnaGajabi: No Capacity weighted is default , see line 347

Abhinandan-Purkait (approved):
Please reduce the comments to be to the point, otherwise it just looks like noise in the code.

Abhinandan-Purkait:
Also, squash and rebase.

krishnaGajabi:
> Please reduce the comments to be to the point, otherwise it just looks like noise in the code.

done

krishnaGajabi:
> Also, squash and rebase.

done

tiagolobocastro (approved):

Abhinandan-Purkait:
@mergify queue

PR #752, opened 2026-08-18
feat(controller): clone, restore and capacity tracking under a poolpattern

krishnaGajabi:
## Summary

Completes the `poolpattern` work on the paths the fresh-create PR (#743) left
untouched: the clone and snapshot-restore guard, and capacity tracking (OEP-4268,
items 3 and 4). Rebased onto `develop` now that #743 has merged, so this is one
commit.

## Clone and snapshot restore

The guard compared the source pool against the `poolname` parameter, which a
`poolpattern` StorageClass leaves empty, so `source.Spec.PoolName != ""` was
always true and **every clone and restore under a pattern class was rejected**.

`sourcePoolAllowed` now checks the source pool against what the StorageClass
declares: an exact `poolname` as before, dataset path included, or the source
pool's **root** matched against a `poolpattern`.

The pool itself is still inherited from the source and never resolved here.
`getNodeMap`, `getSuitableNodes` and `resolvePool` deliberately do not run on
these paths: a clone cannot cross pools in ZFS, and resolving would pick a
*different* matching pool and break the clone.

## Reporting the resolved pool

`CreateZFSVolume`, `CreateVolClone` and `CreateSnapClone` now return the pool
alongside the node, so `CreateVolume` reports the pool the volume actually landed
in through the CSI volume context instead of the StorageClass parameter, which is
empty under a pattern class. A pattern-provisioned PV therefore shows its pool in
`spec.csi.volumeAttributes` rather than an empty string.

The pool is taken from the object each path provisioned, not read back from the
API server, so there is no extra `GET` per volume and no path that can report an
empty pool if such a read were to fail.

## Capacity tracking

`GetCapacity` matched pools by exact `poolname`, so a `poolpattern` class reported
zero capacity. With storage-capacity tracking enabled — which it is by default —
zero tells the scheduler that no node can hold the volume, and pods using such a
class never get scheduled.

It now matches with the compiled pattern and reports the free capacity of the
roomiest matching pool, keeping the maximum-volume-size semantics the function
already documented, and gaining the `QuoteMeta` quoting an exact `poolname` needs
in order not to over-match. A class declaring neither parameter or both still
reports zero; that is logged at `V(4)` rather than as a warning, since
`GetCapacity` is polled on an interval and the actionable rejection happens at
CreateVolume.

The hand-rolled pool-root parsing goes away with it — `poolRoot` and `maxFreePool`
already do that.

## StorageClass validation

Both-set, neither-set and an invalid regex are all rejected with
`InvalidArgument` by `compilePoolPattern`, which every create path calls before
it needs the compiled pattern. OEP-4268 originally listed this as a separate
change in `validateVolumeCreateReq`; that would only duplicate the check, since
each path needs the compiled `*regexp.Regexp` afterwards and would otherwise have
to compile the same expression twice or thread it down from the validator. The
OEP is being corrected accordingly in openebs/openebs#4304.

## Behaviour changes worth calling out

- A source pool outside what the StorageClass declares now fails with
  `FailedPrecondition` instead of `Internal`, matching how the fresh-create path
  reports the same class of misconfiguration. **This applies to fixed-`poolname`
  classes too**, so external-provisioner stops retrying something that cannot
  succeed.
- The pattern is compiled before the source lookup, so a request with both a
  malformed StorageClass and a missing source now returns `InvalidArgument`
  rather than `NotFound`.
- `CreateZFSVolume`, `CreateVolClone` and `CreateSnapClone` are exported, so
  their signatures change incompatibly for anything importing this package.
  `CreateVolume` is the only in-tree caller. This affects nothing at deploy time
  — no CRD, CSI or `ZFSVolume` spec change — but it is a source-level break for
  any out-of-tree importer, and worth a release-note line. Unexporting them would
  be the cleaner long-term answer, since they were only ever called from within
  the package; that is left out of this PR.

## Left for later

Docs and samples (OEP-4268 item 6).

## Testing

`TestSourcePoolAllowed` covers nine cases: exact `poolname` match and mismatch;
a dataset-path `poolname` against the same path, the pool root alone, and another
dataset in the same pool; a pattern matching a pool and matching a dataset path's
root; a non-match; and an unanchored pattern covering another pool as a substring,
which pins the OEP's unanchored-RE2 decision as intended rather than accidental.

`GetCapacity` itself has **no unit test** — it needs the ZFSNode informer cache
and `zfs.GetNodeID`. Its pool matching is now `maxFreePool`, which is covered, but
the parameter handling and the zero-capacity fallback are exercised only by BDD.

`gofmt`, `go vet ./pkg/driver`, `go build ./pkg/... ./cmd/...` and
`go test ./pkg/...` are clean. Note `go vet ./pkg/apis/...` already fails on
`develop`, on an `errors.Errorf` with a `%s` and no argument in `zfsvolume.go`;
unrelated to this PR.

`tests/bdd/poolname_poolpattern.feature` already specifies clone and restore under
a pattern class, that a clone inherits the source pool even when another matching
pool has more free space, and that capacity is reported non-zero per topology
segment.

### DIFF
--- a/pkg/driver/controller.go
+++ b/pkg/driver/controller.go
@@ -223,7 +223,7 @@ func waitForReadySnapshot(ctx context.Context, snapname string) error {
 }
 
 // CreateZFSVolume create new zfs volume from csi volume request
-func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string, error) {
+func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string, string, error) {
 	volName := strings.ToLower(req.GetName())
 	size := getRoundedCapacity(req.GetCapacityRange().RequiredBytes)
 
@@ -263,19 +263,19 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 		if vol.DeletionTimestamp != nil {
 			if _, ok := parameters["wait"]; ok {
 				if err := waitForVolDestroy(volName); err != nil {
-					return "", err
+					return "", "", err
 				}
 			}
 		} else {
 			if vol.Spec.Capacity != capacity {
-				return "", status.Errorf(codes.AlreadyExists,
+				return "", "", status.Errorf(codes.AlreadyExists,
 					"volume %s already present", volName)
 			}
 			if vol.Status.State != zfs.ZFSStatusReady {
-				return "", status.Errorf(codes.Aborted,
+				return "", "", status.Errorf(codes.Aborted,
 					"volume %s request already pending", volName)
 			}
-			return vol.Spec.OwnerNodeID, nil
+			return vol.Spec.OwnerNodeID, vol.Spec.PoolName, nil
 		}
 	}
 
@@ -283,12 +283,12 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 	// name being an anchored, quoted one
 	pattern, err := compilePoolPattern(pool, poolpattern)
 	if err != nil {
-		return "", status.Error(codes.InvalidArgument, err.Error())
+		return "", "", status.Error(codes.InvalidArgument, err.Error())
 	}
 
 	nmap, err := getNodeMap(schld, pattern)
 	if err != nil {
-		return "", status.Errorf(codes.Internal, "get node map failed : %s", err.Error())
+		return "", "", status.Errorf(codes.Internal, "get node map failed : %s", err.Error())
 	}
 
 	var prfList []string
@@ -302,7 +302,7 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 	}
 
 	if len(prfList) == 0 {
-		return "", status.Error(codes.Internal, "scheduler failed, node list is empty for creating the PV")
+		return "", "", status.Error(codes.Internal, "scheduler failed, node list is empty for creating the PV")
 	}
 
 	reserves := reservesSpace(vtype, tp)
@@ -322,14 +322,14 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 
 		suitable, matched, serr = getSuitableNodes(pattern, size)
 		if serr != nil {
-			return "", status.Errorf(codes.Internal, "get suitable nodes failed : %s", serr.Error())
+			return "", "", status.Errorf(codes.Internal, "get suitable nodes failed : %s", serr.Error())
 		}
 
 		// retrying will not fix a pattern which matches no pool anywhere. Only a
 		// poolpattern is held to this, since a fixed poolname is used as it is
 		// and "zfs create" stays its arbiter even when an agent is behind.
 		if poolpattern != "" && !matched {
-			return "", status.Errorf(codes.FailedPrecondition,
+			return "", "", status.Errorf(codes.FailedPrecondition,
 				"no pool matching %s is present on any node, volume %s",
 				poolDesc(pool, poolpattern), volName)
 		}
@@ -344,7 +344,7 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 		if prfList = filterKeep(prfList, suitable); len(prfList) == 0 {
 			// the pools are there but full: external-provisioner retries with a
 			// backoff and reschedules once a node has room
-			return "", status.Errorf(codes.ResourceExhausted,
+			return "", "", status.Errorf(codes.ResourceExhausted,
 				"no pool matching %s has %d bytes free, volume %s",
 				poolDesc(pool, poolpattern), size, volName)
 		}
@@ -374,7 +374,7 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 		Build()
 
 	if err != nil {
-		return "", status.Error(codes.Internal, err.Error())
+		return "", "", status.Error(codes.Internal, err.Error())
 	}
 
 	klog.Infof("zfs: trying volume creation %s in %s on nodes %v", volName, poolDesc(pool, poolpattern), prfList)
@@ -420,7 +420,7 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 
 		timeout, err = zfs.ProvisionVolume(ctx, vol)
 		if err == nil {
-			return nodeid, nil
+			return nodeid, volPool, nil
 		}
 
 		// if timeout reached, return the error and let csi retry the volume creation
@@ -434,16 +434,17 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
 		zfs.DeleteVolume(volName) // ignore error
 	}
 
-	return "", status.Errorf(codes.Internal,
+	return "", "", status.Errorf(codes.Internal,
 		"not able to provision the volume, nodes %v, err : %s", prfList, err.Error())
 }
 
 // CreateVolClone creates the clone from a volume
-func CreateVolClone(ctx context.Context, req *csi.CreateVolumeRequest, srcVol string) (string, error) {
+func CreateVolClone(ctx context.Context, req *csi.CreateVolumeRequest, srcVol string) (string, string, error) {
 	volName := strings.ToLower(req.GetName())
 	parameters := req.GetParameters()
 	// lower case keys, cf CreateZFSVolume()
 	pool := helpers.GetInsensitiveParameter(&parameters, "poolname")
+	poolpattern := helpers.GetInsensitiveParameter(&parameters, "poolpattern")
 	size := getRoundedCapacity(req.GetCapacityRange().RequiredBytes)
 	volsize := strconv.FormatInt(int64(size), 10)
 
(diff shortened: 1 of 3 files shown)


pkg/driver/controller.go
```diff
@@ -223,7 +223,7 @@ func waitForReadySnapshot(ctx context.Context, snapname string) error {
 }
 
 // CreateZFSVolume create new zfs volume from csi volume request
-func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string, error) {
+func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string, string, error) {
```
tiagolobocastro: Can't we just return the volume here rather than a tuple of strings?
Then caller can just take `vol.Spec.OwnerNodeID` or `vol.Spec.PoolName` ?

pkg/driver/controller.go
```diff
@@ -524,17 +538,19 @@ func CreateSnapClone(ctx context.Context, req *csi.CreateVolumeRequest, snapshot
...
 	}
 
-	if snap.Spec.PoolName != pool {
-		return "", status.Errorf(codes.Internal,
-			"clone to a different pool src pool %s dst pool %s",
-			snap.Spec.PoolName, pool)
+	// the restore lands in the snapshot's pool, inherited with the spec below
+	// and only checked here. Retrying will not widen the storageclass
+	if !sourcePoolAllowed(snap.Spec.PoolName, pool, pattern) {
+		return "", "", status.Errorf(codes.FailedPrecondition,
```
tiagolobocastro: Shouldn't this be invalid argument?

pkg/driver/controller.go
```diff
@@ -434,36 +434,44 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
...
 }
 
 // CreateVolClone creates the clone from a volume
-func CreateVolClone(ctx context.Context, req *csi.CreateVolumeRequest, srcVol string) (string, error) {
+func CreateVolClone(ctx context.Context, req *csi.CreateVolumeRequest, srcVol string) (string, string, error) {
 	volName := strings.ToLower(req.GetName())
 	parameters := req.GetParameters()
 	// lower case keys, cf CreateZFSVolume()
 	pool := helpers.GetInsensitiveParameter(&parameters, "poolname")
+	poolpattern := helpers.GetInsensitiveParameter(&parameters, "poolpattern")
```
tiagolobocastro, line 447: Shouldn't we verify here that only 1 must be set?

pkg/driver/schd_helper_test.go
```diff
@@ -133,6 +133,59 @@ func TestCompilePoolPattern(t *testing.T) {
...
+			srcPool: "zfspv-pool-b/k8s/localpv", poolpattern: "^zfspv-pool-", allowed: true,
+		},
+		"poolpattern not matching the source pool": {
+			srcPool: "other-pool", poolpattern: "^zfspv-pool-", allowed: false,
+		},
+		// a poolpattern is unanchored, like lvm-localpv's vgpattern, so it covers
+		// every pool whose name contains it. Users anchor with ^...$ to narrow it
+		"unanchored poolpattern covering another pool as a substring": {
+			srcPool: "other-pool", poolpattern: "pool", allowed: true,
+		},
```
tiagolobocastro, line 176: missing test for both name and pattern specified

pkg/driver/controller.go
krishnaGajabi: Yes its doable, ill make the changes

pkg/driver/controller.go
krishnaGajabi: It is verified, just below at compilePoolPattern(pool, poolpattern) on line 455 — that returns an error when both are set, when neither is, or when the pattern does not compile, and all three come back as InvalidArgument. Doing it in the one place that has to compile the expression anyway avoids a second check that would either duplicate the logic or compile the regex twice.

pkg/driver/schd_helper_test.go
krishnaGajabi: The both-set case is covered in TestCompilePoolPattern — "both set is rejected": {poolname: "zfspv-pool", poolpattern: "zfspv-pool.*", wantErr: true} — and end to end by the A storage class must set exactly one of poolname and poolpattern scenario in poolname_poolpattern.feature. 

It is absent from TestSourcePoolAllowed because it cannot reach that function: the table calls compilePoolPattern with require.NoError first, exactly as the clone paths do, so a both-set StorageClass fails before sourcePoolAllowed is ever called.

pkg/driver/controller.go
```diff
@@ -592,24 +607,24 @@ func (cs *controller) CreateVolume(
...
 	}
 
-	klog.Infof("created the volume %s/%s on node %s", pool, volName, selectedNodeID)
+	klog.Infof("created the volume %s/%s on node %s", selectedPool, volName, selectedNodeID)
 
 	sendEventOrIgnore(pvcName, volName, strconv.FormatInt(int64(size), 10), analytics.VolumeProvision)
 
 	topology := map[string]string{zfs.ZFSTopologyKey: selectedNodeID}
-	cntx := map[string]string{zfs.PoolNameKey: pool, zfs.OpenEBSCasTypeKey: zfs.ZFSCasTypeName}
+	cntx := map[string]string{zfs.PoolNameKey: selectedPool, zfs.OpenEBSCasTypeKey: zfs.ZFSCasTypeName}
```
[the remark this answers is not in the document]
krishnaGajabi: Agreed on the gap, but it's larger than an added assertion: the poolpattern scenarios in tests/bdd/poolname_poolpattern.feature are spec only — there is no Go implementation for any of the 20, and no poolpattern reference anywhere in tests/*.go. Implementing them also needs the CI cluster to have more than one pool, which ci/ci-test.sh does not set up today. I'd rather do that as its own PR than bolt a partial version onto this one; the volume-context assertion you're asking for falls out of it naturally.

pkg/driver/controller.go
krishnaGajabi: I think you're right for this one, and the distinction is whether waiting can ever help. On the fresh-create path FailedPrecondition is for "the pattern matches no pool anywhere", which is genuinely cluster state — a node agent that hasn't reported yet can make it true later. Here the source volume's pool is fixed for the life of the volume, so no amount of retrying or state change will ever make this clone valid; the caller has handed us a StorageClass and a source that cannot go together. That is an argument problem, so InvalidArgument fits better. I'll change both clone paths.

krishnaGajabi: addressed in new commit

pkg/driver/controller.go
krishnaGajabi: addressed in new commit

pkg/driver/controller.go
tiagolobocastro: That is a bit confusing, we should not do that in `compilePoolPattern` IMHO as it's definitely not clear from the function name
At the very least we should at least create a comment above the calling side

krishnaGajabi: You are right, I have changed the name of the function in the new commit. Also added appropriate function usage comment for it

pkg/driver/controller.go
```diff
@@ -322,14 +322,14 @@ func CreateZFSVolume(ctx context.Context, req *csi.CreateVolumeRequest) (string,
...
 		if serr != nil {
-			return "", status.Errorf(codes.Internal, "get suitable nodes failed : %s", serr.Error())
+			return nil, status.Errorf(codes.Internal, "get suitable nodes failed : %s", serr.Error())
 		}
 
-		// retrying will not fix a pattern which matches no pool anywhere. Only a
-		// poolpattern is held to this, since a fixed poolname is used as it is
-		// and "zfs create" stays its arbiter even when an agent is behind.
+		// a poolpattern matching no pool anywhere cannot be resolved yet; the
+		// pools may still be reported, so this stays retryable, unlike the
```
Abhinandan-Purkait: nit: external-provisioner treats FailedPrecondition, InvalidArgument and Internal all as ProvisioningFinished with the same backoff, so this is not actually more retryable than the clone guards, maybe reword to avoid implying that.

pkg/driver/schd_helper.go
```diff
@@ -94,6 +94,17 @@ func poolDesc(poolname, poolpattern string) string {
 	return fmt.Sprintf("poolname %q", poolname)
 }
 
+// Reports whether the pool a clone's source lives in is covered by what the
+// storageclass declares. A clone always lands in the source's pool, so this
+// only validates. A poolname is compared as it is, dataset path included, a
+// poolpattern against the source pool's root.
+func sourcePoolAllowed(srcPool, poolname string, pattern *regexp.Regexp) bool {
```
Abhinandan-Purkait, line 101: Worth calling out in #761 that a clone/restore under a poolpattern class is rejected when the pattern does not cover the source pool, since the clone inherits the pool instead of resolving one.

Abhinandan-Purkait (approved):
LGTM, fixed-poolname paths behave the same as before, just a couple of small notes.

pkg/driver/controller.go
krishnaGajabi: Done

pkg/driver/schd_helper.go
krishnaGajabi: Done

abhilashshetty04 (approved):

tiagolobocastro (approved):

Abhinandan-Purkait:
@mergify queue

PR #754, opened 2026-08-19
fix(ci): remove hard-coded HOME from Nix setup action

rohan2794:
The HOME override breaks on runners where the actual user doesn't match the hard-coded path. Removing it lets the OS provide the correct value for both GitHub-hosted and self-hosted runners.


Abhinandan-Purkait (approved):

tiagolobocastro:
@mergify backport release/2.11

PR #759, opened 2026-08-28
fix(xfs): keep new xfs volumes mountable on kernels below 5.19

Abhinandan-Purkait:
mkfs.xfs from xfsprogs 6.5, which the alpine based image ships, turns on nrext64 by default. Only kernel 5.19 and above can mount such a filesystem, so on older nodes the xfs volume fails to mount with `Superblock has unknown incompatible features (0x20) enabled`.

The kernel does not list the xfs features it knows anywhere, so at startup the node agent formats a throwaway image with nrext64 and tries to mount it. If the kernel turns it down, the volumes of that node are formatted with `-i nrext64=0`. A custom kernel is then taken as it is, whether the feature is backported or left out. The kernel version is only used when the probe can not run.

Tested on kernels 5.15, 5.18, 5.19 and 6.8, with xfsprogs 5.16, 6.2 and 6.17.

### DIFF
--- a/pkg/driver/agent.go
+++ b/pkg/driver/agent.go
@@ -53,6 +53,10 @@ type node struct {
 func NewNode(d *CSIDriver) csi.NodeServer {
 	var ControllerMutex = sync.RWMutex{}
 
+	// find out how the filesystems of this node have to be formatted, the
+	// answer holds for as long as this agent runs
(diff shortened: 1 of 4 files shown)


Abhinandan-Purkait:
Closing this in support of #760

PR #761, opened 2026-08-31
docs(poolpattern): document pool patterns and the SpaceWeighted scheduler

krishnaGajabi:
## Summary

Documentation, sample and release notes for the `poolpattern` work — OEP-4268
item 6, and the last piece of the series. #743 and #752 are both merged, so this
describes shipped behaviour rather than anticipated behaviour.

## What changed

**`docs/storageclasses.md`**

- New **poolpattern** parameter section: RE2, unanchored like lvm-localpv's
  `vgpattern`, matched against pool names and never dataset paths, and not
  required to exist on every node. Records that the resolved pool is written to
  the volume and never changes, and that a clone or snapshot restore inherits
  its source's pool — so a pattern that does not cover that pool rejects it.
- **poolname** now states that exactly one of the two must be set. Its heading is
  unchanged so existing anchors keep working.
- The scheduler section covers three algorithms rather than two, and notes that
  the chosen algorithm also picks the pool among a node's matching pools.
- New usage example, *StorageClass Selecting Pools by Pattern*, with the
  `kubectl` command to read the chosen pool back off the PV.

**`docs/scheduler.md`**

- Three algorithms, with guidance on when `SpaceWeighted` beats the
  `CapacityWeighted` default.
- New *Choosing the pool* and *Capacity* subsections: how the pool is picked
  within a node, which volumes are restricted by free space, the
  `ResourceExhausted` / `FailedPrecondition` split, and that the fit is best
  effort with `zfs create` as the final arbiter.

**`deploy/sample/zfspoolpattern.yaml`** — new sample StorageClass and PVC,
following the naming of the other `zfs*.yaml` samples and the shape of
`fio.yaml`.

**`CHANGELOG.md`** — the two features, plus the behaviour changes existing
StorageClasses see on upgrade.

## Behaviour changes recorded

- `CapacityWeighted` now weighs a node by the pool's real used capacity from the
  node agent rather than the summed capacity of driver-provisioned volumes, so
  node ordering can differ from previous releases for existing StorageClasses.
- A space-reserving volume is placed only where the reservation fits, and fails
  fast with `ResourceExhausted` or `FailedPrecondition` rather than retrying a
  create that cannot succeed. Thin volumes are unaffected.
- A cross-pool clone or restore fails with `InvalidArgument` rather than
  `Internal`.
- `CreateZFSVolume`, `CreateVolClone` and `CreateSnapClone` return the
  provisioned `ZFSVolume` rather than the node id. They are exported, so this is
  source incompatible for an out-of-tree importer; no effect on the driver as
  deployed.

## One thing to confirm

The changelog gains a **Behaviour Changes** heading, which v2.11.0 does not have
— it has *New Features and Enhancements*, *Bug Fixes and Improvements* and
*Testing and Continuous Integration*. These four items are neither features nor
bug fixes and I did not want to bury them, but happy to fold them into *Bug Fixes
and Improvements* if you would rather keep the existing headings. The entries
also carry no `by @author in PR NNN` line, matching the current unreleased
v2.12.0 entry; the released sections all have them, so I assume they are added at
release time.

## Testing

Docs only, no code changes. The sample manifest parses as valid YAML with the
expected kinds, and every internal anchor in both documents resolves, including
the cross-document link from `scheduler.md` to the `poolpattern` section.


Abhinandan-Purkait:
One small docs nit: the FailedPrecondition and capacity-fit behavior only applies to fresh poolpattern provisioning, not fixed poolname or clone/restore paths.

Abhinandan-Purkait (approved):

tiagolobocastro (approved):

Abhinandan-Purkait:
@mergify queue

tiagolobocastro:
@mergify queue

